PR #4, opened 2026-06-13
Add Lean boilerplate / CI and a formalization of the binding-signature balance argument

daira:
Make a start on bringing formal-verification work into this repo, alongside the mdbook.

Formalizes the balance soundness of the Zcash binding signature (protocol spec §4.13 Sapling / §4.14 Orchard) in Lean 4. This is done over an abstract `F`-module so it applies to both the Pallas (Orchard) and Jubjub (Sapling) value-commitment groups. (The inclusion of Sapling here adds very little extra code.)

The mdbook under `book/` is untouched. The Lean lives under `Zcash/` (package and lib both `Zcash`). Both CIs are path-gated via a gate-job pattern: the `Lean` and book workflows run their heavy job only on relevant changes, but still report a required check when the build is skipped.

## Commits

1. Lean project scaffolding (lakefile, pinned toolchain `v4.30.0-rc2`, Mathlib manifest).
2. Lean CI via `leanprover/lean-action` (toolchain + Mathlib olean cache + `lake build`), plus the gate-job path-gating of both CIs.
3. onwards — the binding-signature balance argument, cherry-picked from the [original `zcash-lean` branch](https://github.com/daira/zcash-lean/pull/2) (history preserved), then tidied; the two pools' no-overflow bounds unified through a shared helper; and the scalar field orders imported from CompElliptic. The final commit restates the balance theorems as explicit DLR reductions (see below).

## The argument (`Zcash/Security/BindingSignature/`)

- `Balance.lean` — the shared algebraic core (its lemmas are mapped to the spec in the table below). Binding is modelled as a **reduction**: a non-balancing verifying bundle exhibits an explicit nontrivial relation between the value base $\mathcal{V}$ and randomness base $\mathcal{R}$ (`HasNontrivialRelation`). So each balance theorem concludes `... = 0 ∨ HasNontrivialRelation V R` — it balances, or it solves the discrete-log relation problem in the relevant group. The integer no-overflow argument reduces to a shared `natAbs_lt_of_vSumBound` (`vSumBound N = N·(2^64−1) + 2^63`), instantiated per pool.
- `Orchard.lean` / `Sapling.lean` — the per-pool no-overflow arguments, each a thin wrapper over `natAbs_lt_of_vSumBound`:
  - **Orchard**: each action's signed net value `v = v_spend − v_output ∈ [−2^64+1, 2^64−1]`; the action count is bounded directly by the consensus rule `n ≤ 2^16−1`.
  - **Sapling**: the counts are derived from the 2 MB transaction-size limit against the minimum description sizes (summed field-by-field, over v4 and v5); `|vSum|` is bounded by the uniform per-element magnitude `|v| ≤ 2^64−1` (as Orchard's signed values are — by concatenating the spends with the negated outputs) rather than the spec's tighter asymmetric range, which suffices since any bound below the field order does.
  - The Pallas and Jubjub scalar field orders are imported from [CompElliptic](https://github.com/daira/CompElliptic) (daira/CompElliptic#1), where they carry machine-checked Lucas/Pratt primality certificates. The `*VSumBound_lt_*ScalarOrder` theorems  instantiate each bound at the concrete field.

## Relation to the spec (§4.13 / §4.14)

The development follows the spec's pencil-and-paper argument step for step:

| This formalization | Spec §4.13 / §4.14 |
|---|---|
| `V`, `R` | value base $\mathcal{V}$, randomness base $\mathcal{R}$ |
| `valueCommit V R v rcv` | $\mathsf{ValueCommit}_{\mathsf{rcv}}(\mathsf{v}) = [\mathsf{v}] \mathcal{V} + [\mathsf{rcv}]\mathcal{R}$ |
| `bindingVK` / `bsk` | $\mathsf{bvk}$, the binding validating key / $\mathsf{BindingPrivate}$, the signing key |
| `bindingVK_decomp` | $\mathsf{bvk} = [\mathsf{v_{sum}}] \mathcal{V} + [\mathsf{bsk}]\mathcal{R}$ |
| `hExtract : bvk = bsk • R` | a binding signature proves knowledge of $\mathsf{bsk} = \mathsf{dlog}_{\mathcal{R}}\,\mathsf{bvk}$ |
| `A` (stage 1) / `vSum` (stage 2) | $\mathsf{v_{sum}} = \sum \mathsf{v_{in}} - \sum \mathsf{v_{out}} - \mathsf{vBalance}$ |
| `HasNontrivialRelation` | a nontrivial DL relation $[a] \mathcal{V} + [b] \mathcal{R} = \mathcal{O}$, $(a,b) \neq (0,0)$ — a DLR solution |
| `relation_of_imbalance` → `value_coeff_zero_reduction` → `bundle_mod_balances_reduction` | $\mathsf{v_{sum}} \neq 0 \pmod r$ *exhibits* a DLR solution |
| `intBalance_eq_zero_of_lt` | $\mathsf{v_{sum}}$ does not overflow $\mathsf{ValueCommitType}$ ⟹ $\mathsf{v_{sum}} = 0$ over $\mathbb{Z}$ or a DLR solution |

The two stages are the spec's two steps: the algebraic core gives balance modulo the scalar-field order (or, in the reduction's other branch, a DLR solution), and the no-overflow lift upgrades it to integer balance. The whole development is over an abstract `F`-module, which is instantiated at `ZMod r` for the Pallas and Jubjub scalar fields in the `Orchard` / `Sapling` modules. RedDSA extractability is assumed (`hExtract`).

## Status

Builds green on Lean `v4.30.0-rc2` / Mathlib (pinned), with CompElliptic pinned by git rev.

The per-pool capstones `orchard_bundle_balances_reduction` / `sapling_bundle_balances_reduction` wire the `*_natAbs_lt` bounds into `bundle_integer_balances_reduction` end-to-end, so each consumes only the value-type hypotheses and RedDSA extractability (`hExtract`), concluding the reduction `vSum = 0 ∨ HasNontrivialRelation V R`. There is no information-theoretic binding assumption (it would be false in a prime-order group).

Remaining: decide what framework we will use to model the AGM, and tie it into that framework.


daira (commented):
Self-review ACK

Zcash/Security/BindingSignature/Orchard.lean
```diff
@@ -0,0 +1,68 @@
...
+This module instantiates the generic `natAbs_lt_of_abs_le` from `Balance` at those concrete bounds,
+discharging the `hbound` hypothesis of `bundle_integer_balances` for Orchard.
+-/
+
+namespace Zcash.Security.BindingSignature
+
+/-- The Pallas scalar field order `r_ℙ` — the order of the Pallas group, which is the scalar field of
+the Orchard value commitment (spec § Pallas and Vesta). Imported from `CompElliptic.Fields.Pasta`,
+where it carries a Lucas/Pratt primality certificate. -/
+@[reducible] def pallasScalarOrder : ℕ := CompElliptic.Fields.Pasta.PALLAS_SCALAR_CARD
```
TalDerei, line 22: noticed a typo in the spec; it specifies the prime order of the jubjub curve rather than pallas under the orchard section.

Image

Zcash/Security/BindingSignature/Balance.lean
```diff
@@ -0,0 +1,307 @@
...
+## How binding is expressed (and why not as "no relation exists")
+
+In a prime-order group, `V` and `R` are *always* discrete-log-related — a nontrivial `(a,b)` with
+`a • V + b • R = 0` exists; you simply cannot find it. So the information-theoretic statement
+"the only relation is trivial" is *false* in the group setting and must not be used as the
+binding hypothesis. Instead we phrase binding as a reduction:
+
+* `§ Binding reduction` below proves, with no cryptographic hypothesis, that a non-balancing
+  verifying bundle **exhibits** an explicit nontrivial relation between `V` and `R`
+  (`relation_of_imbalance`), equivalently the discrete log `dlog_R V` (`rand_log_of_imbalance`).
```
TalDerei: ACK – modeling the binding argument as a reduction rather than assuming no nontrivial DLOG relation exists.

Zcash/Security/BindingSignature/Balance.lean
```diff
@@ -0,0 +1,307 @@
...
+
+The range / no-overflow lift from field balance to integer balance is already built (`§ Integer
+balance` below: `intBalance_eq_zero_of_lt`, discharged per pool by `orchard_natAbs_lt` /
+`sapling_natAbs_lt` from the 64-bit value-type ranges). What remains is cryptographic and
+computational, not algebraic:
+
+## Assumptions
+
+* **RedDSA extractability** (`bvk = bsk • R` from a verifying binding signature) — assumed, not
+  proved (its proof needs a random oracle + forking); supplied as the `hExtract` hypothesis.
```
TalDerei, line 46: ACK — i think this is the right abstraction boundary (treating `hExtract` as an assumption). formalizing RedDSA only matters if we want to prove extractability itself; as you noted for the balance argument here, we can black box `bvk = [bsk]R`.

Zcash/Security/BindingSignature/Balance.lean
```diff
@@ -0,0 +1,307 @@
...
+Such a relation always exists in the group; the content is that imbalance *produces* one.
+Under hardness of the discrete-log relation problem, that cannot happen, so the bundle balances. -/
+theorem relation_of_imbalance (V R bvk : M) (A B bsk : F)
+    (hA : A ≠ 0)
+    (hExtract : bvk = bsk • R) (hSum : bvk = A • V + B • R) :
+    A ≠ 0 ∧ A • V + (B - bsk) • R = 0 := by
+  refine ⟨hA, ?_⟩
+  rw [smul_value_eq_smul_rand V R bvk A B bsk hExtract hSum, ← add_smul]
+  have hc : (bsk - B) + (B - bsk) = (0 : F) := by ring
+  rw [hc, zero_smul]
```
TalDerei, line 86: I really like the way this is modeled! `relation_of_imbalance` proves the counterfactual, and we claim DLR hardness in the docs to close it.

Zcash/Security/BindingSignature/Balance.lean
```diff
@@ -0,0 +1,307 @@
...
+  have hVR := smul_value_eq_smul_rand V R bvk A B bsk hExtract hSum
+  have h : A⁻¹ • (A • V) = A⁻¹ • ((bsk - B) • R) := by rw [hVR]
+  rwa [smul_smul, smul_smul, inv_mul_cancel₀ hA, one_smul] at h
+
+/-! ### Abstract model
+
+The information-theoretic binding predicate below is satisfiable only when `V` and `R` are
+genuinely independent — e.g. in a free `F`-module of rank ≥ 2. It validates the algebra of the
+reduction but is *false* in a prime-order group and is never instantiated there; the group uses
+`relation_of_imbalance` and DLR hardness instead. -/
```
TalDerei: "the group uses `relation_of_imbalance`", however `relation_of_imbalance` / `rand_log_of_imbalance` are never called by any theorem (currently only referenced in the docs). is this some forward scaffolding that will be wired in, or will it stay proven in isolation?

TalDerei (commented):
some observations; went through and cross referenced against 4.13/4.14 in the spec.

Zcash/Security/BindingSignature/Orchard.lean
daira: You are correct, thanks. Will fix.

Zcash/Security/BindingSignature/Balance.lean
daira: I do want to prove RedDSA is strongly unforgeable (in the correct sense for a key-rerandomizable scheme); but that's a separate piece. Ariel formally proved the argument we use for the key-rerandomization part of Spend authority at some point for Sapling IIRC, but I don't remember whether that was fully accurate to RedDSA. I'll look at it but I expect we'll want to redo the model to make what we're proving strong enough and precise enough.

Zcash/Security/BindingSignature/Balance.lean
daira: Well spotted. I'll put this back in draft until I have time to clean this up.

Zcash/Security/BindingSignature/Orchard.lean
daira: This will be done in zcash/zips#1300.

Zcash/Security/BindingSignature/Balance.lean
```diff
@@ -87,63 +85,52 @@ theorem relation_of_imbalance (V R bvk : M) (A B bsk : F)
...
-theorem rand_log_of_imbalance (V R bvk : M) (A B bsk : F) (hA : A ≠ 0)
+/-- A non-balancing verifying bundle yields the discrete log of `V` base `R`, namely
+`dlog_R V = A⁻¹ (bsk − B)`. This is just a more explicit way of saying `relation_of_imbalance`;
+it is the latter that is wired into the balance reduction below. -/
+theorem imbalance_yields_discrete_log (V R bvk : M) (A B bsk : F) (hA : A ≠ 0)
     (hExtract : bvk = bsk • R) (hSum : bvk = A • V + B • R) :
     V = (A⁻¹ * (bsk - B)) • R := by
   have hVR := smul_value_eq_smul_rand V R bvk A B bsk hExtract hSum
   have h : A⁻¹ • (A • V) = A⁻¹ • ((bsk - B) • R) := by rw [hVR]
   rwa [smul_smul, smul_smul, inv_mul_cancel₀ hA, one_smul] at h
```
TalDerei, line 96: nit; dangling theorem and should be documented as such at least.

Zcash/Security/BindingSignature/Balance.lean
```diff
@@ -0,0 +1,299 @@
...
+* **RedDSA extractability** (`bvk = bsk • R` from a verifying binding signature) — assumed, not
+  proved (its proof needs a random oracle + forking); supplied as the `hExtract` hypothesis.
+* **DLR hardness** — the assumption the reduction discharges against to conclude actual balance
+  (the discrete-log relation problem, tightly equivalent to DL).
+
+## Not yet built
+
+* **The AGM / DLR wrapper** — the computational / AGM layer that consumes `relation_of_imbalance`
+  against DLR hardness.
+-/
```
TalDerei, line 54: ACK – `relation_of_imbalance` reduction is wired in but not yet discharged, will research AGM impl in ArkLib as a (plausible?) host.

Zcash/Security/BindingSignature/Balance.lean
TalDerei: referencing https://github.com/zcash/ironwood/pull/4#discussion_r3415528911, can we add a TODO comment if proving extractability  is an expected follow-up?

TalDerei (approved):
some minor doc nits, otherwise looks good!

Zcash/Security/BindingSignature/Balance.lean
daira: I'll do that in a follow-up PR.

Zcash/Security/BindingSignature/Balance.lean
daira: Eh, it's a useful thing to prove in its own right, and the documentation almost says it's dangling already.

PR #9, opened 2026-06-27
feat(fv): conditional orchard verifier soundness

TalDerei:
**Human summary:**

Implements the checklist in https://github.com/TalDerei/halo2-fv. It uses personal Halo2 / Orchard forks (TalDerei/halo2#1 and TalDerei/orchard#1) that expose helper APIs for demonstrating MSM fingerprint equality.

Supersedes my fork impl https://github.com/TalDerei/ironwood/pull/1 and rebases commit history into logical groupings. The first 9 commits implement the verifier and fingerprinting faithfulness, and the remaining commits implement special soundness or address review feedback.

Specifically, this implements abstract IPA special soundness (detached from the real verifier), but hasn't yet introduced the binding layer as a DLR hardness reduction that will "bridge" the abstract IPA with the real verifier, doesn't construct the AGM (like https://github.com/zcash/ironwood/pull/4, which will discharge the binding relation and reduce to DL assumption), etc.

This PR proves _conditional_ verifier soundness in Lean for the deployed Halo2/Orchard verifier. It is conditional in the sense that it implements sections 1 and 2 of the checklist (it rederives the verifier logic in Lean and fingerprints it against the Rust implementation, with the VK treated as a given input).

It doesn't implement section 3, which would rederive the Orchard circuit/VK in Lean and prove that the Rust VK encodes exactly the intended high-level Orchard statement. That would require implementing the action circuit and VK derivation itself in Lean. This missing section is what would be needed to lift the result from VK-level verifier soundness to full Orchard SNARK soundness.

**This makes a few assumptions:**
  - The verifying key is taken as an input, not derived in Lean.
  - Fiat-Shamir is hand waved and modeled as a black-box random oracle.
  - Commitment binding is modeled as a discrete-log-relation (DLR) hardness reduction (just not wired into the soundness path).
  - The proof relies on forking/rewinding assumptions for the Fiat-Shamir challenges used in IPA special-soundness extraction.
  - Schwartz-Zippel “good challenge” exclusions hold for the sampled challenges.
  - VK correctness is assumed: the given VK encodes the intended high-level Orchard relation.
  - The Hasse bound for the Vesta curve is assumed (used by CompElliptic to pin the order of the curve).'
 
Follow-up work to address these assumptions is tracked in https://github.com/zcash/ironwood/issues/16.

----------------------------

**AI Summary:**

The deployed Halo2/Orchard verifier is a public-coin IPA argument over the Pasta cycle: it reads the proof as opaque Vesta points (`E_q`) and `F_p` scalars, squeezes Fiat–Shamir challenges, assembles one multiscalar multiplication, and accepts exactly when that MSM is the group identity. The MSM *is* the verifier — a faithful *fingerprint* of every check. This formalizes that fingerprint and its soundness in Lean 4, over an abstract `F`-module so the algebra is curve-agnostic, instantiated at the concrete Vesta curve.

**§1 — the assembly.** `assemble : proof string × challenges → MSM` reproduces the Halo2 construction line for line: the per-argument contributions (gates, permutation, lookup, the vanishing argument), the multiopen compression (`x₁…x₄`, `construct_intermediate_sets`, the `q′` fold), and the IPA fold (`compute_s`/`compute_b`, the `(Lⱼ, Rⱼ)` rounds, the `u`/`w` coefficients). In place of a transpilation proof, a *fingerprint match* against the Rust verifier's captured MSM keeps the reproduction honest — any divergence surfaces as a coefficient mismatch, with a Schwartz–Zippel `≤ d/p` bound for the cheaper random-evaluation check.

**§2 — the soundness composition.** From an accepting fingerprint the SNARK relation is composed modulo named bridge assumptions — a conditional scaffold rather than end-to-end soundness (assumption inventory in the capstone docstrings; see https://github.com/zcash/ironwood/issues/18#issuecomment-4860635622 on the constraint-side hypotheses). The inner-product opening is recovered from an accepting transcript tree by the IPA's special-soundness rewinding (`ipa_soundV`), binding-free; commitment binding is modeled separately as a discrete-log-relation hardness reduction (`relation_of_collision`, `commitmentBinding_iff_no_relation`). The circuit constraint follows from the verifier's gate point-check at `x`, lifted to the polynomial identity by Schwartz–Zippel.

**Relation to the verifier.**

| This formalization | Halo2 / the deployed verifier |
|---|---|
| `assemble`, `eval_assembleFinalMsm` | the verifier's MSM assembly and its closed-form evaluation |
| `compute_s` / `compute_b` | IPA `compute_s` / `compute_b` |
| `DeployedAccepts` | the final `MSM = identity` accept check |
| `ipa_soundV` | IPA special-soundness extraction |
| `relation_of_collision`, `commitmentBinding_iff_no_relation` | commitment binding as a DLR-hardness reduction |
| `orchard_verifier_sound_deployed_opening` / `_constraint` (+ Vesta) | conditional soundness composition (scaffold; hypotheses in the docstrings) |

**Status.** Builds green on the pinned toolchain; no `sorry`, no declared `axiom` (the fixture match uses `native_decide`, which rests on compiler trust via `Lean.ofReduceBool`). The result is a conditional verifier-soundness composition with the VK as input, resting on the named assumptions above (Fiat–Shamir/Blake2b as a random oracle, commitment binding as DLR hardness, IPA special-soundness rewinding, Schwartz–Zippel exclusions, the Vesta group order, VK-correctness). Lifting it to full Orchard SNARK soundness needs §3 — deriving the circuit and VK in Lean — as the human summary notes.


TalDerei:
lifting @daira's review feedback from my fork https://github.com/TalDerei/ironwood/pull/1, which this PR addresses and prunes from the commit history. 

-----------

[Written with Claude's assistance but heavily edited.]

Overall, excellent work, and great on avoiding overclaiming: the `_conditional` family is labelled as scaffolds, every assumption is named, superseded lemmas are flagged, and the input-side VK-correctness / output-side gaps are called out.

The Fiat–Shamir modelling cleanly isolates the random-oracle hand-wave (`hch`) and the forking bridge (`hFS`); good.

*However*, it is a problem that `AugmentedBinding` is unsatisfiable in a prime-order group. That means all of the conclusions instantiated at Vesta are vacuous.

`AugmentedBinding g U W` (`IpaUWS.lean`) requires the $n+2$ elements `g₀, ..., g_{n-1}, U, W` to be $\mathbb{F}_p$-linearly **independent**:

```lean
∀ a α β, commitGen g a + α • U + β • W = 0 → a = 0 ∧ α = 0 ∧ β = 0
```

For `[Module Fp G]` with `G` a prime-order group, `G ≅ Fp` is 1-dimensional, so *any* ≥2 elements are linearly dependent. Hence `AugmentedBinding` cannot hold for any $n \geq 0$ (even `U`, `W` alone). When the live capstones are instantiated at Vesta (`Vesta.lean`'s `hbind : ∀ {m} g', AugmentedBinding g' srs.u srs.w`), the hypothesis is unsatisfiable, so the Vesta verifier-soundness theorems hold only vacuously. That is, the conclusions are empty at the concrete curve and place no constraint on the actual verifier.

The AGM framing doesn't rescue this: the AGM constrains the *adversary*, but a theorem instantiated at Vesta still needs `hbind` satisfiable, and information-theoretic independence of the generators is false in the group regardless.

I think you may have recognized this and started on a fix. The "Binding as a discrete-log-relation hardness assumption" section models `CommitmentBinding` as DLR hardness through a proven reduction (`relation_of_collision` and `commitmentBinding_iff_no_relation`) —the same shape as the binding-signature argument it cites— and its own docstring notes it is "not on the live soundness path — that path models binding via `AugmentedBinding`".

The question is whether we should commit the work that relies on `AugmentedBinding`. I would prefer not to. We're all heavily using AI, and current AIs tend toward verbosity and reasoning biases. My experience is that if you show them a wrong approach and say it's wrong, they will focus on avoiding that approach (which is its own subtle bias) and over-mention it, which leads to out-of-context comments for human readers. They can't just ignore it, or at least it is more work than it should be to redirect their attention.

So, I think we should strip out the parts that rely on `AugmentedBinding` for now (they can be moved to another draft PR if you prefer).

The work to do next to restore the intended conclusions would be to migrate the live path to use a similar style of reduction: extend `relation_of_collision` to account for `U` and `W` and to route `AugmentedBinding.separate`, `deployed_flat_split`, the `_agm`/`_pinned`/`_closed` capstones, and the Vesta wrappers through that, so that soundness is expressed as a reduction to DLR hardness rather than an assumption of an unsatisfiable independence predicate. This is a migration along the template you've established rather than a redesign, though `AugmentedBinding.separate` is used widely enough for it to be a real chunk of work.

----

On the Vesta group order: a comment in `Vesta.lean` says "pinning it is point-counting (Schoof / Hasse)". However, checking a hypothesis for the order of a concrete elliptic curve group —for which you have a verified scalar multiplication— does not require point counting. Hasse is indeed relevant though. In the prime-order case, if we want to prove that the curve has order $q$ (verified prime), then we show that it has a non-identity point, say $P$, for which $q • P$ is the identity. For prime $q$ this implies that $q$ divides the group order, and then by Hasse's theorem (for $q$ sufficiently large, so that $2q$ is above the upper Hasse bound) we know that the group order cannot be a nontrivial multiple of $q$, so it must be $q$. This doesn't need point counting or a proof of any point-counting algorithm's correctness (well, pedantically it is a point-counting algorithm if you guess $q$ :-).

I was originally thinking that Mathlib would have Hasse's theorem proven and we could use it via the bridge to Mathlib's `WeierstrassCurve`, but apparently not. A worthy yak to shave :-) But I'm personally happy to take Hasse's theorem as an axiom for now (we actually only need to upper-bound the order as $< 2q$ which might be easier to prove, I don't know).

Minor: There are a couple of comments incorrectly mentioning BLAKE3 as being used for the txid/auth digest. I've given suggestions to fix that inline.

Zcash/Snark/Field.lean
```diff
@@ -0,0 +1,40 @@
+import Mathlib
+import CompElliptic.Fields.Pasta
+
+/-!
+# The verifier's scalar field `F_p`
+
+An Orchard proof is a halo2 proof over the Vesta curve (the README's `E_q`): commitments are Vesta
```
daira: Does this refer to a README that is not in the PR?

Zcash/Snark/Field.lean
```diff
@@ -0,0 +1,40 @@
...
+points, and challenges and evaluations are elements of Vesta's scalar field `F_p`. By the Pasta
+construction `F_p = vesta::Scalar = pallas::Base = ZMod p`, where `p = PALLAS_BASE_CARD ≈ 2²⁵⁴`. The
+proof system lives over Vesta, so the field is the Pallas *base* order `PALLAS_BASE_CARD` — not the
+Pallas *scalar* order `PALLAS_SCALAR_CARD` of the value commitment.
+
+`F_p` is where the fingerprint MSM's scalars live. The development stays generic over `[Field F]`;
+this module just pins the instantiation `F_p = ZMod p` and records `|F_p| = p`, the cardinality the
+Schwartz–Zippel bound (`Zcash.Snark.Fingerprint`) divides by. We use `F_p` abstractly; its primality
+is CompElliptic's certified fact, and the Pasta arithmetic is hand-waved.
```
daira: It's not so much hand-waved, as that you don't need it for the core of the correctness argument. You do need that the group has the correct order. But now that we have the group order proven (conditional on the Hasse bound holding for Vesta) in CompElliptic, we should be able to plug that gap.

daira: Done!

TalDerei: was loosely conflating things that were hand-waved with assumptions; fixing per resolution in https://github.com/zcash/ironwood/pull/10.

Zcash/Snark/Field.lean
TalDerei: yeah removing these kinds of dangling external references.

Zcash/Snark/Fingerprint/Batch.lean
```diff
@@ -0,0 +1,66 @@
...
+Batch verification combines the per-proof MSM evaluations `gᵢ` with fresh random scalars and accepts
+only if the combination is the group identity. Soundness: if any `gᵢ ≠ 0` (some proof's fingerprint
+is nonzero), the random combination is the identity only with negligible probability.
+
+The atomic fact is `smul_eq_zero_iff_of_ne_zero`: a random scalar annihilates a nonzero group element
+only when it is `0`. `batch_rlc_card_le` amplifies it to the `N`-fold linear combination `∑ rᵢ • gᵢ`:
+if some `gⱼ ≠ 0`, the combination vanishes for at most `p ^ (N-1)` of the `p ^ N` choices of
+randomness — a `1 / p` fraction, negligible for `p ≈ 2²⁵⁴`. (halo2's deployed combiner folds with
+iterated scaling, a degree-`(N-1)` variant bounded by `(N-1) / p` through the same atomic fact; we
+model the linear-combination form, whose `1 / p` is the cleaner statement of the same soundness.)
```
daira: The wording is accurate, but "iterated scaling" reads easily as *powers of a single challenge* (the polynomial-universal-hash / Reed–Solomon-fingerprinting technique the in-protocol multiopen/IPA/gate batching uses). batch.rs instead uses a fresh independent random per fold, so disambiguating here avoids the conflation:

```suggestion
randomness — a `1 / p` fraction, negligible for `p ≈ 2²⁵⁴`.

halo2's deployed combiner folds with iterated scaling, a degree-`(N-1)` variant bounded by
`(N-1) / p` through the same atomic fact. We model the linear-combination form, whose `1 / p`
is the cleaner statement of the same soundness. Note that iterated scaling uses a fresh
*independent* random per fold (`acc.scale(random); acc.add_msm(m)`). This should not be
confused with folding using powers of a single challenge, which is used elsewhere in the
argument of knowledge and formalized in `Constraints.lean`.

Posted with the assistance of Claude Opus 4.8

Comment (TalDerei) [reply Zcash/Snark/Fingerprint/Batch.lean]: that's a good observation, clarifying this distinction.

File: Zcash/Snark/Soundness/CommitFold.lean [hunk 4]
```diff
@@ -0,0 +1,149 @@
...
+Rather than assuming the commitment is binding outright, the binding reduction models it as a reduction
+to a hardness assumption — the same shape as the binding-signature argument's `relation_of_imbalance`:
+the counterfactual that breaking binding produces a nontrivial discrete-log relation among the SRS
+generators (`relation_of_collision`), against the discrete-log-relation (DLR) hardness assumption — no
+such relation exists on the Vesta generators. `commitmentBinding_iff_no_relation` makes precise that
+`CommitmentBinding` is exactly DLR hardness, so the assumption is the standard, named one, with the
+reduction itself proven. -/
+
+/-- A discrete-log relation among the SRS generators: a coefficient vector the generators send to `0`.
+It is nontrivial when `r ≠ 0`. DLR hardness is the assumption that no nontrivial relation exists. -/
```
daira:
```suggestion
It is nontrivial when `r ≠ 0`. DLR hardness is the assumption that no feasible adversary can find a
nontrivial relation. -/

File: Zcash/Snark/Soundness/Constraints.lean
Comment (daira) [inline Zcash/Snark/Soundness/Constraints.lean:1]: Commit e8d46c8bee960507ca2990fc18bed68b57a9944f says "so circuit satisfaction is derived from the verifier's actual check, not assumed", but this is an over-claim. The argument only covers gate satisfaction directly; it folds in the checks for the permutation and lookup arguments but does not connect those to the circuit-level copy and lookup constraints, which are part of circuit satisfaction (#14).

Comment (TalDerei) [reply Zcash/Snark/Soundness/Constraints.lean:1]: good catch! opted to only do gate satisfaction for now because something like #14 seemed pretty convoluted at the time for a first pass.

File: Zcash/Snark/Soundness/KnowledgeSoundness.lean [hunk 5]
```diff
@@ -0,0 +1,126 @@
...
+structure SnarkRelation (srs : SRS G) (P : G) (b : Fin (2 ^ srs.k) → Fp) (v : Fp)
+    (circuitSat : (Fin (2 ^ srs.k) → Fp) → Prop) (a : Fin (2 ^ srs.k) → Fp) : Prop where
+  opens : IpaRelation srs P b v a
+  satisfiesCircuit : circuitSat a
+
+/-- The intended concrete instantiation of `SnarkRelation.circuitSat`: the witness's decoded columns
+satisfy the `y`-combined gates' quotient identity. `decodeAdvice`/`decodeInstance` map the IPA witness to
+the circuit's column polynomials (the multiopen decode — abstract until the assembly is modeled), and
+`combineGates` is the `y`-combination of the gate `Expr`s (`eval_combineGates`). So
+`circuitSat := circuitSatViaGates …` reads "the extracted witness satisfies the circuit's constraints",
```
daira: "satisfies the circuit's gate constraints"

Zcash/Snark/Soundness/KnowledgeSoundness.lean
```diff
@@ -0,0 +1,126 @@
...
+def circuitSatViaGates {k : ℕ} (fixedCols : ℕ → Polynomial Fp)
+    (decodeAdvice decodeInstance : (Fin (2 ^ k) → Fp) → (ℕ → Polynomial Fp))
+    (y : Fp) {ng : ℕ} (gates : Fin ng → Expr Fp) (hpoly : Polynomial Fp) (deg : ℕ)
+    (a : Fin (2 ^ k) → Fp) : Prop :=
+  combineGates fixedCols (decodeAdvice a) (decodeInstance a) y gates = hpoly * (X ^ deg - 1)
+
+/-- `circuitSat`, derived from the deployed constraint check (not assumed). With the numerator the gate
+combination `combineGates …`, an accepting quotient check at the challenge `x` plus a good challenge
+(`x` not a root of the constraint difference — its complement the `≤ deg/p` Schwartz–Zippel bad set) gives
+`circuitSatViaGates …`: the witness's decoded columns satisfy the gates. So the circuit-satisfaction
```
daira: "circuit gate-satisfaction"

Zcash/Snark/Soundness/KnowledgeSoundness.lean
```diff
@@ -0,0 +1,126 @@
...
+    (a : Fin (2 ^ k) → Fp) (x : Fp)
+    (hcheck : quotientCheck
+      (combineGates fixedCols (decodeAdvice a) (decodeInstance a) y gates) hpoly deg x)
+    (hgood : combineGates fixedCols (decodeAdvice a) (decodeInstance a) y gates ≠ hpoly * (X ^ deg - 1) →
+      (combineGates fixedCols (decodeAdvice a) (decodeInstance a) y gates
+        - hpoly * (X ^ deg - 1)).eval x ≠ 0) :
+    circuitSatViaGates fixedCols decodeAdvice decodeInstance y gates hpoly deg a :=
+  constraint_identity_of_accept _ hpoly deg x hcheck hgood
+
+/-- Knowledge soundness (modulo its hypotheses). Given a consistent transcript tree `t` for `a`
```
daira: "modulo its hypotheses and current modelling gaps"

Zcash/Snark/Soundness/Main.lean
```diff
@@ -0,0 +1,187 @@
...
+`hcirc`. -/
+
+open Polynomial in
+/-- The deployed Orchard verifier is sound, opening and constraint both derived. Both conjuncts of
+`SnarkRelation` are derived, not assumed: `IpaRelation` (the opening) from the transcript tree via
+`ipa_soundV`; and `circuitSat` — instantiated to `circuitSatViaGates` — from the verifier's quotient/gate
+point check `hquot` at the challenge `x`, lifted to the polynomial identity by Schwartz–Zippel (`hgood`),
+via `circuitSatViaGates_of_check`.
+
+The black-box `hcirc` is gone. What remains are the verifier's actual checks and the standard assumptions:
```
daira: Comment referring to something that no longer exists in the code; also on line 148. (I had to tell my Claude not to do that; it has a lot of recency bias by default.)

Zcash/Snark/Soundness/Constraints.lean
TalDerei: fixed in latest commit, alongside the other suggestions.

daira (approved):
utACK https://github.com/zcash/ironwood/compare/9caab5ea79db31567cb2b80fc10f8d27c7e57c96..b809e82ef5cff1d9b365bc5ca1b5ac1f9cf89d97

PR #10, opened 2026-06-27
fv: derive the Vesta group order from Hasse via CompElliptic

daira:
Discharges the Vesta group-order assumption in the verifier soundness, building on `fv/orchard-verifier`.

## What

The Vesta capstones assumed `Fact VestaOrder` (every point is `scalarFieldOrder`-torsion) as a bespoke hypothesis. CompElliptic now proves the Pasta curve group orders from Hasse's bound (`Curves/PastaOrder`), so this derives the order rather than assuming it:

- `vestaOrder_of_hasse : HasseBound Vesta.curve → VestaOrder`, from `Pasta.Vesta.card_eq` (`Nat.card VestaG = scalarFieldOrder`) and Lagrange (`addOrderOf_dvd_natCard`).
- A conditional bridging instance supplies `Fact VestaOrder` —hence the `Fp`-module— from `Fact (HasseBound Vesta.curve)`.
- The three Vesta capstones now take `[Fact (HasseBound Vesta.curve)]` in place of `[Fact VestaOrder]`. This changes their public signatures, but it is strictly a weaker assumption: the order follows from Hasse, so callers assume less.
- `Group.lean`'s note is updated: the `Module (ZMod p)` instance on the concrete Vesta curve now lands (conditioned on Hasse), so `G` can be instantiated there. The fingerprint argument stays abstract by choice.

Bumps the CompElliptic dependency `74a3531 → 6c5b1a6` (which adds `Curves/PastaOrder`).

## Why

Replaces an opaque "trust this 254-bit number is the order" assumption with the single named Hasse hypothesis — the one gap Mathlib still lacks. The witness point, the primality of the order, and the field-size facts are all discharged inside CompElliptic.

## Axiom footprint

`#print axioms` on the three capstones (and `vestaOrder_of_hasse`): `[propext, Classical.choice, Quot.sound]` plus four `native_decide` axioms from CompElliptic's concrete Vesta witness facts (`Gpt`, `Gpt_ne_zero`, `baseCard_nsmul_Gpt`, `card_eq`). No `sorry`, no general `axiom`. The `native_decide` is confined to closed per-curve facts (CompElliptic's "independently re-checkable trust" principle) — the trade for dropping the order assumption.

🤖 Claude Opus 4.8


daira (commented):
Self-ACK

TalDerei:
ACK, thanks for integrating this! this was stubbed as one of the stated assumptions in the PR description referencing upstream https://github.com/daira/CompElliptic/pull/2, so will update accordingly.

Issue #11, opened 2026-06-28
Discharge the Fiat-Shamir assumption (random-oracle model + forking/rewinding)


TalDerei:
The conditional soundness result is proven over the *interactive* verifier (challenges as free inputs) and assumes Fiat-Shamir as a black box: `squeeze` is an uninterpreted function, and the bridge from "accepting proof" to "special-soundness transcript tree" (`FiatShamirTree`) is taken as a hypothesis rather than proven.

This issue tracks making that step rigorous: give the Fiat-Shamir hash real random-oracle semantics and prove the forking/rewinding argument that produces the transcript tree the extractor consumes, so the non-interactive (deployed) verifier inherits the interactive verifier's soundness rather than assuming it.

**Scope (broad):**
- Model the challenge derivation under a random oracle instead of an opaque function.
- Match the deployed Halo2 transcript transition exactly, including `BLAKE2B_PREFIX_CHALLENGE` domain separation and no challenge self-absorption after a squeeze.
- Prove the forking/rewinding lemma yielding the accepting-transcript tree (including the Schwartz-Zippel good-challenge exclusions that fall out of challenge uniformity).
- Replace the assumed bridge with the proven result and re-instantiate the deployed soundness theorems.

daira:
Cross-reference of the Fiat–Shamir transcript order — Lean `deriveChallenges` ↔ the deployed halo2 Rust verifier. Net: every absorb/squeeze step matches **except** one concrete, actionable ordering bug (**D1**, lookup permuted commitments) plus the two modelling gaps this issue already names (**D2** challenge self-absorption, **D3** domain-separation prefixes). Intent: tag each step in both sources so the mapping stays pinned. *(Updated: Lean links now point at `Zcash/Snark/Verifier/FiatShamir.lean` after the Core/Verifier/Fingerprint/Soundness reorg.)*

---

# Fiat–Shamir transcript order: Lean ↔ Rust cross-reference

Cross-references the Lean Fiat–Shamir schedule (`deriveChallenges`) against the deployed halo2 Rust verifier, step by step. Purpose: give every absorb/squeeze step a unique tag to embed as a comment in both the Rust and Lean source, so the correspondence is checkable mechanically and stays pinned as either side changes.

**References**
[...]

PR #19, opened 2026-06-28
fv: fix Fiat-Shamir lookup permuted-commitment absorb order (D1)

daira:
Fixes the lookup permuted-commitment absorb order in the Fiat–Shamir schedule (`deriveChallenges`).

`deriveChallenges` absorbed the lookup permuted commitments as **all permuted-input commitments, then all permuted-table commitments** (`absorbPoints2 lookupPermutedInput ++ absorbPoints2 lookupPermutedTable`). The deployed halo2 verifier instead reads, **per lookup, the input commitment then the table commitment** — [`lookup/verifier.rs:43-44`](https://github.com/zcash/halo2/blob/261faaccd5a30c19bb8468600841a0dac305adf5/halo2_proofs/src/plonk/lookup/verifier.rs#L43-L44) (`read_permuted_commitments`), looped per lookup at [`plonk/verifier.rs:134`](https://github.com/zcash/halo2/blob/261faaccd5a30c19bb8468600841a0dac305adf5/halo2_proofs/src/plonk/verifier.rs#L134) — i.e. `input₀, table₀, input₁, table₁, …`. The two orders differ whenever `numLookups ≥ 2` (Orchard has `numLookups = 3`), so `deriveChallenges` did not reproduce the deployed transcript.

The fix adds `absorbLookupPermuted` (interleave per proof, per lookup: input then table) and uses it in the θ→β step.

**Scope.** Only the Fiat–Shamir schedule (`deriveChallenges` / `nonInteractiveFingerprint` / `FixtureFiatShamir`) is affected. `Fixture.fingerprint_matches` uses the *captured* challenges and is unchanged, so this was latent (the assumption `deriveChallenges fs init ps = ch` is what was wrong, undetectable while `fsHash` is abstract). This is **D1** from the transcript cross-reference posted on #11 — the concrete, self-contained instance of that issue's point 2 ("match the deployed transcript exactly"). D2 (challenge self-absorption) and D3 (domain-separation prefixes) remain part of the broader #11 work.

Builds: `lake build Zcash.Snark.Verifier.FiatShamir` green.

🤖 Claude Opus 4.8

### DIFF
--- a/Zcash/Snark/Verifier/FiatShamir.lean
+++ b/Zcash/Snark/Verifier/FiatShamir.lean
@@ -58,6 +58,13 @@ def absorbPoints2 {F G : Type*} {a b : ℕ} (f : Fin a → Fin b → G) : List (
 def absorbScalars2 {F G : Type*} {a b : ℕ} (f : Fin a → Fin b → F) : List (TranscriptElt F G) :=
   (List.ofFn (fun i => absorbScalars (f i))).flatten
 
+/-- Absorb the lookup permuted commitments in the deployed order (halo2 `read_permuted_commitments`):
+per proof, per lookup, the permuted-input commitment then the permuted-table commitment. -/
+def absorbLookupPermuted {F G : Type*} {a b : ℕ} (input table : Fin a → Fin b → G) :
+    List (TranscriptElt F G) :=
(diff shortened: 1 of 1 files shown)


TalDerei (approved):
ACK

Issue #14, opened 2026-06-28
Formalize soundness of the permutation and lookup arguments


daira:
The verifier-soundness chain in #9 currently reaches the **gate constraints only**: `circuitSatViaGates` / `SnarkRelation` concludes that the extracted witness's decoded columns satisfy the gates, and `Constraints.lean`'s `combineGates` folds only `gates : Fin n → Expr` into the Schwartz–Zippel-bounded numerator.

The **permutation** (copy-constraint) and **lookup** (table-membership) arguments are *assembled and fingerprinted* — their constraint expressions are recomputed in `Expressions.lean` (`permutationExpressions` / `lookupExpressions`) and folded into the verifier's `expected_h_eval` — but their **soundness as arguments** is not yet proven:

- **Permutation** ([design](https://zcash.github.io/halo2/design/proving-system/permutation.html)): that the grand-product identities (with challenges `β`, `γ`) imply the intended copy-constraint / permutation relation on the advice cells.
- **Lookup** ([design](https://zcash.github.io/halo2/design/proving-system/lookup.html)): that the permuted-column + grand-product identities imply multiset / table membership.

Concretely missing:
1. §2 lifts only the gate polynomial identity to the witness (`circuitSatViaGates_of_check`); the permutation/lookup polynomial identities are not bridged to their combinatorial relations.
2. `β` and `γ` are carried in `Challenges` but used in no soundness argument — yet they are exactly the randomness those two arguments' soundness rests on.

In scope for the formalization: the full Orchard circuit relation includes copy constraints and lookups, and the VK-correctness assumption presumes the VK encodes the *whole* relation. Tracking separately from #9.

🤖 Claude Opus 4.8 (1M context)

PR #21, opened 2026-06-28
fv: permutation & lookup argument soundness + permutation construction (towards #14)

daira:
Towards #14 — soundness of the permutation and lookup *arguments* (the gates-only gap left open by #9), plus correctness of the permutation *construction*. All proofs are complete: no `sorry`, no new `axiom`; builds green.

## The idea

Both the permutation and lookup arguments enforce their relation by a **grand product of factors linear in the challenges**, and the soundness step is identical: *such a product, equal at the random challenge, forces the underlying multiset identity*. So rather than tackle the two arguments' machinery separately, this PR isolates that shared core as a reusable kernel, then builds each argument's structural step on top, and finally verifies (up to order within each cycle) the construction that produces the permutation in the first place.

### `Soundness/GrandProduct.lean` — the shared kernel
Over Mathlib `Polynomial` (these are soundness facts — nothing computes — leaning on Mathlib's roots / unique-factorization theory):
- `prod_X_add_u_inj` — `∏ (X + uᵢ) = ∏ (X + wᵢ)` over an integral domain ⟹ `{uᵢ} = {wᵢ}`. Products represent multisets *injectively* (factor theorem over a domain). The load-bearing fact.
- `card_eval_prod_eq_le` — univariate Schwartz–Zippel at a point: for `s ≠ t`, the challenges `β` where the *field* products `∏ (xᵢ + β)` collide are roots of the (nonzero) difference polynomial — a bad set of size `≤ max |s| |t|`.
- `prod_pair_inj` — the **pair** version `∏ (vᵢ + nᵢ·β + γ) = … ⟹ {(vᵢ,nᵢ)} = {(cⱼ,dⱼ)}`, obtained by running `prod_X_add_u_inj` over `R = F[β]` (variable `γ`) and reading off the two coefficients. This is what the permutation argument needs (matching `(value, name)` *pairs*); the lookup argument uses `prod_X_add_u_inj` twice with independent `β`, `γ`.

### `Soundness/Lookup.lean` — lookup run-structure
`run_structure`: the permuted-column step toward `{input} ⊆ {table}`. Faithful to the constraint system — `a₀ = s₀`, the row-0 instance `a₀ = s₀ ∨ a₀ = aPrev` with the wrapped predecessor `aPrev` arbitrary, and `a_{i+1} = s_{i+1} ∨ a_{i+1} = a_i` — so every `A′` value coincides with some `S′` value.

### `Soundness/Permutation.lean` — permutation argument structural soundness
- `perm_values_eq_iff` — the multiset of `(value, label)` pairs is invariant under permuting labels by `σ` iff `value c = value (σ c)` for all `c`. With `label` injective (δ-coset name distinctness), this turns the kernel's multiset-of-pairs identity (`prod_pair_inj`) into the per-cell relation.
- `value_const_on_sameCycle` — lifts that one-step invariance to all cells in a `σ`-cycle (`value c = value (σ c)` is *not* the same as "constant on cycles"; the orbit lift iterates over `σ.SameCycle`).
- `perm_copy_constraints` — composes the two: multiset-of-pairs identity + injective `label` + `σ.SameCycle c d` ⟹ `value c = value d`, the copy-constraint relation.

### `Soundness/PermutationConstruction.lean` — correctness of the construction
Verifies the union-find algorithm in halo2's `plonk/permutation/keygen.rs` (`Assembly::copy`) that builds `σ` from the circuit's copy constraints.
- `sameCycle_mul_swap` (the crux — **no Mathlib equivalent**): composing `π` with the transposition `(a b)` *merges* the cycles of `a` and `b` when they're in different cycles, and would *split* a cycle otherwise — which is exactly why the algorithm's "already in the same cycle" skip is essential (cf. the design doc's "Broken alternatives"). The whole combinatorial content reduces to one orbit-walk fact (`sameCycle_mul_swap_self`, via `minimalPeriod`); the rest is relational algebra (`MergeRel` equivalence, zpow-closure subset, monotonicity).
- `build_correct`: two cells share a `σ`-cycle iff the copy constraints force them equal (the `Relation.EqvGen` closure of the pairs).
- `value_eq_of_constraints` / `value_eq_of_mem` close the loop: a verifier-accepting multiset identity for `σ = build cs` plus a copy constraint forcing `x ≡ y` gives `value x = value y` — the verifier enforces equality on exactly the cells the circuit declared equal.

## Known gaps

* In `PermutationConstruction.lean`, the disjoint-set `aux` / `sizes` arrays are modelled abstractly: the "same cycle?" test is modelled by `SameCycle` directly. A bug in updating `aux`, or in failing to map through `aux` in that test, could affect soundness; proving the imperative `aux`-walk maintains `aux(x) = aux(y) ↔ SameCycle x y` is separate loop-invariant reasoning, not yet done. The `sizes` (union-by-size) choice is a performance optimization that does not affect the resulting partition.
* `build` reproduces the cycle *partition* exactly — which is all the copy-constraint soundness needs — but not necessarily halo2's exact *permutation*. The order in which `copy` calls are made determines the order of each cycle up to rotation, and hence the permutation-polynomial commitments in the verifying key; reproducing the deployed commitments would additionally require modelling that order.

## Still to come (the remaining wrappers)

- The **grand-product ⟹ multiset identity** step for both arguments: telescoping the running product over the usable rows, with the boundary / blinding-row rules (`l₀(1−z)`, `z²−z`, inter-chunk linking), feeding the multiset identity that `prod_pair_inj` / `prod_X_add_u_inj` consume.
- The **lookup final assembly** (`{input} ⊆ {table}` from the kernel ×2 + `run_structure`).
- The **Schwartz–Zippel bridge** from these polynomial identities to the verifier's actual field-element product check (`card_eval_prod_eq_le` is the univariate seed; the multivariate version reuses `Fingerprint/SchwartzZippel.lean`), tying into the FS / good-challenge work in #11/#12.

🤖 Claude Opus 4.8

### DIFF
--- a/Zcash/Snark/Soundness/Lookup.lean
+++ b/Zcash/Snark/Soundness/Lookup.lean
@@ -1,0 +1,58 @@
+import Mathlib
+import Zcash.Snark.Soundness.GrandProduct
+
+/-!
+# Lookup argument soundness
+
+(Towards #14.) The [lookup argument](https://zcash.github.io/halo2/design/proving-system/lookup.html)
+proves multiset inclusion {input} ⊆ {table}. Its soundness combines the grand-product → multiset kernel
+(`GrandProduct.lean`, giving {A′} = {input} and {S′} = {table}) with the **permuted-column run-structure**
+proved here: the verifier's ℓ_0·(A′ − S′) and active·(A′ − S′)·(A′ − A′_prev) constraints force every A′
+value to coincide with some S′ value, which together with the multiset equalities yields {input} ⊆ {table}.
+
+This file currently provides `run_structure` (the run-structure step over the usable rows); the
+grand-product ⟹ multiset equalities and the final assembly will follow.
+-/
+
+namespace Zcash.Snark
+
+/-- **Run-structure for the lookup argument**. This verifies part of the abstract argument, not yet
+the mapping to constraints.
+
+We allow the prover to supply permutation columns of A and S — called A′, S′ in the Halo 2 book, and
+`a`, `s` here. As in the book, ℓ_i evaluates to 1 at row i and 0 otherwise, and the running-product
+constraint refers to the previous row A′_prev = A′(ω⁻¹·), which at row 0 wraps cyclically to the last
+blinding row, an arbitrary value `aPrev`.
+
+The constraints checked by the verifier are:
+
+* `a_0 = s_0`, from ℓ_0·(A′ - S′) = 0.
+* `a_0 = s_0 ∨ a_0 = aPrev`, from active·(A′ - S′)·(A′ - A′_prev) = 0 at row 0, with `aPrev` arbitrary.
+* `a_{i+1} = s_{i+1} ∨ a_{i+1} = a_i`, the later-row instances of the same constraint.
+
+The first two together are equivalent to just `a_0 = s_0` for any `aPrev` (the wrapped predecessor is
+harmless), so row 0 is matched. With every later row matched or A′ repeating its predecessor, every
+`a_i` equals some `s_j`. So the permuted-input column A′ is elementwise contained in the permuted
+table S′ — the membership the lookup argument needs.
+
+Indices are `Fin (n + 1)` so the index set is nonempty (row 0 always exists), and the proof uses
+`Fin.induction` so the `succ` case's hypothesis is exactly the predecessor row `i.castSucc`. -/
+theorem run_structure {F : Type*} {n : ℕ} (a s : Fin (n + 1) → F) (aPrev : F)
+    (h0 : a 0 = s 0)
+    (hstep0 : a 0 = s 0 ∨ a 0 = aPrev)
+    (hstep : ∀ i : Fin n, a i.succ = s i.succ ∨ a i.succ = a i.castSucc) :
+    ∀ i, ∃ j, a i = s j := by
+  intro i
+  induction i using Fin.induction with
+  | zero =>
+    -- the row-0 running constraint with the arbitrary wrapped predecessor is subsumed by `a 0 = s 0`
+    rcases hstep0 with h | _
+    · exact ⟨0, h⟩
+    · exact ⟨0, h0⟩
+  | succ i ih =>
+    rcases hstep i with hm | hr
+    · exact ⟨i.succ, hm⟩
+    · obtain ⟨j, hj⟩ := ih
+      exact ⟨j, hr.trans hj⟩
+
+end Zcash.Snark
--- a/Zcash/Snark/Soundness/Lookup.lean
+++ b/Zcash/Snark/Soundness/Lookup.lean
@@ -4,11 +4,12 @@ import Zcash.Snark.Soundness.GrandProduct
 /-!
 # Lookup argument soundness
 
-(Towards #14.) The [lookup argument](https://zcash.github.io/halo2/design/proving-system/lookup.html)
-proves multiset inclusion {input} ⊆ {table}. Its soundness combines the grand-product → multiset kernel
-(`GrandProduct.lean`, giving {A′} = {input} and {S′} = {table}) with the **permuted-column run-structure**
-proved here: the verifier's ℓ_0·(A′ − S′) and active·(A′ − S′)·(A′ − A′_prev) constraints force every A′
-value to coincide with some S′ value, which together with the multiset equalities yields {input} ⊆ {table}.
+The [lookup argument](https://zcash.github.io/halo2/design/proving-system/lookup.html) proves that
+every input value occurs in the table — set inclusion {input} ⊆ {table}, ignoring multiplicity. Its
+soundness combines the grand-product → multiset kernel (`GrandProduct.lean`, giving {A′} = {input} and
+{S′} = {table} as multisets) with the **permuted-column run-structure** proved here: the verifier's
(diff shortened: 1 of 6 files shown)


TalDerei:
I think the shared kernel lemmas here + wrappers are a reasonable (_probably_ the optimal) decomposition, and don't have a preference on the other design notes.

Zcash/Snark/Soundness/PermutationConstruction.lean
```diff
@@ -0,0 +1,328 @@
...
+  if π.SameCycle ab.1 ab.2 then π else π * swap ab.1 ab.2
+
+/-- Run the algorithm over a list of copy constraints, starting from the identity permutation (every
+cell its own 1-cycle). The halo2 implementation keeps the cycle partition in a disjoint-set structure
+(`aux` / `sizes`) purely to make the `SameCycle` test and the merge efficient; the resulting
+permutation — hence its cycle partition — is exactly this. The order in which constraints are processed
+does not affect the final cycle partition (it is the equivalence closure either way; `build_correct`). -/
+def build : List (α × α) → Perm α
+  | [] => 1
+  | ab :: rest => step (build rest) ab
```
TalDerei, line 243: similar to [D1](https://github.com/zcash/ironwood/issues/11#issuecomment-4823812750) (inconsistent FS transcript absorption order), flagging that this build doesn't model halo2's copy call order.

Zcash/Snark/Soundness/Lookup.lean
```diff
@@ -0,0 +1,58 @@
...
+
+(Towards #14.) The [lookup argument](https://zcash.github.io/halo2/design/proving-system/lookup.html)
+proves multiset inclusion {input} ⊆ {table}. Its soundness combines the grand-product → multiset kernel
+(`GrandProduct.lean`, giving {A′} = {input} and {S′} = {table}) with the **permuted-column run-structure**
+proved here: the verifier's ℓ_0·(A′ − S′) and active·(A′ − S′)·(A′ − A′_prev) constraints force every A′
+value to coincide with some S′ value, which together with the multiset equalities yields {input} ⊆ {table}.
+
+This file currently provides `run_structure` (the run-structure step over the usable rows); the
+grand-product ⟹ multiset equalities and the final assembly will follow.
+-/
```
TalDerei, line 16: nit; "multiset inclusion" is a misnomer here.

TalDerei (commented):
some consistency observations against the native implementation.

Zcash/Snark/Soundness/PermutationConstruction.lean
```diff
@@ -0,0 +1,328 @@
+import Mathlib
+import Zcash.Snark.Soundness.Permutation
+
+/-!
+# Correctness of the halo2 permutation construction
+
+(Towards #14.) The permutation `σ` that the permutation argument (`Permutation.lean`) operates on is
+*built* by halo2 from the circuit's copy constraints, so that its cycles are exactly the
+equality-constraint classes. This file verifies that construction.
```
TalDerei: nit; don't think we need to reference specific issues.

Zcash/Snark/Soundness/Lookup.lean
daira: You're right; multiset inclusion would mean that there are no more of a given element in input than there are in table. We only mean that all of the elements of input are in table, ignoring multiplicity.

Zcash/Snark/Soundness/PermutationConstruction.lean
daira: Right; the order in which constraints are processed does not affect the final *partition*, but it *does* affect how that partition is represented as a permutation. That includes the order of each cycle, and affects the permutation commitments; we want to reproduce the latter exactly.

Zcash/Snark/Soundness/PermutationConstruction.lean
```diff
@@ -0,0 +1,334 @@
...
+* if `left` and `right` are already in the same cycle, do nothing;
+* otherwise swap `mapping(left)` and `mapping(right)`, i.e. replace the permutation `π` by
+  `π * (left right)`.
+
+We model this as `build`, a fold of `step` over the constraint list starting from the identity. The
+halo2 `aux` / `sizes` arrays are a [disjoint-set](https://en.wikipedia.org/wiki/Disjoint-set_data_structure)
+structure that makes the same-cycle test and the cycle-walk efficient. How the same-cycle test is
+implemented —as long as it is correct— does not affect the resulting permutation, so we model the test
+by `SameCycle` directly. The size comparison (union by size) only chooses which cycle to relabel —a
+performance optimization that also does not affect the resulting permutation— so it is elided.
```
daira, line 28: ~~It does affect the resulting permutation, I think.~~ It doesn't affect the resulting *partition*.

I'm mistaken. It doesn't affect the permutation either, because "swap mapping(left) with mapping(right)" is symmetric in left and right.

TalDerei (approved):

Zcash/Snark/Soundness/PermutationConstruction.lean
daira: It occurs to me that it is not actually necessary to reproduce the order within cycles. It suffices, for both soundness and completeness, to provide the permutation output by the Rust code and verify that it represents the same partition as that output by the Lean model. That has the advantage of not having to define copy constraints in the exact-same order in the Lean code, which has advantages for abstraction and for stability of the model wrt circuit and optimization changes. (The latter is not critical for the Orchard circuits but is nevertheless a nice feature.)

PR #20, opened 2026-06-28
feat(fv): binding reduction

TalDerei:
**Human Summary:**

References follow-up work in https://github.com/zcash/ironwood/issues/13. 

Binding bridges the deployed verifier (`assemble`) with the abstract IPA soundness component, and it's modeled as a DLR-hardness reduction (emulating the same `HasNontrivialRelation` shape per https://github.com/zcash/ironwood/pull/4#discussion_r3415620886).

Recall, FS is currently treated as an assumed hypothesis in the soundness definition. The actual forking / rewinding to produce the transcript tree (`FiatShamirTree`) is assumed, but the gap this binding impl closes is the other side of that: once you have the tree (again, it's assumed), you can peel and separate the U/W/S (extra terms the verifier mixes into the MSM) to get a clean IPA transcript tree the extractor can operate on. That "peeling" piece was previously assumed, and is now proved. 

Out of scope: the rewinding/forking half of `FiatShamirTree` is #11. 

------------

**AI Summary:**

This PR puts commitment binding to work on the deployed soundness path. Mental
model: start from the real deployed MSM, peel off the extra `U`/`W`/`S`
machinery, and either recover the clean IPA opening or surface a discrete-log
relation.

## What Is Proven

- **Augmented binding**: a collision in the deployed `(g, U, W)` basis yields a
  nontrivial discrete-log relation.
- **Deployed-to-clean peel**: `DeployedIpaAcceptV` reduces to the clean IPA
  accept structure `ipa_soundV` consumes, with the opened commitment and value
  pinned to the proof's `multiopenCommitment` / `multiopenValue`.
- **Equation match**: `assembleFinalMsm` entails Halo2's explicit IPA verifier
  equation, in the direction soundness needs.
- The Vesta opening endpoints now conclude `SnarkRelation ∨ NontrivialRelation`,
  with the relation *computed as data* (`NontrivialRelation.ofUnopenedFork` /
  `ofDeployedTree`) rather than asserted — an `∃`-closed `HasNontrivialRelation`
  is vacuously true at a prime-order curve.

## Remaining Floor

The relation branch is the computational assumption: Lean *exhibits* a concrete
`NontrivialRelation` (as data), but ruling out an efficient adversary that
*finds* one is the DLR-hardness layer outside Lean (AGM handoff in #28).
`FiatShamirTree` is still assumed here — a data-producing `Type` yielding a
`ForkedTranscript` (`DeployedIpaVerifierEq → ForkedTranscript`), the residual
[...]

Issue #18, opened 2026-06-28
Bind the extracted IPA witness to the circuit columns via the multiopen decode


TalDerei:
The verifier opens all the circuit columns as one batched commitment, so the IPA extracts a single combined polynomial — not the individual columns. The gate check needs the columns, so the extracted witness has to be un-batched back into them.

Today `circuitSatViaGates` / `orchard_verifier_sound_deployed_constraint` (`Soundness/{KnowledgeSoundness,Main}.lean`) leave that un-batching as a free function (`decodeAdvice`/`decodeInstance`) — so the gate check runs on an abstract decode, not the prover's real columns.

Fix: recover each column for real, from `n` openings at distinct challenges via `batch_open_soundV` (`Soundness/IpaSoundness.lean`, already proven), so the gate check binds the actual columns.

This is the "witness → real columns" half of the `hquot` hypothesis; the other half (that the gate check itself comes from `assemble.eval = 0`) is #11/#13.

Note: the `x₁`/`x₄` rewinding this needs must also extend the round-by-round transcript-ordering treatment (#23, `TranscriptOrdering` in #26) to the multiopen squeeze points (`q′` absorbed before `x₃`, `u` before `x₄`) — #23/#26 cover the IPA rounds only.

TalDerei:
We should find other abstract free functions like this currently being used (don't think there is any).

TalDerei:
Additional follow-up surfaced in PR #24 review (the forking value/constraint-soundness gap, ≈ this issue):

- **Compose the per-hypothesis soundness budgets into the capstone probability, and restate the constraint side over the *extracted* witness.** The deployed forking capstones open the multiopen commitment at `v − ξ·⟨s,b⟩` (not the claimed `v`) and *assume* `hξ : ξ·⟨s,b⟩ = 0`; `hprob` samples only the IPA round challenges, not `ξ`/`z`/the gate challenge, and the standalone `1/p` exclusion lemmas (`blinder_shift_badSet_measure`, `blinder_value_recovery_badSet`) are never folded into the capstone measure. Composing them — and running the gate check on the columns recovered by the decode above rather than the free `decodeAdvice`/`decodeInstance` under the (documented) unsatisfiable `hquot`/`hgood` premise — is what turns the constraint capstones into deployed gate-level soundness.

PR #24, opened 2026-06-29
feat(fv): formalize Fiat-Shamir under RO and round-by-round soundness

### LINKED ISSUE #23: Track round-by-round Fiat-Shamir transcript soundness
Agreed that we should explicitly capture the round-by-round transcript soundness, as pointed out by @ValarDragon @daira. 

The current model initially treated challenges as free inputs by design, so the verifier-side algebra could be developed independently. The follow-up is to prove that the FS/RO model enforces the intended online ordering:

- each IPA round point `(L_j, R_j)` is absorbed before challenge `u_j` is derived;
- `u_j` is sampled from the transcript prefix containing that round point;
- later prover messages may depend on `u_j`, but `(L_j, R_j)` cannot;
- the modeled prover/rewinding tree matches this dependency structure.

This is transcript/protocol soundness, not Orchard circuit semantics. State-restoration soundness and VK/circuit relation correctness remain out of scope for this issue.

TalDerei:
**Human Summary:**

Builds on #20, and progresses [zcash/ironwood#11](https://github.com/zcash/ironwood/issues/11) and closes [zcash/ironwood#23](https://github.com/zcash/ironwood/issues/23) (which were tightly coupled after trying to implement them separately, so I combined the PRs).

The status quo hand-waved Fiat-Shamir in two places: (1) challenges came from an opaque squeeze function (opaque in the sense that it did not model the squeeze at a RO abstraction level), and (2) the bridge from an accepting deployed proof to the transcript tree was assumed. The objective here is to give FS real RO semantics and prove forking / rewinding as lemmas that produce the transcript tree that the extractor ultimately consumes. Simply, this formalizes Ironwood's Fiat–Shamir/IPA forking argument in Lean.

The random oracle is currently a plain deterministic function `O` (transcript → `Fp`) standing in for the black-box hash. Randomness comes from sampling `O` as a uniform random function, so outputs at distinct queries are uniform and independent; oracle reprogramming and the challenge schedule are built on top of that.

Strictly, the goal is to reduce the FS assumptions to the random-oracle idealization itself (where the Blake2b transcript is modeled as a RO). We no longer assume either of:

- forking/rewinding: that the forked transcript tree exists; it is now derived by rewinding in the abstract RO model.
- tree assembly: that the forked transcript tree yields an accepting IPA tree; that assembly is now proved.

The deployed verifier itself is still non-interactive in shape (it takes challenges as free inputs, a deliberate initial design choice), and the soundness layer analyzes it through an interactive lens by relating those challenges to verifier coins in the interactive protocol.

Additionally, this implements round-by-round special soundness lifted from #26.

**Out of scope:** a general oracle-querying adversary and query loss; that is part of #56.

The probability-level capstones here remain `noncomputable` because their probability premise proves that a fork certificate exists. Given a certificate and the commitment representation as explicit inputs, the deployed reduction itself is a plain computed definition; #56 is the operational adversary/extractor follow-up.

----------

**AI Summary:**

This PR replaces the assumed Fiat–Shamir fork with a random-oracle rewinding proof that constructs the accepting transcript tree consumed by the IPA extractor.

## What Is Proven

- **Random-oracle model:** models Fiat–Shamir with an abstract random oracle matching Halo2's challenge schedule.
- **Transcript ordering:** proves challenge ordering, oracle reprogramming, and challenge-vector uniformity.
- **Fork construction:** constructs and assembles the forked transcript tree above the knowledge-error threshold.
- **Computed reduction:** given a fork certificate and explicit commitment representation, computes either an IPA opening or a nontrivial relation.
- **Vesta endpoints:** adds fixed-proof and round-adaptive capstones over the deployed curve.

## Remaining Floor

`hprob` is still the accept-probability premise, so the probability wrappers obtain the fork certificate existentially and remain `noncomputable`. PR #56 adds the general querying-adversary/query-loss layer; an operational recursive extractor must ultimately return that certificate as data. The remaining assumptions are Blake2b-as-random-oracle and the AGM/DL discharge tracked by #28. Constraint-side binding and budget composition are tracked by #18.

## Issue Scope Check

- #23 is implemented.
- #11 is progressed; #56 supplies the querying-adversary/query-loss follow-up.

Builds are green on the current PR head.

### DIFF
--- a/Zcash/Snark/Soundness/Vesta.lean
+++ b/Zcash/Snark/Soundness/Vesta.lean
@@ -155,12 +155,10 @@ theorem commit_adjustedWitness {G : Type*} [AddCommGroup G] [Module Fp G] (urs :
     intro a a'; simp only [commit, Pi.sub_apply, sub_smul, Finset.sum_sub_distrib]
   rw [adjustedWitness, commit_add, csub, commit_single, commit_smul]
 
-/-- **The deployed commitment lies in the `g`-span (discharging `hcommit`).** At Vesta's prime order a nonzero
-generator `urs.g 0` generates the whole group: `c ↦ c • g 0` is injective over `Fp` (nonzero scalars are
-invertible), and `Fp` and `VestaG` are equinumerous (`scalarFieldOrder`), so it is surjective — every group
-element, `deployedCommitment` included, is the `g`-commitment of some witness. So `hcommit` is always
-satisfiable: the witness is the discrete log of `P` base `g 0`, existing but not feasibly constructible —
-the same DLR-side status as the `⊕' NontrivialRelation` disjunct. -/
+/-- **Every Vesta point propositionally lies in the `g`-span.** At Vesta's prime order a nonzero generator
+`urs.g 0` generates the whole group. This theorem is proof-level only: selecting its existential witness
+would compute a discrete logarithm, so computational reductions must receive the representation as input and
+must not use `Classical.choose` on this result. -/
 theorem commit_surjective (urs : URS VestaG) (hg0 : urs.g 0 ≠ 0)
     (P : VestaG) : ∃ aMulti : Fin (2 ^ urs.k) → Fp, commit urs aMulti = P := by
   have hinj : Function.Injective (fun c : Fp => c • urs.g 0) := by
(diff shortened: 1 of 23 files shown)


TalDerei:
sorry manual rebase introduced some merge conflicts; engaging in another rebase to reduce diff against base branch (stacks on `fv/binding-reduction`).

TalDerei:
@daira staging as ready for review now.

Zcash/Snark/Soundness/Vesta.lean
```diff
@@ -137,4 +138,593 @@ theorem orchard_verifier_vesta_constraint_of_forked [DecidableEq VestaG] [Inhabi
...
+    s P hz hg0 hs ?_
+  simpa only [roChallenges_reprogramRounds, roChallenges_spliceIpa_pre] using hprob
+
+open Polynomial in
+open scoped ENNReal in
+open Classical in
+/-- The constraint companion of `orchard_verifier_vesta_forking_opening_adaptive_rewind`: the staged
+round-adaptive capstone with its accept event grounded in reprogrammed-oracle runs on each spliced proof. The
+`hquot`/`hgood` caveat is unchanged — see `orchard_verifier_deployed_constraint_of_forked`'s caveat. -/
+noncomputable def orchard_verifier_vesta_forking_constraint_adaptive_rewind
```
TalDerei: **important**. `orchard_verifier_vesta_forking_constraint_adaptive_rewind` is the top-level capstone. however this is currently marked as `noncomputable` and will need https://github.com/zcash/ironwood/pull/56 and https://github.com/zcash/ironwood/pull/28 to make the reduction computable, so rebasing those PRs accordingly. 

It's still operating on a vacuous existence statement rather than computable def shape (the `Classical.choose` sites that make it `noncomputable` are intentional, a side-effect of how i've decomposed the work).

TalDerei: specifically, the top-level capstone is marked `noncomputable` at this stage because (1) the opening theorem beneath it chooses a DL witness and (2) the generic forking theorem chooses an existential fork certificate (loosely for collected transcript data that a successful forking/rewinding run would hand you). so like previous patterns, they're obtained by existence (`Classical.choose`) rather than produced as data. 

the extractor is modeled in https://github.com/zcash/ironwood/pull/56 which resolves the certificate part, and the AGM is modeled in https://github.com/zcash/ironwood/pull/28 which resolves the commitment opening / DL witness and computes it as data.

TalDerei: filed https://github.com/zcash/ironwood/issues/59 for tracking (this should be resolved collectively again by #28 and #26).

TalDerei:
simplified doc prose by running this through our doc skill file, which should make it more readable.

daira (approved):
Reviewed at branch head `b1c0bb0` (10 commits over main), full local build green (8391 jobs).

## Verdict

No showstoppers. This is a strong PR: the forking layer is genuinely proven rather than assumed, the deterministic extraction is computed data end to end, and the residual floors are documented thoroughly. Recommend merging after the rebase (see the conflict note).

## What it establishes

- **The multi-round forking lemma, proven by counting** (`Tree.extractable_of_kerr_lt`, `Probability.extractable_of_prob`): acceptance above `kerr/N^d` forces a full `(3,...,3)` accepting tree, in one induction over all rounds — no compounding of a per-round cubic loss. The ≥4-good-challenges step correctly buys three *nonzero* ones.
- **Rewinding = reprogramming, sealed to the deployed schedule**: `roChallenges_reprogramRounds` proves redrawing the IPA round vector is exactly reprogramming the oracle at the round prefixes, and `deriveChallenges_ipaRound_eq` pins the reconstructed round transcripts to `deriveChallenges` itself, so any ordering change breaks the proof. Distinctness of prefixes is by length — simple and robust.
- **Challenge-vector uniformity derived, not assumed** (`roChallenges_ipaRound_uniform`): from one idealization (Blake2b squeeze as a uniform random function), for a fixed proof.
- **Extraction as computed data**: `vandermonde3` and `ipa_soundV` become data-returning (`ipa_extractV`), the fork-certificate consumer (`produceDeployed`, `deployed_forking_tree`, `deployed_forking_relation`) computes the tree/opening/relation with no `Classical.choose`, and the Deployed trust boundary is extended with `assert_no_sorry` + pinned axiom sets for all five. The root-consistency argument (`fold_inj` after threading `key`) is correct.
- **The prover-as-oracle bridge's deterministic half is proven**: constant strategies (`deployedVerifierEq_iff_flatAccept`) and prefix-respecting adaptive strategies (`..._adaptive` via `pathData`/`spliceIpa`); the fold-convention mismatch is handled by the `invProver` involution with a measure-equality lemma.
- **D2 from the transcript cross-reference is resolved**: `deriveChallenges` now records the `.challenge` domain marker instead of reabsorbing the squeezed value, and `markerSchedule` re-encodes the Rust-captured schedules so the fixtures still compare (`native_decide`).
- The Vesta rung ladder (`_forking` → `_deployed` → `_adaptive` → `_rewind`, opening and constraint forms) states the capstones at the deployed curve with the structural residuals (`hz`, `hg0`, `hs`, `hξ`) and measure hypotheses (`hprob`, `hbridge`) carried explicitly.

## Candidate follow-up issues

1. **`commit_surjective` is unused and should be deleted.** No consumer in this PR, in main, or in any open stack PR. Its only consumer was the good-challenges PR (#31, closed in favour of folding into #30), and the landed fold derives the good challenge without it. If it ever returns, its own docstring warning governs the use: selecting the witness computes a discrete logarithm, so reductions must receive the representation as input — `Prop`-side use only.
2. **The single-point `reprogram` family is unused across the whole stack** (checked #28/#30/#31/#56/#57): `reprogram`, `reprogram_self`, `reprogram_ne` have no consumers — the live primitive is `reprogramRounds`. Drop them, or factor `reprogramRounds` through iterated `reprogram` so the module docstring's description matches usage.
3. **Restore the computed opening through the probabilistic layer.** `deployed_forking_soundness` computes the opening via `deployed_forking_relation` and then forgets it to `∃ a, ...` because the certificate arrives by `Classical.choose` from `proverAccept_forkValid`'s existential (the inline comment says exactly this). When the adversary experiment lands and certificates arrive as data, the computed form should survive to the capstones.
4. **Account for the `Challenge255 → Fp` reduction bias.** `uniformChallenge` documents the bias as "negligible, unaccounted". It is ≈ `p/2^255` per squeeze; once query loss is accounted, a `q·p/2^255`-style term (or a citation for why it folds into the RO idealization) would close the gap between the model and halo2's actual conversion.
5. **The adversary experiment and query loss** are the documented floor (`Forking/Oracle` module doc, "Remaining adversary model"): `hprob` measures a fixed strategy over challenge vectors, not a querying Fiat–Shamir adversary. If this is not already covered by the existing tracking (#11/#16), it deserves its own issue; the PR's own docs enumerate the missing pieces precisely.

## Nit (no action needed unless convenient)

- The book glossary/proof map on main reference stack-wide modules (`AGM.*`, `Multiopen.*`) that land later in the stack — deliberate docs-first publishing, fine once the stack merges.

## Rebase note (for the merge, not the review)

The `Soundness/Main.lean` conflict with main predates #50: main's `1896a62`/`b3c66ee` dropped the `∀ a` quantification from the deployed constraint/opening hypotheses — which is exactly the restatement this PR's own module docstring calls for ("must be restated over the extracted witness"). The resolution should keep main's restated forms.

TalDerei:
@daira filed https://github.com/zcash/ironwood/issues/70 to track follow-up work.

PR #27, opened 2026-07-01
feat(fv): multi-action fixture support

### LINKED ISSUE #17: Capture a multi-action (numProofs ≥ 2) fingerprint fixture
The fingerprint match (`Zcash.Snark.Fixture`, #9) validates the Lean `assembleFinalMsm` against the deployed Rust verifier on a **single captured proof with `numProofs = 1`** (`shape.numProofs := 1`). A real Orchard *bundle* has multiple actions (`numProofs = N`); the proof string's per-sub-proof vectors are indexed by `Fin numProofs`, and the assembly iterates them with per-sub-proof `flatMap`/folds (`subProofExpressions` / `allExpressions`, `assembleQueries`).

With only a `numProofs = 1` capture, the **multi-action path is never exercised by the fingerprint**: a transcription error specific to the per-sub-proof iteration (ordering across sub-proofs, the `flat_map` schedule, index wiring at `numProofs > 1`) would not be caught, because the empirical Rust↔Lean check runs only at length 1.

Note this is purely an **assembly-fidelity coverage** gap — the soundness theorems are generic over `numProofs`; it's the fingerprint validation of the transcription that is under-exercised.

Suggested fix: dump and check a second fixture at `numProofs ≥ 2` (a real multi-action Orchard bundle), so the per-sub-proof assembly is fingerprinted against Rust.

🤖 Claude Opus 4.8 (1M context)

TalDerei:
**Human Summary:**

Closes https://github.com/zcash/ironwood/issues/17 and https://github.com/zcash/ironwood/issues/32. 

This parameterizes the Lean verifier obgligations over arbitrary `numProofs`, but i couldn’t realistiically generate a Rust fixture for Lean to consume because it’s be hundreds of megabytes and CI would choke. Instead, it (a) adds a fixture that instanitate those generic obligations at the Orchard max consensus N = 2^16 - 1 (this is basically a compile time kernel check that sanity checks none of the defintions were explicitly defined or specialized under the assumption of a specific N), and (b) a real multi-action fingerprint fixture for N = 2.

Additionally, it adds some other checks.
- Since the challenges `ch` are inputs, their FS schedule `hch` was assumed so we weren't checking `deriveChallenges` actually reproduces the captured `ch`; now the FS fixture checks the Lean `deriveChallenges` absorb/squeeze schedule against the transcript (a fixture oracle built from the Rust-captured schedule). So the fixture now exercises the schedule as well.
- Additionally, it adds concrete negative fixtures for malformed/tampered fingerprints that excerise a rejection path.
- Finally, there are faithfulness fixtures that check list/layout consistency and range safety for the generated VK. Though it still doesn't derive the Orchard VK from Rust keygen or model circuit semantics inside Lean. These checks tighten some of those boundaries (eg. FS schedule and shape/VK sanity), and they're distinct from the plain MSM fingerprint.

--------

**AI Summary:**

This PR turns multi-action fixture coverage into a small test suite. Mental
model: the Lean verifier is generic in `numProofs`, but selected fixtures pin
real Rust-captured cases. This PR adds the real `N = 2` capture, a captured
Fiat-Shamir schedule check, negative/tamper cases, shape faithfulness checks,
and a max-consensus parametric instantiation.

## What Is Proven

- **Parametric verifier shape**: `Verifier.Parametric` exposes the per-proof
  folds for expressions, Fiat-Shamir absorbs, `assembleQueries`, and
  `deriveChallenges` for arbitrary `shape.numProofs`.
- **Two-action fingerprint**: `Fixtures.TwoAction.Fixture` proves that Lean's
  `assemble vk ps ch` matches the captured Rust MSM for a real two-action
  Orchard bundle.
- **Fiat-Shamir schedule fixture**: `Fixtures.TwoAction.FiatShamir` uses the
  Rust-captured typed transcript prefixes, including `capturedInit`, to check
  that Lean derives the captured challenges in the expected order.
- **Faithfulness and negative checks**: the two-action suite checks list
  lengths, query-layout ranges, VK shape sanity, malformed/tampered rejection
  paths, and fingerprint sensitivity to swapped or edited proof data.
- **Max-shape instantiation**: `Fixtures.MaxShape` specializes the parametric
  verifier obligations at Orchard's consensus maximum action count.

## Remaining Floor

The empirical coverage is still by selected captures, not every possible
transaction size. `Fixtures.MaxShape` is a compile-time parametric instantiation,
not a real 65,535-action Rust capture. The Fiat-Shamir checks are typed
fixture-oracle checks; byte decoding, Blake2b/RO semantics, generated VK/proof
fields, Rust capture faithfulness, and `native_decide` compiler trust remain
outside this PR.

## Issue Scope Check

Issue #17 asks for multi-action fingerprint coverage, and issue #32 asks for
negative/adversarial fixture coverage. This PR covers both: it adds the real
two-action fingerprint and Fiat-Shamir fixtures, wires in faithfulness and
negative checks, and pins the generic verifier shape at the consensus maximum.


Zcash/Snark/Fingerprint/FixtureFiatShamir2.lean
```diff
@@ -8,21 +8,25 @@ fingerprint (`nonInteractiveFingerprint`, i.e. `assemble` at `deriveChallenges`)
...
-disagree with the captured challenges here. See zcash/ironwood issue #17.
+With `numProofs = 2` this theorem puts the schedule's *per-sub-proof absorb interleavings* on the
+conditional path — all proofs' advice commitments before `θ`, per-proof-per-lookup permuted pairs before
+`β`/`γ`, per-proof evaluation blocks before `x₁`, and so on — which the single-action fixture exercises
+only at length 1.
+
+TODO(zcash/ironwood#24): once the Fiat-Shamir transcript-ordering work lands, revisit this module and
+either connect the two-action capture to a concrete `fs`/`init` captured-challenge check, or point this
+comment at the landed schedule-ordering theorem. Until then, this theorem remains conditional on `hch`;
+it does not by itself detect a misordered `deriveChallenges`. See zcash/ironwood issue #17.
```
TalDerei: depends on https://github.com/zcash/ironwood/pull/24.

TalDerei:
Uses upstream https://github.com/zcash/halo2/pull/914 and https://github.com/zcash/orchard/pull/531 for generating the rust fixtures.

Zcash/Snark/Fixtures/MaxShape.lean
```diff
@@ -0,0 +1,149 @@
...
+
+open Zcash.Snark
+
+abbrev G := ℕ
+
+/-- The captured Orchard verifier shape (`Fixture`/`Fixture2`) with `numProofs` at the consensus
+maximum. -/
+def shape : Shape := {
+  k := 11,
+  numProofs := orchardConsensusMaxProofs,
```
daira: We could prove the `_at_consensus_max` things instead for any number of proofs ≤ consensus max.

Also, reference the protocol spec for the consensus maximum (it's in section 7.1.2).

daira (approved):
ACK

Zcash/Snark/Verifier/Assemble.lean
```diff
@@ -295,7 +295,9 @@ def assembleFinalMsm? {shape : Shape} {F G : Type*} [Field F] (ps : ProofString
   | none => none
 
 /-- The multiopen inverse factors are defined only when the IPA challenge `x₃` is not one of the opened
-points in any derived point set. -/
+points in any derived point set. In the deployed verifier this case is a panic, not an error return
+(`(x₃ - point).invert().unwrap()` in `multiopen/verifier.rs`); the Lean rejection abstracts that crash —
+both are non-accepting, which is what soundness consumes. -/
```
daira: Separately, prove that panic cases happen for a negligible proportion of challenges.

TalDerei: addressed with two new theorems

Zcash/Snark/Fixtures/MaxShape.lean
TalDerei: extended to all proofs up to consensus max and reference consensus rules in protocol ZIP.

TalDerei:
upstream branches opened in https://github.com/zcash/halo2/pull/914 and https://github.com/zcash/orchard/pull/531.

Zcash/Snark/Fixtures/MaxShape.lean
```diff
@@ -135,11 +145,11 @@ theorem deriveChallenges_parametric_at_consensus_max (fs : FiatShamir Fp G)
         ipaRound := fun j => ipaRes.2.getD j.val 0 } :=
   deriveChallenges_parametric_numProofs fs init ps
 
-/-- `subProofOpeningQueries_commId_disjoint` at the consensus-max shape: even at `N = 65535`, no two
-distinct sub-proofs' opening queries share a commitment slot, so the multiopen grouping cannot merge
-across sub-proofs. -/
-theorem commId_disjoint_at_consensus_max (vk : VerifyingKey shape Fp G)
-    (ps : ProofString shape Fp G) (x xInv xNext xLast : Fp) {p p' : Fin shape.numProofs}
+/-- `subProofOpeningQueries_commId_disjoint` at the captured Orchard shape: for every action count `n`,
```
TalDerei: consensus-valid action count

Issue #15, opened 2026-06-28
Reduce commitment binding to the discrete-log assumption in the AGM


TalDerei:
The binding step (#13) expresses commitment binding as a discrete-log-relation (DLR) hardness assumption, but that relation is itself still assumed hard. This issue tracks discharging it under the algebraic group model (AGM) heuristic: for an adversary restricted to algebraic strategies —every group element it produces comes with a known representation over the generators— a binding collision yields an explicit nontrivial relation, reducing binding to the plain discrete-log assumption.

This replaces the assumed DLR-hardness with plain DL hardness against the algebraic adversary class; the AGM itself is a heuristic restriction of that class, not an additional falsifiable assumption.

**Scope (broad):**
- Model the prover as algebraic (group elements carry their representations over the URS generators).
- Show that a commitment collision yields an explicit nontrivial discrete-log relation in the AGM.
- Reduce the DLR-hardness assumption from #13 to plain discrete log, closing out the binding assumption.

PR #28, opened 2026-07-02
feat(fv): AGM binding reduction

TalDerei:
**Human Summary:**

Builds on #24 and closes https://github.com/zcash/ironwood/issues/15.

References https://github.com/zcash/ironwood/pull/4#discussion_r3483411048 for the motivation: discharge binding capstones through the Algebraic Group Model (AGM) by modeling an algebraic adversary. This PR also defines the data that the Fiat–Shamir adversary layer must populate/supply to the AGM reduction (this PR largely defines those interfaces like `AlgebraicProver`, `AlgebraicDForkCert`, etc).

----------

**AI Summary:**

This PR turns computed algebraic relations into discrete-log breaks, for an algebraic adversary. It hides a DL challenge in a random basis slot, then extracts the discrete log whenever the adversary's returned relation hits that slot.

The algebraic adversary is the AGM view of the same prover, not a separate attacker. This PR defines and consumes its representation-carrying interface; PR #56 connects the real bounded-query Fiat–Shamir adversary to that interface and produces the deployed instance family.

## What Is Proven

- **Computed relation path:** `AGM/Adapter.lean` consumes explicit `NontrivialRelation` data; it no longer selects a witness from an existential proposition.
- **Probability accounting:** `AGM/Probability.lean` proves the random-slot hit bound and relation-finding-to-DL reduction; `ProbabilityVesta.lean` specializes it to Vesta.
- **Algebraic-prover wiring:** `AGM/Peel.lean` and `Capstone.lean` use representation-carrying prover outputs to compute either an IPA opening or a relation.
- **Binding-signature reuse:** `AGM/BindingSignature.lean` converts the already-computed Orchard/Sapling binding relation into the shared AGM witness and a DL representation.
- **Trust boundary:** `AGM/TrustBoundary.lean` checks the data-producing definitions are `sorry`-free and remain plain computable definitions.

## Remaining Floor

This PR proves that `AlgebraicProver + Extractable` gives an algebraic fork certificate, but its executable reduction starts from a supplied `DeployedAlgebraicForkingInstance`. PR #56 must produce those instances from the real Fiat–Shamir adversary and preserve its acceptance probability. Plain-DL hardness, the AGM, and generator hash-to-curve-as-random-oracle remain model assumptions.

## Issue Scope Check

This closes issue #15 at the AGM boundary: PR #28 defines the algebraic certificate/instance interface and reduces its relation event to plain DL. PR #56 is responsible for producing those instances from the real Fiat–Shamir adversary.


Zcash/Snark/Soundness/AGM/Adapter.lean
```diff
@@ -0,0 +1,430 @@
...
+This module records the AGM-facing layer that consumes that relation branch. It follows the useful part of
+ArkLib's AGM scaffold: group elements are paired with representations over a public basis. The load-bearing
+theorem is the deterministic DLR-to-DL adapter: if all non-challenge bases have known discrete logs with
+respect to a base `B`, then any found relation with a nonzero coefficient on the challenge target recovers
+the target's discrete log.
+
+The probabilistic computational wrapper — placing the DL challenge into a random basis slot and accounting
+for the chance that the relation's coefficient there is nonzero — is not an oracle-machine model here. The
+Lean content below is the algebraic core that such a wrapper calls.
+-/
```
TalDerei, line 38: we model our construction in the spirit of ArkLib (https://github.com/Verified-zkEVM/ArkLib/blob/main/ArkLib/AGM/Basic.lean#L13-L14). it seems to be just a skeleton afaict (has an unproven sorry stub + todos in the adversary definition), and it doesn't model the original AGM paper (https://eprint.iacr.org/2017/620). regardless, we can source inspiration from their scaffolding.

Zcash/Security/BindingSignature/Balance.lean
```diff
@@ -28,17 +29,15 @@ binding hypothesis. Instead we phrase binding as a reduction:
...
-  efficient adversaries, supplied by the algebraic group model or a DLR hardness assumption at the
-  computational layer. This is the shape of the spec's argument: *if you can unbalance, you can
-  solve DL*. The relation and discrete-log problems are tightly equivalent (Jaeger and Tessaro,
-  https://eprint.iacr.org/2020/1213 Lemma 3 in general, but this case is very simple), so this is
-  no stronger than DL.
+  "The bundle balances" then reduces to the discrete-log relation problem (DLR). The AGM handoff
+  below consumes the relation branch as a plain-DL witness under the known-log embedding. This is the
+  shape of the spec's argument: *if you can unbalance, you can solve DL*. The relation and
+  discrete-log problems are tightly equivalent (Jaeger and Tessaro, https://eprint.iacr.org/2020/1213
+  Lemma 3 in general, but this case is very simple), so this is no stronger than DL.
```
TalDerei: binding-signature balance arguments (sapling and orchard) now import AGM module and show DLR-hardness reductions through AGM.

daira (approved):
Reviewed at branch head `cefaf9e` (13 commits over #24), with the stack top (#56) built green locally to validate composition.

## Verdict

No showstoppers. The fixed-slot AGM reduction is implemented the right way — the extraction is exact computed data, the probability layer's success event never references the reduction's knowledge, and the distributional obligations are isolated into explicitly named assumptions at the Vesta boundary. Recommend merging in stack order after #24.

## What it establishes

- **The FKL fixed-slot adapter as computed data** (`AGM/Adapter`, citing eprint 2017/620): representations over an arbitrary public basis, `AlgebraicRelationWitness`, and `discreteLogOfBasis_of_relation` — an *exact* recovery of the challenge slot's log from any relation hitting it, with `noncomputable` confined to the counting side (`nonzeroCoeffSlots`).
- **The slot-loss probability accounting is sound** (`AGM/Probability`): `relSet ≤ succSet` counting gives `Pr[relation]/|ι| ≤ Pr[hit]`, hit ⟺ win is justified by the Adapter's exact recovery, and the embedded game (where the reduction samples *all* logs — explicitly "not yet textbook DL hardness") transfers to the textbook game by the `winSet` coin bijection, proven as a measure equality rather than asserted. The `|ι| = 2^k + 2` loss is honestly flagged, with the tighter (`1/|F|`-loss) reduction named as follow-up.
- **"The reduction knows the logs" is handled correctly**: the success event depends only on the adversary's output; the slot is jointly independent of the presented basis (proven via the coin bijection); and the legitimacy of self-generating the basis is exactly the `OrchardUniformURSIdentification` obligation — proven inside the generator-RO model (`orchard_uniformURSIdentification_of_generatorRO`, with `hB : B ≠ 0` explicit and the injective domain-separated queries modeling `gᵢ = H(0‖i)`, `W = H(1)`, `U = H(2)`), with hash-to-curve-as-oracle priced as the remaining assumption.
- **The peel with representations** (`AGM/Peel`): `separateOrRelationWitness` / `deployedLeafPeelWitness` / `deployedToAcceptVWitness` compute the clean-checks-or-relation dichotomy at every step, no `Classical.choose` anywhere on the data path.
- **The capstone plumbing** (`AGM/Capstone`): `deployedAlgebraicForkingRelation` / `...FixedSlot` compose the #24 kernel with the AGM types; the instance/finder/event scaffolding (`DeployedAlgebraicForkingInstance`, `run`, `deployedAlgebraicRelationFinder`) is exactly what the probability experiment consumes, and `AlgebraicDForkCert.toDForkCert` erases to #24's checked certificate so `DeployedForkValid` is shared, not duplicated.
- **The binding-signature handoff** (`AGM/BindingSignature` + the `Balance.lean` doc update): our computed `NontrivialRelation V R` converts to the generic witness, and —nicely— to a DL solution *deterministically, with no slot loss*: with two slots and `R ≠ 0` the `α = 0` branch is impossible. The `Balance.lean` docstring now routes binding through this handoff while noting DLR hardness remains available directly at the lower layer.
- **Trust boundary**: `assert_no_sorry` plus pinned axiom sets over the entire computed-data chain, through to the binding-signature endpoints.
- **The API is validated by its consumer**: #56 uses `AlgebraicRelationWitness`, `toDForkCert`, the generator-RO setup/basis, `OrchardUniformURSIdentification`, and `deployedAlgebraicForkingRelation` as-is — no interface rework needed downstream.

## Candidate follow-up issues

1. **The tighter AGM reduction** (lose `1/|F|` instead of `|ι| = 2^k + 2`): named as follow-up in `AGM/Probability`'s module doc; worth an issue so it doesn't live only in a docstring.
2. **Mirror duplication in the Adapter**: `discreteLogOfU_of_augmentedRelation` and `discreteLogOfW_of_augmentedRelation` are near-identical, and `NontrivialRelation.toDiscreteLog` already demonstrates the generic form by instantiating the `U`-variant at `n = 0`. One slot-generic lemma could serve all three.
3. **The FS→instances gap**: the boundary "the Fiat–Shamir layer must produce `DeployedAlgebraicForkingInstance` values and relate its acceptance event to the relation event" is this PR's honest floor (the proof map's *gap* edge). #56 addresses it; confirm the existing tracking (issue #15) still describes the post-#56 residue accurately.

## Nits (no action needed)

- The binding-side accounting (the `(V, R)` distribution identification and DL bound) arrives only in #56; until then `Balance.lean`'s "assumes no more than DL hardness" is a small forward reference. Accurate once the stack lands.
- The uniform-basis model gives each slot the identity with probability `1/|Fp|`, which the real hash-to-curve derivation may exclude. This is folded into the RO idealization, and below the bound's noise floor.

## Rebase note

Inherits #24's `Soundness/Main.lean` conflict transitively; nothing #28-specific. The `fv/agm`-side `Vesta.lean` changes (opening the rungs at declared commitment components) sit in the same region main's `∀ a`-restatement touched — same resolution guidance as #24.

TalDerei:
@daira folded up the cross-PR follow-ups in https://github.com/zcash/ironwood/issues/70.

PR #30, opened 2026-07-03
feat(fv): bind IPA witness to verifier columns

### LINKED ISSUE #18: Bind the extracted IPA witness to the circuit columns via the multiopen decode

TalDerei:
**Human Summary:**

Closes https://github.com/zcash/ironwood/issues/18 and progresses foundations for eventual capstone soundness wiring in https://github.com/zcash/ironwood/issues/67.

Before, the IPA witness was free floating and not formally tied to the columns the verifier opens. Now, the decoded capstones recover those columns from the batched multiopen data, run the gate checks on them, and show the extracted IPA witness is the current batch-combination of them. So we're checking that the extracted witness satisfies the gate checks associated with the unbatched decoded columns.

This only establishes the foundations for #67 and doesn't yet wire into existing top-level capstone `orchard_verifier_sound_vesta_conditional`. That wiring remains in #67 scope, after the lookup/permutation integration in #36.

-----------

**AI Summary:**

This PR binds the extracted IPA witness back to the columns opened by the
verifier. Mental model: the verifier checks one batched multiopen, but the gate
check needs the individual columns, so the proof now decodes that batch and makes
the capstones read the decoded columns instead of an arbitrary decode function.

## What Is Proven

- **Decode layer**: rewound batched openings at distinct challenges recover the
  underlying column witnesses, and the current extracted witness is their batch
  combination.
- **Deployed batching**: the verifier's `x₄` multiopen collapse is proved to be
  the flat power batch consumed by the decode, rather than assumed as a model
  boundary.
- **Member binding**: the `x₁` layer unbatches each point-set aggregate down to
  the member commitments, and `deployed_witness_member_binding` connects the
  extracted witness to those queried commitments.
- **Capstone foundations**: the deployed constraint rungs take decoded columns
  rather than a free `decodeAdvice` / `decodeInstance` input. This is the
  foundation the top-level capstone integration builds on; the integration
  itself is follow-up work.
- **Budgeted floors**: the derived capstone needs a separate accept-probability
  floor for every squeeze at every rewound run — a shape no single
  knowledge-error number can consume. The budgeted form (`VestaBudget`)
  re-proves the extraction from one joint accept floor, so the knowledge error
  is a single budget `Σtᵢ`.
- **Ordering and fixtures**: the `x₁` / `x₄` transcript prefixes are sealed, and
  fixtures cover the decoded-column and deployed-batching paths.

## Layout

`Soundness/Composition/` joins the two halves of soundness — forking's accepting ⟹ clean opening and the decode's opening ⟹ those columns satisfy the gates — into one quantitative accepting ⟹ a witness can be extracted, except with bounded probability.

`Soundness/Composition/` holds the step from *the verifier accepted* to *a witness can be
extracted*. `Bridge` identifies the extracted IPA opening with the deployed commitment and ties it
to the batch witness; `Decomposition` splits the knowledge-error bound into the clean-opening bound
plus a residual; `Residual` and `Assembly` bound that residual, by forking and by the adaptive
coupling respectively; `Prefixes` pins where the multiopen challenges sit in the transcript; and
`Completeness` supplies the containment those bounds consume.

## Remaining Floor

The gate-check content itself, including the `x` to `x₃` transport and claimed
evaluation faithfulness, remains with the fingerprint work (#11/#13). The
DLR/AGM relation branch remains the surrounding binding layer (#28), and the
usual random-oracle / rewinding budget is still explicit.

## Issue Scope Check

Issue #18 asks to recover the real opened columns from the batched IPA witness
and extend the needed multiopen transcript-ordering treatment. This PR does
that: the decoded witness is produced from rewound multiopens, the deployed
`x₄` and `x₁` batching shapes are proved, and the deployed capstones consume the
decoded columns.

## Follow-up work

- **#67** — composing the forking extraction with these decoded capstones: the adaptive coupling
  (conditional-continuation weights + the `G1 ≤ τ` gate, into the `AdaptiveWeight` interface this
  PR provides), the concrete `hcont`, the terminal `hgood` wiring, and the cleanup of the
  fixed-interface decode consumers. The handover, precise obligations, and sequencing are recorded
  in comments there.
- **#36** — the permutation + lookup fold-in: discharging `hfp` and restating `hfold` / `hgood`
  over the extended expression list. No re-capture needed.

Neither blocks this PR.


Zcash/Snark/Soundness/Multiopen/Decode.lean
```diff
@@ -0,0 +1,459 @@
...
+   extracted witness as the explicit two-level (`x₄`-then-`x₁`) power combination of member-column
+   witnesses. The honest run sits inside every family by structure eta (`honestX1Run`); the per-run
+   aggregate witnesses are each run's own `x₄`-level decode (the same stacked floors, per run), the
+   hypothesis shape `hwC`/`hwu` carries.
+
+**Delegated to the equivalence fingerprint — unchanged.** Per the principle raised in
+zcash/ironwood#21 (the `PermutationConstruction.lean:243` review thread, comment `r3493240277`:
+supply the Rust-produced object and check equivalence rather than reproduce the deployed computation
+in Lean; `Fingerprint.Match` is its in-tree home), the *faithfulness* obligations stay with the
+fingerprint (`Fingerprint.Match`, `MsmMatch`): the
```
TalDerei: another targeting case here where i decided to model after https://github.com/zcash/ironwood/pull/21#discussion_r3489262683 and relegate to fingerprinting equivalence rather than making the lean code more complex.

book/src/formal-verification/glossary.md
daira: This is out-of-date; there is no `BindingSignature.TrustBoundary`. We're in an intermediate state before #71 is done, so we should just say that here.
```suggestion
<div class="g"><div class="g-head"><span class="term">checked trust boundary</span><span class="anchor">Zcash.TrustBoundary</span> (and temporarily other TrustBoundary modules)</div><div class="def">Build-time pins on what a theorem may rest on: <code>assert_axiom</code> asserts a bound on the axioms used by a definition, so that a stray <code>sorry</code> or a new axiom fails the build instead of silently widening the trusted base. <code>assert_computable</code> additionally asserts that the definition is a plain `def`, ensuring constructivity of security reductions. Until <a href="https://github.com/zcash/ironwood/issues/71">#71</a> is done, some of the TrustBoundary modules also use a <code>#guard_msgs</code>-pinned <code>#print axioms</code> check instead of these macros. See <a href="../formal-verification.html#trust-discipline">Trust discipline</a>.</div></div>
```

TalDerei: fixing as part of the intended rebase.

daira (commented):
Reviewed at branch head `c5fd9cef` (12 commits over main), full local build green. The Rust capture pipeline (zcash/orchard#531) was read alongside.

## Verdict

No showstoppers. The gap due to free `decodeAdvice`/`decodeInstance` that let `circuitSat` float loose of the extracted witness is genuinely closed:

* The batched multiopen is decoded by proven Vandermonde inversion with distinctness obligation coming from the accept measure.
* The deployed $x_4$ collapse is proven to be the flat power batch rather than modelled as one.
* The terminal capstone derives `hquot` from node-binding-shaped hypotheses instead of assuming it.

The remaining floors (the gate-structure fingerprint, the adversary experiment) are documented.

However,

1. **`commit_surjective` is back, still with zero consumers.** The rebase resurrected the declaration deleted after the #24 review (its only historical consumer was the folded good-challenges work, which does not use it). Delete it again.
2. **The tracked decode-gap comment is now stale**: `KnowledgeSoundness.lean:32` still says `batch_open_soundV` is "currently unused", but this PR's `Multiopen/Decode` consumes it and declares the residual closed. Repoint the comment at the decode layer.
3. **The node binding is not yet wired into the terminal capstone; until it is, several docstrings implicitly overclaim.** The terminal capstone `orchard_verifier_vesta_member_constraint_deployed_terminal` takes `hfixed`/`hadvice`/`hinstance` as hypotheses whose intended discharge it names in prose. The composition —from the accept measures through the $x_3$ and $x_2$ sample identities to those hypotheses— is the deferred #67 work. In particular, `deployed_setAggregate_node_binding` is proven but referenced *only* in docstrings. The `orchard_verifier_vesta_member_constraint_deployed_terminal` docstring (and the `Decode.lean` module docstring) imply that it makes use of `deployed_setAggregate_node_binding` but it does not. The same applies to `hgood`: the production machinery exists (`GoodChallenge`), and the `_xgood` rung that would consume it is present but unconsumed.

## What it establishes

- **The deployed batching is a theorem, not a model boundary**: `deployedCommitment_x4_batch` / `multiopenValue_x4_batch` prove the pinned commitment and value are flat power batches $\sum_j \xi^j \bullet C_j$ and $\sum_j \xi^j e_j$ in the $x_4$ squeeze, retiring the old flat-model seam.
- **The decode is computed and correctly gated**: `decodedColumnFamily_of_batch_openings` inverts the Vandermonde system under `challengesDistinct`, and that distinctness is *produced* by the counting floor `exists_injective_accepting_of_measure` from the accept measure — a genuine obligation discharged, not assumed. The $x_1$ level unbatches aggregates to member commitments (`x1_batch_open_soundV`, `member_binding_of_x1_samples`), and `deployed_witness_member_binding` / `deployed_witness_two_level` state the endpoint: the extracted witness is the explicit two-level power combination of member-column witnesses.
- **The value-check core is proven with guarded algebra**: the r-polynomial layer (`RPoly`) and cleared-identity un-batching (`node_binding_of_samples`) use inverses only under off-node hypotheses (`mul_inv_cancel₀` under `≠ 0`), and the coefficient isolation is the power-sum Vandermonde argument. `deployed_setAggregate_node_binding` instantiates it at the fingerprinted grouping with no `hconsistent` assumed.
- **`hgood` is producible, constructively**: `GoodChallenge.lean` derives the good gate-check challenge from the accept measure by pigeonhole (`exists_accepting_good_challenge`), with the per-site $\mathbb{F}_p$-squeeze exclusions composed into one budget (`uniformChallenge_szBadSet_union`). No choice, no new axioms.
- **The capstones consume what the layers produce**: `hrel` and the explicit witness are discharged (extracted from the clean accept via `ipaRelation_of_acceptV`); the intermediate `_x4` rungs pin `hquot`/`hgood` at the canonical decode at the extracted witness (avoiding the jointly-unsatisfiable quantified form the module doc itself warns about); and `orchard_verifier_vesta_member_constraint_deployed_terminal` derives `hquot` from `hfixed`/`hadvice`/`hinstance` + `hfold` via `quotientCheck_of_claimed`.
- **The halo2-faithfulness audit items are structural**: rotation-correct feeds (`rotatedFeed`, degree preserved), layout-forced column selection (`hadviceLayout`/`hinstanceLayout` against the VK query layout), and committed quotient (`hquotCommitted`). The `getD` defaults in these hypotheses are fail-safe: each default differs from the demanded identity, so a short layout list makes the hypothesis unsatisfiable rather than silently satisfied.
- **The ordering seals are model-side**: the `preX{1,2,3,4}` prefixes seal by `rfl` against `deriveChallenges` (the marker-based transcript), not against captured fixture schedules — which is necessary, since the Rust dumper's schedule is explicitly non-injective (it appends challenges as plain scalars; zcash/orchard#531's module doc says it must not carry transcript-binding weight).
- **Hygiene**: no `sorry`, no `native_decide` anywhere in the diff; `Classical.choose` only inside `noncomputable def`s extracting accepting-run families from counting floors, with the load carried by injectivity fields; `DecodeFixture` uses kernel `decide` only. The AGM trust-boundary file is net-unchanged (an intra-PR regression was restored by `fix(ci)`).

## Candidate follow-up issues

1. **Wire the node binding into the terminal capstone.** This is #67. Only the documentation of the current status needs to be fixed now.
2. **Trust-boundary coverage for the new endpoints.** The four new Vesta capstones and the `GoodChallenge`/`Multiopen` surface have no assert/guard coverage. Natural to fold into #79's consolidation rather than grow the old idiom here.
3. **Unused declarations** (repo-wide reference count 1, verify before deleting): `claimedCombined_of_x2Prob` (self-documented as structural-only, deferred to #67), `gateGood_of_xProb` (docstring references only), `member_constraint_of_relation_and_batch_xgood`, `member_constraint_of_relation_and_batch_hbound`, and `opened_constraint_of_opening_or_relation`. Wire or prune alongside follow-up 1, which is what most of them are shaped for.

## Nits

- `Claimed.lean:27`'s module doc names `gateGood_of_x3Bound_of_xProb`; the theorem is `gateGood_of_xProb`.
- `DecodeFixture` witnesses hypothesis-shape satisfiability on a $k = 0$ toy whose commitment kernel is trivial — by design, but it therefore does not witness non-vacuity of the general endpoints. The vacuity concern is instead addressed by the capstones pinning `hquot`/`hgood` at the canonical decode.
- The source docstring's `zcash/ironwood#27` (`Claimed.lean:79`) appears to be a typo for #67 — it points at the merged multi-action fixture PR.

[Edited: #27 → #67 in three places above, per the typo mentioned in Nits.]

TalDerei:
@daira already tracking some of the capstone integrations in https://github.com/zcash/ironwood/issues/67. I'd like to fold that in here so these statements aren't free floating.

daira:
I had missed #67 because `Claimed.lean:79` incorrectly references #27 instead.

daira:
One gap in #67 as scoped: its endpoint needs the value-check supply — discharging `hfixed`/`hadvice`/`hinstance` from the accept measures through the $x_3$ and $x_2$ sample identities. `Claimed.lean:79` defers that composition to `zcash/ironwood#27`, which I think is a typo for #67 — so #67's scope should name the value-check supply explicitly alongside the forking composition.

Five small things are still, I think, cheaper to fix in this PR than to carry:

1. Delete the resurrected `commit_surjective` (a rebase artifact; no consumers).
2. The stale `KnowledgeSoundness.lean:32` comment ("currently unused `batch_open_soundV`").
3. If not otherwise addressed, the docstring wording in the terminal capstone and the `Decode.lean` module doc. "the deployed instantiation of `deployed_setAggregate_node_binding`" reads as already-wired; "to be discharged by" matches the formal state (same for the `hgood` phrasing).
4. The `Claimed.lean:27` module-doc lemma name (`gateGood_of_x3Bound_of_xProb` -> `gateGood_of_xProb`).
5. The `zcash/ironwood#27` typo in `Claimed.lean:79` (should be #67).

None of these touch the composition itself.

TalDerei:
> One gap in #67 as scoped: its endpoint needs the value-check supply

appended to #67 scope.

Zcash/Snark/Soundness/Composition/Bridge.lean
```diff
@@ -0,0 +1,690 @@
...
+composition chain carries no shift hypothesis.
+
+## The extracted `U`/`W` coordinates (`pU`, `pW`)
+
+`pU`/`pW` are outputs, never hypotheses: the AGM representation (`AlgebraicWfProof.multiopen_repr`)
+exposes the adversary's aggregate commitment as `commit(aMulti) + pU•u + pW•w`, so the extracted
+opening is the generalized Pedersen triple `(aMulti, pU, pW)` in the augmented basis `(g, u, w)`.
+The relation is stated at the de-blinded point `deployedCommitment − pU•u − pW•w` because
+`IpaRelation` is the `g`-span opening. `pW` is the Pedersen blinder — nonzero for honest proofs, so
+it can never be forced to `0`. Weight on `u` shifts the opened value by `z⁻¹·vU`; against the
```
daira: Why does it being nonzero for honest proofs imply that it can never be forced to 0?

daira (commented):
Reviewed at branch head `25377c3d` (23 branch commits over main, +14.3k/−0.1k over 38 files), CI fully green. Round 1 reviewed the decode layer at `c5fd9cef`; this round covers the response to that review and the new composition stack (`Soundness/Composition/`, `VestaBudget`, `Multiopen/{FloorBudget,BudgetedExtraction,NodeBinding}`, `Forking/AdaptiveCoupling`).

## Verdict

All five round-1 asks are resolved: `commit_surjective` is deleted, the stale `KnowledgeSoundness.lean` comment is gone, the `Claimed.lean` module-doc lemma name is fixed, the `#27` typo is fixed, and the docstring-overclaim concern is superseded by the real thing — the composition layer round 1 said was deferred to #67 now exists in graded form, and the scope prose (PR description, `Bridge.lean:472`, the `ExtractableFromAcceptance` correction in the final commit) states the remaining boundary accurately. The knowledge-error accounting is built from sound primitives:

- the per-squeeze floor-failure charges are valid by construction (`μ{accept ∧ floor-fails} ≤ t` holds by cases on the floor);
- the coin–challenge coupling uses the standard reprogramming/union-bound skeleton, with a tidy `Function.update_idem` blindness gadget generalizing it to table-dependent accept events;
- the budgeted member endpoint (`deployed_member_budget`) has the classical proof-of-knowledge dichotomy shape: acceptance mass above the budget `t₁ + t₂ + t₃ + t₄` extracts, on pain of a computed `(g, U, W)` relation.

## What the graded capstone family establishes

- `…_deployed_terminal` derives `hquot` (via `quotientCheck_of_claimed`);
- `…_member_constraint_derived` additionally derives `hfixed`/`hadvice`/`hinstance` from per-squeeze accept floors (`deployed_member_node_binding_at_point`);
- the budgeted pair (`…_member_constraint_budgeted`, `orchard_verifier_sound_vesta_budgeted`) derives all four from a single joint accept floor, so the knowledge error is one budget with one count-over-`|F|` term per Fiat–Shamir squeeze;
- `orchard_verifier_sound_vesta_computed` routes the deployed gate data through the decoded columns, with the honest qualification that `pbatch`/`mdec`/`hquot`/`hgood` are supplied by hand — producing them from a bare accepting run (the "coin–challenge coupling") is named as the open surface, and the conditional endpoints still carry `hExtract`. `hgood`/`hfold` remain the supplied gate surface throughout, consistent with the stated #36 → #67 sequencing.

## Questions

1. **`havoid` vs the `t₃` charge.** The budgeted endpoints take `havoid` — every accepting canonical x₃ run avoids every opened set point — as a universal hypothesis, while the `t₃` threshold `(max(2^k, |allPts|) + |allPts|)/|F|` carries a `+ |allPts|` summand that looks like a probabilistic charge for exactly that collision event. Is `havoid` dischargeable from the accept predicate's off-node division structure (the multiopen check degenerates at a colliding challenge), or is it a genuine model boundary? If the latter, it would help to document how the hypothesis and the `t₃` summand divide the work; if the former, a follow-up discharging it would remove the last non-floor hypothesis from the budgeted path.
2. **Docstring contradiction on the bridge.** `Bridge.lean:438` (`snarkRelation_of_memberColumns`) says the assumed payload is "now delivered on the computed path, so the bridge is retired", while `Bridge.lean:472` carefully states the opposite ("What this does *not* do is retire `ExtractableFromAcceptance`"). The `:472` wording is the accurate one; `:438` should be qualified the same way.
3. **Aggregate node binding, narrowed rerun of round 1.** The member/value-check node-binding chain is now genuinely consumed up into the capstones — resolving round 1's "proven but docstring-only" finding for that chain — but the two aggregate entry points (`NodeBinding.lean:42` `deployed_aggregate_node_binding_of_x3consistency`, `:221` `…_or_dlr`) are still statement-only with no proof-term consumers.

## Hygiene

- **Clean surface**: no `sorry`, no new axioms, no live `native_decide` (the four added occurrences are census expected-output pins naming the pre-existing Vesta point-count axiom). `Classical.choose` appears live only inside three `noncomputable def` witness extractors (`deployedMultiopenRewind_of_x4Prob`, `openedX4Rewind_of_x4Prob`, `openedMemberDecode_of_x1Prob`), widening only to the already-trusted `Classical.choice`. The 59 added `noncomputable def`s are measure-valued, polynomial-valued, or choice-selectors — proof-side machinery where that is expected; the constructivity discipline applies to the reduction producers, which are untouched.
- **Sealing instead of budget-raising**: `attribute [local irreducible]` on the deployed grouping defs (following the `preIpaTranscript` precedent) keeps index-type defeq checks syntactic, and the branch removes all 47 `set_option maxHeartbeats` overrides — the `Soundness/` tree ends with none. The two proofs that had relied on delta-reducing a sealed def are repaired with named unfolding lemmas at the definition site. This is the right trade and worth keeping.
- **Census idiom and coverage.** The 90 new `assert_no_sorry` entries and 7 pinned `#print axioms` land in `AGM/TrustBoundary.lean` using the old idiom — the very file #79 deletes and migrates to the `assert_axioms`/`assert_computable` macros. Whichever lands second inherits a translation pass; it may be cheaper for this PR to use the macros for its new entries now (the glossary change here already describes the macro discipline). Separately, five substantive new modules have public theorems with no direct census pin (`Multiopen/{Opened,RPoly,Compat,Claimed,ValueCheckDeployed}` — `Opened` includes the `Classical.choose` extractors), relying on transitive coverage through the pinned capstones; a stray axiom there would surface via the capstone pins, but the theorems are not individually asserted. Natural to fold into the #79 consolidation if not here.

## Nits

- PR description: "there's decoded capstones (**decided to make those mostly additive which doesn't hurt**) recover those columns" — the parenthetical breaks the sentence; and "doesn't doesn't".
- `glossary.md`: `assert_axiom` should be `assert_axioms`.

TalDerei:
@daira internalized and addressed feedback.

TalDerei:
@daira this is ready for review now.

TalDerei:
rebased over #85 to thread the public instance commitments.

TalDerei:
@daira sorry staging this for review now, it required a tricky rebase to pass CI after #85 landed.

book/src/formal-verification/glossary.md
```diff
@@ -85,7 +85,7 @@ development. Anchors point to a module + name under `Zcash/`.
...
+<div class="g"><div class="g-head"><span class="term">checked trust boundary</span><span class="anchor">Zcash.TrustBoundary</span> (and temporarily other TrustBoundary modules)</div><div class="def">Build-time pins on what a theorem may rest on: <code>assert_axioms</code> asserts a bound on the axioms used by a definition, so that a stray <code>sorry</code> or a new axiom fails the build instead of silently widening the trusted base. <code>assert_computable</code> additionally asserts that the definition is a plain <code>def</code>, ensuring constructivity of security reductions. Until <a href="https://github.com/zcash/ironwood/issues/71">#71</a> is done, some of the TrustBoundary modules also use a <code>#guard_msgs</code>-pinned <code>#print axioms</code> check instead of these macros. See <a href="../formal-verification.html#trust-discipline">Trust discipline</a>.</div></div>
```
daira:
```suggestion
<div class="g"><div class="g-head"><span class="term">checked trust boundary</span><span class="anchor">Zcash.TrustBoundary</span> and <code>Zcash.Fixtures.*.TrustBoundary</code> modules)</div><div class="def">Build-time pins on what a theorem may rest on: <code>assert_axioms</code> asserts a bound on the axioms used by a definition, so that a stray <code>sorry</code> or a new axiom fails the build instead of silently widening the trusted base. <code>assert_computable</code> additionally asserts that the definition is a plain <code>def</code>, ensuring constructivity of security reductions. Some of the <code>TrustBoundary</code> modules also use <code>#guard_msgs</code>-pinned <code>#print axioms</code> checks, e.g. to pin specific native axioms. See <a href="../formal-verification.html#trust-discipline">Trust discipline</a>.</div></div>
```

Zcash/Snark/Soundness/Composition/Completeness.lean
```diff
@@ -0,0 +1,549 @@
+import Zcash.Snark.Soundness.Composition.Residual
+import Zcash.Snark.Soundness.Composition.Bridge
+
+/-!
+# The honest-completeness containment for the composition ladder (`hcont`)
```
daira, line 5: Define "honest-completeness". (I think it means that a correct honest prover, as modelled in Lean, will find a proof. It does not imply completeness of the Rust prover, either in general or for any particular circuit, but it does imply &mdash;at this layer&mdash; that a completeness bug in the `halo2` prover code would be fixable without changing the proving system incompatibly.)

Zcash/Snark/Soundness/Multiopen/RPoly.lean
```diff
@@ -0,0 +1,286 @@
...
+import Zcash.Snark.Soundness.Multiopen.Deployed
+
+/-!
+# The r-polynomial core: interpolants, node values, and identity-from-samples
+
+Generic pieces for the claimed-evaluation binding: `poly_eq_of_agree_on_family` (two polynomials of
+bounded degree agreeing on enough distinct points are equal — samples force identity),
+`lagrangePoly` (the interpolant of one point set, taking the claimed value at each node), and
+`lagrangePoly_eval` (the deployed fold `lagrangeEval` is the interpolant's evaluation, so deployed
+identities transfer to the polynomial).
```
daira, line 12: Are we sure these aren't already available in CompPoly (or Mathlib)?

daira (commented):
Reviewed at branch head `9bda7ed` (post-#85 rebase), CI green under the warning-strict build.

All three round-2 questions are substantively resolved:

1. **`havoid` is paid, not assumed.** The resolution is sharper than either option the question offered: acceptance cannot discharge sample-avoidance — with `0⁻¹ = 0` a colliding challenge *degenerates* the multiopen check instead of failing it — so the avoidance is charged in the counting floor (`exists_injective_accepting_avoiding_of_measure`), retiring the hypothesis and confirming the `t₃` summand's role as exactly that payment. The module doc states the `0⁻¹` reason plainly.
2. **The `Bridge` retirement contradiction is gone** — the "bridge is retired" wording is removed and the accurate qualification stands alone.
3. **The aggregate node-binding pair is kept with an accurate disposition**: the docstring now says the discharge attempt bottoms out in the open `hx2cons` and that the live chain reaches the same conclusion from the accept floors (`deployed_value_check_node_binding`). Documented-as-superseded is a fine resolution; pruning can ride #51.

The hygiene coordination point also resolved well: the census entries moved into the consolidated `Zcash/TrustBoundary.lean` on the macro idiom (~150 `assert_axioms` entries with `+native` where the Vesta endpoints need it), and the rebase absorbed #85's instance-commitment threading and main's warning-strict CI.

Remaining open threads are the inline ones from Daira-Emma (the `Bridge.lean:28` nonzero-forcing question, the honest-completeness definition, the glossary suggestion, and whether `RPoly`'s polynomial lemmas duplicate CompPoly/Mathlib); nothing further from my side once those settle.

Zcash/Snark/Soundness/Composition/Completeness.lean
TalDerei: that's the right definition and the was i was internalizing it, and it's used in a few other places in the repo. maybe we should have a glossary entry for it eventually.

Zcash/Snark/Soundness/Multiopen/RPoly.lean
TalDerei: good observation, these are already available mathlib.

daira (approved):
Reviewed at branch head `b74c022`, CI fully green. Since round 3 the history was reshaped (fixups folded into their targets, messages rewritten), but the merge base is unchanged (`4c38b69`) and the tree differs from the round-3 head by only six files, +37/−7 — so I reviewed the PR at the tree level, and the delta is exactly the review-feedback changes. All four of Daira-Emma's open threads are now addressed in code:

1. **The `pW` nonzero-forcing question (`Bridge.lean`) is answered by withdrawal, which is the right answer.** The module doc no longer claims "nonzero for honest proofs, so it can never be forced to 0"; it now says `pW` is the `w`-weight of the AGM representation, the adversary may set it to anything, and its value is immaterial because the relation is stated at the de-blinded point `deployedCommitment − pU•u − pW•w`. I verified this against the code: `pU pW : Fp` are plain parameters with no nonzero hypothesis anywhere in the file (the only nonzero hypothesis is `hz : ν 10 ≠ 0`, the `z` challenge), and the `u`-weight — the one that does shift the opened value — goes through the `member_relation_or_dlr_of_instance` dichotomy as before. I have resolved the thread on this basis.
2. **Honest-completeness is now defined** at the top of `Completeness.lean`, matching the definition from the thread: an honest prover as modelled in Lean produces an accepting proof; no claim about the Rust `halo2` prover, only that a completeness bug there would be fixable without an incompatible proving-system change. The glossary entry Tal floated can be follow-up.
3. **The `RPoly` lemmas are documented as Mathlib repackagings, and the delegation is real**: `lagrangePoly` is literally `Lagrange.interpolate`, `lagrangePoly_eval_node` is `Lagrange.eval_interpolate_at_node`, and `poly_eq_of_agree_on_family` applies `Polynomial.eq_zero_of_natDegree_lt_card_of_eval_eq_zero`. Only the fold bridge (rewriting the deployed hand-rolled `lagrangeEval` `foldl` into the `Finset` forms Mathlib evaluates) is local, and that bridge is deployed-shape-specific — the note that switching to CompPoly would not remove it is right.
4. **The glossary suggestion is applied as proposed** (`assert_axioms`, the `Zcash.Fixtures.*.TrustBoundary` anchor, the stale "until #71" phrasing gone).

The rest of the delta is thirteen lint-required `omit [AddCommGroup G] [Module Fp G]` lines in `NodeBinding.lean` (consistent with the instance-hygiene convention) and the proof-journey pin moving to the full SHA of the new head. Nothing further from my side; ready to merge.

Issue #12, opened 2026-06-28
Schwartz–Zippel good-challenge exclusions from challenge uniformity


TalDerei:
Subtask of #11.

The soundness argument assumes the verifier's challenges avoid the Schwartz–Zippel "bad set" (the `hgood` hypotheses) — i.e. that a sampled challenge is not a root of the relevant difference polynomial. This is currently taken as a hypothesis.

Once the challenges are modeled under a random oracle (#11), these exclusions should follow rather than be assumed: a uniformly sampled challenge lands in the bad set with probability at most `d / p`, so the good-challenge condition holds with overwhelming probability.

**Scope (broad):**
- Bound the bad set for each Schwartz–Zippel use site (degree `d` over the field of size `p`).
- Derive the good-challenge condition from challenge uniformity under the random-oracle model.
- Replace the assumed `hgood` hypotheses with the derived result in the deployed soundness theorems.

---

**Stack ordering (#67 → #12 → #33).** All three discharge premises from the *same* deployed capstones, so they land in a fixed order:

- **#67 first** — it rewrites the deployed capstones into their final form (computed path, `circuitSatViaConstraints`, `ExtractableFromAcceptance` removed). #12 and #33 wire into that stable shape; done before #67, their wiring is redone when #67 rewrites the signatures.
- **#12 next** — replace the assumed `hgood` (a premise today; its failure is priced at `d/p`) in the theorems #67 re-instantiated, deriving it from challenge uniformity.
- **#33 last** — the true stack tip: its budget composition sums every sub-budget *including the extraction budget #67 produces*, so it cannot close until #67's composed bound exists. Its `hquot` half is a transport (#13), not a `d/p` exclusion.

The order-sensitive part is only the *wiring* into the capstones; the per-site pricing lemmas (#12) and budget-sum machinery (#33) are predicate-agnostic and can proceed in parallel.

TalDerei:
## Current status and scope routing

The current stack implements this issue’s Schwartz–Zippel pricing foundation:

- `hgood_failure_priced` bounds failure of the terminal capstone’s exact `hgood` implication by
  `max (deg numerator) (deg hpoly + deg) / |Fp|` under a uniform squeeze; and
- `hgood_of_good_challenge` supplies that implication for every challenge outside the priced bad
  set.

These lemmas are pinned in `TrustBoundary`. What remains is their deployed composition, now split
between #36 and #67:

[...]

PR #31, opened 2026-07-03
feat(fv): schwartz zippel good challenges

TalDerei:
**Human Summary:**

Closes https://github.com/zcash/ironwood/issues/12.

This PR proves the old `hgood` assumption instead of taking it as input. The argument is: once the proof data before the challenge `x` is fixed, only a small root set of `x` values can fool the gate check. A random-oracle challenge hits that set with probability at most `d/p`, so a good accepting `x` can be derived rather than assumed.

-----------

**AI Summary:**

This PR derives the verifier's Schwartz-Zippel good-challenge condition instead
of leaving it as an unpriced `hgood` assumption. Mental model: once the relevant
commitments are fixed, the bad `x` values are just the roots of one polynomial;
a uniform random-oracle challenge hits them with probability at most `d/p`, and
the new `_xgood` rungs spend that budget to find an accepting good challenge.

## What Is Proven

- **Bad set and degree bound**: `szBadSet C` names the roots of the constraint
  difference polynomial, `x ∉ szBadSet C` matches the old `hgood` shape, and
  root counting bounds the bad set by the relevant degree.
- **Uniformity budget**: `GoodChallenge.lean` turns the root-count bound into a
  random-oracle measure bound, including the quotient-check form with explicit
  degree `max (deg numerator) (deg h + n)`.
- **Budget spent across the whole ladder**: the pigeonhole
  (`exists_accepting_good_challenge_quotient`) derives a good accepting `x` from
  an accept-measure hypothesis, and every constraint capstone gets an `_xgood`
  twin routed through the common terminal
  `decoded_constraint_of_opening_or_relation_xgood` — the abstract, deployed,
  x4, staged (`_adaptive`), rewound (`_rewind`), attack-event
  (`_adaptive_rewind`) rungs and their `_agm_dl` wrappers. None of them takes
  `hgood`.
- **Ordering and pinning**: the `x` transcript prefix is sealed (`reprogramX` /
  `preXTranscript`), and decoded columns are pinned up to the existing relation
  branch (`decodedCols_eq_or_relation`), so the measured polynomial is fixed
  before `x` is sampled.
- **Fixtures**: toy decoded-column fixtures exercise the old bad-set shape and
  the new `_xgood` derivation path (`toy_xgood_discharged`, no `hgood` input).

## Remaining Floor

The per-instance capstones still expose `hgood` as a modular, priced hypothesis
beneath the derived `_xgood` ladder. What stays outside this PR is tracked in
#33: the gate-check content (`hquot`, the `x`→`x₃` transport) and the
end-to-end composition of all ROM / query-loss budgets. Both sit under the
Fiat–Shamir discharge (#11); composition is inherently top-of-stack, so it lands
as a new PR rather than folding into an open one.

## Issue Scope Check

Issue #12 asks to bound each Schwartz-Zippel bad set, derive good challenges
from random-oracle uniformity, and replace assumed `hgood` at the deployed
soundness endpoints. This PR does that: the bad sets are named and bounded, the
`d/p` budget is proved, and the `_xgood` rungs spend that budget instead of
assuming `hgood` — across the whole constraint ladder, not just the constant
endpoints.


TalDerei:
closing in favor of folding into #30 directly.

PR #39, opened 2026-07-08
docs(book): tighten the proof comments

### LINKED ISSUE #35: docs: tighten up the proof comments
**Motivation.** As the soundness stack has grown, the docstrings could use a tightening pass to keep them a durable record of the proof structure — clear enough that the argument stays reproducible from the comments alone, rather than living in whoever wrote it. #34 covers the top-down view (map + glossary); this is the bottom-up complement: the docstrings themselves.

**Ask.** A short comment convention, then a pass over the densest files (`Soundness/Main.lean`, `Vesta.lean`, `AGM.lean`, `Forking/*`) applying it:

- **Lead with the plain claim.** First sentence says *what* the lemma establishes, in words, before the *how*, caveats, or history.
- **Define jargon on first use**, or link the #34 glossary (*peel, rung, fixed-slot, adjusted commitment, bad set*).
- **Trim to one caveat, not a paragraph.** A nuance that needs three sentences is probably its own note.
- **Concrete over abstract** where a name or number will do.

**Scope.** Comments/prose only — no proof-term changes, so it reviews as a pure diff.

**Non-goals.** Not the map/glossary itself (#34); not a background reading list (a separate, smaller ask).

TalDerei:
Closes https://github.com/zcash/ironwood/issues/35.

Modifies the existing documentation prose on main that mostly landed as part of https://github.com/zcash/ironwood/pull/9 according to the attached skill file.


TalDerei:
@daira i'll rebase my stack of PRs in this style or a modified style you suggest after this is merged (a priority review in that respect would be good).

.claude/skills/proof-comment-style/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+2. `typos <files>` — the repo runs the typos checker (`_typos.toml`).
+3. `lake build Zcash` — docstrings are syntax; the build is the parse gate.
+
+To audit the whole tree for adherence (after a pass, before a docs PR, when new proof files
+land), sweep every `.lean` file under `Zcash/`, skipping auto-generated files (header says "Do
+not edit by hand"):
+
+- **No blanket passes.** The verdict for an unmodified file is a per-docstring justification
+  ("all ≤ 4 lines, formula-led, anchors present"), never "looks fine". Well-organised prose is
+  not exempt from rule 11.
```
daira: Avoid referring to things by fragile references. In this case: "not exempt from the rule to introduce before use".

daira (approved):
The skill file looks great, and spot-checking the changes, they look like good improvements. ACK.

.claude/skills/proof-comment-style/SKILL.md
TalDerei: fixed fragility.

TalDerei: Addressed in 8ede5d2 — "rule 11" is now "the rule to introduce before use". I extended the same fix to the one other bare rule-number reference ("rule 10" → "short by default") so no fragile cross-references remain in the file.

PR #42, opened 2026-07-10
feat(fv): enforce verifier computability in CI

### LINKED ISSUE #40: Check computability in CI
Our CI should enforce computability (at the executable layer) which is currently missing.

TalDerei:
Closes #40. 

 CI runs lake build, which builds the fixture whose `native_decide` compiles and runs the Lean verifier (so anything non-computable fails to build), while the `TrustBoundary` module additionally confirms the match theorem has no hidden sorry and rests on exactly its expected axioms.

### DIFF
--- a/.github/workflows/lean.yml
+++ b/.github/workflows/lean.yml
@@ -59,7 +59,7 @@ jobs:
       # `defaultTargets = ["Zcash", "FixtureCheck"]`. Building `FixtureCheck`
       # elaborates `Zcash.Snark.Fingerprint.Fixture`, whose
       # `fingerprint_matches := by native_decide` compiles and *runs* the Lean
-      # verifier `assemble` on the captured proof: this is the authoritative,
+      # verifier `assemble` on the captured proof. This is the authoritative,
       # whole-path computability check (anything noncomputable on that path
       # fails to compile), and `Zcash.Snark.Fingerprint.TrustBoundary`
       # additionally pins the theorem's sorry-freeness and axiom set.
--- a/.github/workflows/lean.yml
+++ b/.github/workflows/lean.yml
@@ -55,7 +55,14 @@ jobs:
         with:
           persist-credentials: false
       # Installs the toolchain from lean-toolchain, fetches the Mathlib olean
-      # cache, and runs `lake build` (defaultTargets = Zcash).
+      # cache, and runs plain `lake build`, which builds
+      # `defaultTargets = ["Zcash", "FixtureCheck"]`. Building `FixtureCheck`
+      # elaborates `Zcash.Snark.Fingerprint.Fixture`, whose
+      # `fingerprint_matches := by native_decide` compiles and *runs* the Lean
+      # verifier `assemble` on the captured proof: this is the authoritative,
+      # whole-path computability check (anything noncomputable on that path
+      # fails to compile), and `Zcash.Snark.Fingerprint.TrustBoundary`
+      # additionally pins the theorem's sorry-freeness and axiom set.
       - uses: leanprover/lean-action@38fbc41a8c28c4cbaec22d7f7de508ec2e7c0dd9 # v1.5.0
 
   lean-ci:
--- a/Zcash/Snark/Fingerprint/TrustBoundary.lean
+++ b/Zcash/Snark/Fingerprint/TrustBoundary.lean
@@ -1,0 +1,41 @@
+import Zcash.Snark.Fingerprint.Fixture
+import Mathlib.Util.AssertNoSorry
+
+/-!
+# Checked trust boundary of the fingerprint match
+
+This module is built by CI (it belongs to the `FixtureCheck` lake target) and turns the trust boundary of
+the captured fingerprint match into *checked*, build-time obligations. It complements — it does not
+duplicate — the computability guarantee that the match already carries: `fingerprint_matches` is proved by
+`native_decide`, which compiles and *runs* the Lean verifier `assemble` on the captured proof, so building
+`FixtureCheck` already fails if anything on the assembled-MSM path is `noncomputable` (Lean's own
+per-definition, whole-path determination). What a build does *not* otherwise pin down is whether the
+theorem secretly rests on a `sorry` reached through some dependency, or on an unexpected axiom. Those are
+what this module fixes in place.
+
+Both checks below follow Lean's elaborated dependency graph (via `Lean.collectAxioms`), so they see holes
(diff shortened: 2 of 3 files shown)


daira:
[Claude, lightly edited] Nice to have a computability guard. Two things though: the grep has some gaps because it's syntactic, and there's a semantic check that's likely already stronger.

### The grep is syntactic → false positives *and* false negatives

- **False positives.** `grep -wF -e sorry -e noncomputable` also matches those words in comments/docstrings/strings. This codebase documents its computability discipline, so a future docstring like *"we deliberately avoid `noncomputable` here"* would fail the lint. `-w` doesn't exclude comments.
- **False negatives (computability).** Noncomputability — or, worse, a *wrong* reduction — can reach the `assemble` path without the keyword: an `@[implemented_by]` / `@[extern]` / `opaque` shim, or a `Decidable`/instance that doesn't actually reduce. The grep sees none of these, and it never touches `native_decide`'s own trust (the `Lean.ofReduceBool` axiom). The path can also transitively use a def *outside* the three dirs; a `noncomputable` there breaks the fixture but isn't grepped.
- **False negatives (sorry).** `admit`, `sorryAx`, or a macro-generated hole aren't the literal token `sorry`; and a `sorry` in a *dependency* — the exempt `Soundness/` layer, or Mathlib — that `fingerprint_matches` / `assemble` transitively uses is invisible to a three-dir grep, even though it makes the theorem unsound.

### `lake build` is already the authoritative (and stronger) computability check

`theorem fingerprint_matches : … := by native_decide` runs `assemble` at compile time, so if anything on that path is `noncomputable` it simply won't compile — Lean's own per-definition determination, semantic and whole-path. Since `defaultTargets = ["Zcash", "FixtureCheck"]`, a plain `lake build` exercises it, so a **green build is the guarantee**, strictly stronger than the grep.

Worth confirming CI actually builds `FixtureCheck`, though: `lean.yml` comments say *"defaultTargets = Zcash"*, which reads as stale against the lakefile. If CI is only building `Zcash`, the fixture isn't being run in CI — and *that's* the real gap (more valuable to close than adding the grep).

### For the no-sorry / trust-base half, use the semantic checks

These follow the full dependency graph, so they catch what the grep can't:

```lean
import Mathlib.Util.AssertNoSorry   -- already available via `import Mathlib`
open Zcash.Snark

-- fails the build if any dependency (incl. Soundness/ or Mathlib) has a sorry:
assert_no_sorry fingerprint_matches
assert_no_sorry assemble

-- pin the exact trust base, so a new axiom (e.g. `sorryAx`) fails the guard, and so
-- `Lean.ofReduceBool` (native_decide's compiler trust) is documented as appearing only
-- in the concrete fixture, never in a general theorem:
/-- info: 'Zcash.Snark.…fingerprint_matches' depends on axioms: [ … paste what `#print axioms` currently reports … ] -/
#guard_msgs in
#print axioms fingerprint_matches
```

Put that in a small checked module (built by CI). It's a few lines, encodes the trust boundary explicitly, and catches sorries/axioms reaching the fixture from anywhere — the exact thing the grep misses.

The grep might still be useful as a fast, zero-build smoke-test, but even in that usage it will have false positives. Treat **build-the-fixture + `assert_no_sorry` / `#print axioms`** as the actual enforcement.

daira (changes requested):
Claude's suggested approach seems significantly better, both for false negatives and false positives.

daira (approved):
ACK

.github/workflows/lean.yml
```diff
@@ -55,7 +55,14 @@ jobs:
         with:
           persist-credentials: false
       # Installs the toolchain from lean-toolchain, fetches the Mathlib olean
-      # cache, and runs `lake build` (defaultTargets = Zcash).
+      # cache, and runs plain `lake build`, which builds
+      # `defaultTargets = ["Zcash", "FixtureCheck"]`. Building `FixtureCheck`
+      # elaborates `Zcash.Snark.Fingerprint.Fixture`, whose
+      # `fingerprint_matches := by native_decide` compiles and *runs* the Lean
+      # verifier `assemble` on the captured proof: this is the authoritative,
```
daira:
```suggestion
      # verifier `assemble` on the captured proof. This is the authoritative,
```

Zcash/Snark/Fingerprint/TrustBoundary.lean
```diff
@@ -0,0 +1,41 @@
...
+/-!
+# Checked trust boundary of the fingerprint match
+
+This module is built by CI (it belongs to the `FixtureCheck` lake target) and turns the trust boundary of
+the captured fingerprint match into *checked*, build-time obligations. It complements — it does not
+duplicate — the computability guarantee that the match already carries: `fingerprint_matches` is proved by
+`native_decide`, which compiles and *runs* the Lean verifier `assemble` on the captured proof, so building
+`FixtureCheck` already fails if anything on the assembled-MSM path is `noncomputable` (Lean's own
+per-definition, whole-path determination). What a build does *not* otherwise pin down is whether the
+theorem secretly rests on a `sorry` reached through some dependency, or on an unexpected axiom. Those are
```
daira:
```suggestion
theorem rests on a `sorry` reached through some dependency, or on an unexpected axiom. Those are
```

Zcash/Snark/Fingerprint/TrustBoundary.lean
```diff
@@ -0,0 +1,41 @@
...
+what this module fixes in place.
```
daira:
```suggestion
what this module rules out.
```

Zcash/Snark/Fingerprint/TrustBoundary.lean
```diff
@@ -0,0 +1,41 @@
...
+
+Both checks below follow Lean's elaborated dependency graph (via `Lean.collectAxioms`), so they see holes
+anywhere in the transitive closure — including the `Soundness/` proof layer and Mathlib — which a
+syntactic scan of the verifier sources cannot.
+
+* `assert_no_sorry` — fails the build if the named constant transitively depends on `sorryAx`. Applied to
+  `fingerprint_matches` (the captured match) and `assemble` (the verifier assembly it runs).
+* `#print axioms` pinned by `#guard_msgs` — freezes the exact axiom set `fingerprint_matches` rests on, so
+  a newly introduced axiom (or a `sorryAx` from a stray `sorry`) changes the set and fails the build. The
```
daira:
```suggestion
  a newly introduced axiom changes the set and fails the build. The
```

daira (approved):
ACK

PR #43, opened 2026-07-10
`HasNontrivialRelation` was vacuous

daira:
There always exists such a relation in a prime-order group. Show how to fix the problem (for one case) by defining the reduction as a function, not a theorem (which could be nonconstructive). This works because `NontrivialRelation` contains data (it would not solve the problem if all fields were `Prop`s, since then the structure would behave like a `Prop`). Strictly speaking the function should also be efficient, but it's sufficient to eyeball it and see that it is efficient, especially in simple cases.


TalDerei:
closing as experimental; will follow-up with proper implementation that demonstrates a knowledge soundness reduction (the spec is currently trivially satisfiable).

PR #46, opened 2026-07-11
fv: SNARK binding relation as computed data

daira:
Applies the [breaks-as-computed-data convention](https://zcash.github.io/ironwood/formal-verification.html#breaks-as-computed-data) for the binding shapes in `Soundness/CommitFold.lean`.

## The problem being fixed

`CommitmentBinding urs := Function.Injective (commit urs)` is information-theoretically false at any instantiation — the commitment compresses `2^k` coefficients into one group element — so every theorem taking `hbind : CommitmentBinding urs` was vacuous at the group (unsatisfiable hypothesis). Dually, `commitmentBinding_iff_no_relation`'s right-hand side ("every relation is trivial") is false there, and `relation_of_collision`'s content, while stated with explicit coefficients, was only consumable through those vacuous forms.

## Changes

* `Soundness/CommitFold.lean`: `NontrivialDLRelation` — the relation coefficients as data with `Prop` certificates; `NontrivialDLRelation.ofCollision` (computed from a binding collision: two distinct openings of one commitment) and `NontrivialDLRelation.ofIpaOpenings` (computed from two distinct IPA openings, via the commitment components). `relation_of_collision` and `commitmentBinding_iff_no_relation` are subsumed and removed.
* `Soundness/InnerProduct.lean`: `CommitmentBinding` and `ipaRelation_unique` removed; opening uniqueness holds up to a computed relation.
* `Soundness/KnowledgeSoundness.lean`: `knowledge_sound` drops the binding hypothesis and the `∀ a', IpaRelation … → a' = a` conjunct (which was usable only under the unsatisfiable hypothesis); uniqueness-up-to-a-computed-relation is delivered by `ofIpaOpenings`. This is the API-visible change.
* `Soundness/Main.lean`, `Soundness/Vesta.lean`: the conditional capstones drop `hbind` — their proofs never used the uniqueness conjunct, so their conclusions are unchanged.
* `Soundness/TrustBoundary.lean` (new, wired into the aggregator): `assert_no_sorry` plus `#guard_msgs`-pinned `#print axioms` for both reductions; `Classical.choice` enters only through erased `Prop` certificate fields, with computability compiler-enforced.
* Docstring updates in the touched files describe the present boundary (DLR hardness consumed only at the computational layer) without prescribing future work.

The deployed `_opening`/`_constraint` capstones and the Fiat–Shamir / forking territory are untouched; the corresponding shapes in the in-flight AGM work (#28) are that PR's to align.

### DIFF
--- a/Zcash/Snark.lean
+++ b/Zcash/Snark.lean
@@ -32,6 +32,7 @@ import Zcash.Snark.Soundness.Lookup
 import Zcash.Snark.Soundness.Permutation
 import Zcash.Snark.Soundness.PermutationConstruction
 import Zcash.Snark.Soundness.InnerProduct
+import Zcash.Snark.Soundness.TrustBoundary
 import Zcash.Snark.Soundness.Extraction
 import Zcash.Snark.Soundness.Constraints
 import Zcash.Snark.Soundness.CommitFold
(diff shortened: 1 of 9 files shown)


daira (commented):
Self-ACK

TalDerei (approved):
utACK :)

PR #44, opened 2026-07-11
fv: ledger-layer foundations — collision vocabulary, Merkle position binding, abstract Action statement

daira:
This is the first slice of the ledger-model security layer, which will eventually have a Lean formalization of the high-level Zcash security properties (Balance, Spendability, and Spend Authority) — not in this PR.

The intended approach is to define these properties in the strong adversary model of my Zcon3 talk [*Understanding the Security of Zcash*](https://github.com/daira/zcash-security) ([video](https://www.youtube.com/watch?v=f6UToqiIdeY)), composing the security arguments of [ZIP 2005](https://zips.z.cash/zip-2005). Currently we have only the core of the ZIP 2005 arguments.

This PR lands the shared vocabulary, the Merkle lemma, the abstract Action-statement interface with its pinning lemmas; and documents two development-wide conventions on modelling of security breaks and on trust discipline. The security-game theorems will build on these in follow-up PRs.

## Contents

* **Shared random-oracle collision vocabulary** (`Zcash/Security/RandomOracle.lean`): `Collision` and `CollisionUpToSign` (the ±-collision shape produced by coordinate-extractor arguments), as data-bearing structures. Shared with the in-flight key-binding workstream; this commit is the common base of both branches.
* **Merkle position binding** (`Zcash/Security/Ledger/Merkle.lean`): fixed-depth Merkle trees over path-indexed leaves (`Fin d → Bool`, so no position arithmetic). `collisionOfWrongLeaf` computes a hash collision from a validating path for a wrong leaf; `pathRoot_authPath` is the completeness direction. No hardness assumption appears; the computed collision is consumed at the computational layer (Sinsemilla/DLR pre-quantum, a collapsing hash for the ZIP 2005 rehashed tree).
* **Abstract Action statement and pinning lemmas** (`Zcash/Security/Ledger/Statement.lean`): the Orchard-shaped Action statement over abstract primitives, as used by the security games. This models both the deployed Orchard Action statement and the ZIP 2005 Recovery Statement, with key material decoupled behind `KeyBindingInterface`. Proven pinning steps: `ivk_pinned` (ZIP 2005's `ivk`-pinning lemma; pure module algebra), `nk_eq_or_break`, nullifier determinism up to a key-binding break, and the `noteCommitBreakOfNe` reduction.
[...]

Issue #49, opened 2026-07-11
Single top-level `TrustBoundary` file


TalDerei:
Currently every directory have a TrustBoundary file; instead we should have a single top-level file that segments them accordingly.

PR #50, opened 2026-07-11
feat(fv): key-binding layer — witness, pinning, and the opening-break collision

daira:
The key-binding layer of the high-level security-properties formalization, building on `RandomOracle` and `Ledger.Statement` from #44.

## What's here

Formalizes the [ZIP 2005 key-binding theorem (ROM)](https://zips.z.cash/zip-2005#thm-key-binding-rom) over the shared abstract setting: a prime-order group as an `F`-vector space, base field `B` with an `Extract` operation from group elements respecting the ±-property (if two elements extract to the same `B` value, they are equal up to sign). A Pallas instantiation will be added later via CompElliptic.

**`Zcash/Security/KeyBinding/Basic.lean`**

- `Witness`, with the qk/sk split as a `Branch` sum type in a field `qk_or_sk` — a witness carries exactly one of `qk`/`sk`, so the ⊥-structure constraint is enforced by the type; the `ivk` commitment as a Pedersen lift (`Commitivk rivk ak nk = Extract ((h ak nk + rivk) • S)`). The key-binding conditions are Prop structures with named fields — `KBOpening`, `KBDerivation`, `KB`, the opening-level `OpeningBreak`, and the full `Break`, whose `proj_ne` field uses the named `BreakProj` (`qk_or_sk`/`ak`/`nk`/`rivk`) — so consumers project by name rather than by position.
- `commit_scalar_pm` — the deterministic algebraic core: an opening collision forces the Pedersen scalars equal-or-negated (`Extract` ±-property + injectivity of `· • S`). The analogue of `Balance.lean`'s `relation_of_imbalance`.
- `CollisionUpToSign.ofOpeningBreak` — an `OpeningBreak` turned by a computable function into a `RandomOracle.CollisionUpToSign` of the Pedersen-scalar map. Documented as an *intermediate certificate*, not a break event: `pedersenScalar` is affine in `rivk`, so a standalone inhabitant can be fabricated. The security content is conditional on the `OpeningBreak` hypothesis (see the review thread on the former `commitCollision`).
- `FinalQuery` / `finalQueryOf` / `rivk_eq_finalOracle` — the ZIP's "Final-random-oracle structure" as data, and the proof that under the derivation constraints `rivk` is the combined final oracle's output at the decoded query. The external/internal choice is a decode of `rivk`'s value, with the fixpoint caveat documented on `finalQueryOf`.
- `sameIvk_finalOracle_pm` — the H^* reduction core: any two same-`ivk` openings (no distinctness needed) satisfy the break equation `h(ak₁,nk₁) + H^*(q₁) = ±(h(ak₂,nk₂) + H^*(q₂))`. `openingBreak_finalOracle_pm` and `break_finalOracle_pm` are the corollaries from an `OpeningBreak` and from the full `Break`.
- `shiftedFinalOracle` and `CollisionUpToSign.ofBreak` — the conclusion of the deterministic layer: a full `Break` computes a ±-collision of the *shifted* combined final oracle (`H^*` offset by the non-querying `hfn(ak, nk)`, which the derivation constraints make a function of the query alone) at *distinct* queries. Query distinctness is the substantive step (`residual_of_finalQuery_eq`): coinciding final queries force both witnesses internal and relocate the collision to the two `rivk_ext`-derivation queries. This isolates the load-bearing ROM use to one statement: ±-colliding the shifted oracle at distinct queries is hard. `shiftedFinalOracle` is definitionally `shiftOf q + q.eval`, and `collision_mem_shifted_pm` places a collision's output pair in the set counted by the birthday layer — the bridge to `Birthday.lean`.
- `nk_pinned`, `ak_pinned` (up to y-sign), `qk_or_sk_pinned` — two valid same-`ivk` witnesses that are *not* a key-binding break share `nk`, `ak`, and the same `qk` or the same `sk`, including which of these backs the witness. (The branch form will be strengthened in ZIP 2005.)

**`Zcash/Security/Common/RandomOracle.lean`** — gains `EqUpToSign` (an abbrev) with scoped notation `a =± b`, which now states every equal-up-to-sign disjunction: the `CollisionUpToSign` `pm` field, the `hExt` ±-property hypotheses, the ±-equation conclusions, and the `Birthday` filters.

**`Zcash/Security/Common/Birthday.lean`** — the birthday bound of the theorem, in the counting-fraction style of `fingerprint_schwartz_zippel`. A "probability" is the fraction of a finite ambient space on which a bad event occurs, in `ℚ≥0`; no probability monad. This is generic and reusable for the nullifier `H^rcm` ±-collision.

- `card_shifted_pm_collision_le` / `shifted_pm_collision_fraction_le` — the shifted ±-collision event `s₁ + o₁ = ±(s₂ + o₂)` over `F × F` has at most `2·|F|` outcomes (fraction `2/|F|`): shifts do not change the count.
- `birthday_closed_form` — `2/|F|` per pair over the `C(q,2) = q(q-1)/2` pairs of an adversary's `q` queries gives `ε_kb ≤ q·(q-1)/|F|`. The residual needs no separate term: it relocates the collision to the `rivk_ext`-derivation queries, a query pair the union bound already counts. This sharpens the previously published ZIP 2005 bound `3·q·(q-1)/(2·r)`; the ZIP proof update adopting the sharpened accounting is in progress.

**`Zcash/Security/KeyBinding/Instance.lean`** — bridges to `Ledger.KeyBindingInterface`: `KeyBinding.toInterface` exposes the concrete `Witness`/`KB`/`Break` through the games-facing view and discharges `break_of_nk_ne` from `nk_pinned`.

**`Zcash/Meta/AxiomCheck.lean` and `Zcash/TrustBoundary.lean`** — the trust claims as build-time checks. `assert_axioms` upper-bounds a declaration's axioms at the standard tier. `assert_computable` requires a plain `def` (not a theorem, not marked `noncomputable`) with axioms bounded by `propext`/`Quot.sound`, `+choice` marking the reductions where `Classical.choice` enters via erased `Prop` fields. The top-level census covers the key-binding layer and absorbs the former Ledger and BindingSignature per-module TrustBoundary files; it is imported by the library root, so a plain `lake build` enforces it.

## Remaining (follow-up)

- The Layer-C capstone: a named uniformity hypothesis for distinct final queries, consuming `CollisionUpToSign.ofBreak` and the `collision_mem_shifted_pm` bridge — to be discharged when RO semantics are modelled (#24).
- `PRF^nf`-pinning (Spendability), and wiring `ak`/`qk_or_sk` pinning into a games-facing interface (the current `KeyBindingInterface` exposes only `nk` / `break_of_nk_ne`) — to coordinate with the games side.

🤖 Claude Opus 4.8, Claude Fable 5


Zcash/Security/KeyBinding.lean
```diff
@@ -0,0 +1,369 @@
...
+
+The *deterministic reduction* `OpeningBreak ⇒ CollisionUpToSign` is the composable core;
+`openingBreak_scalar_pm` below is its algebraic heart. The probabilistic birthday bound
+`ε_kb ≤ 3q(q-1)/2r` is a separate concern.
+
+Abstract setting: a prime-order group `G` as an `F`-vector space (`F = ZMod r` the scalar field),
+a base field `B` (x-coordinates, `= ZMod q`), and `Extract : G → B` with the ±-property (`hExt`).
+
+`Commitivk` is required to have the Pedersen structure (not opaque), which is what makes the break
+reduction provable. Pallas is instantiated last via CompElliptic.
```
daira: Not instantiated yet.

Zcash/Security/KeyBinding.lean
```diff
@@ -0,0 +1,369 @@
...
+/-- The `ivk` commitment as a Pedersen lift:
+`Commitivk rivk a n = Extract ((h a n + rivk) • S)`, with `h` abstract-but-non-querying and `S` a
+fixed base. Mirrors the `NoteCommit` repair `[H^rcm + f]·R`. -/
+def Commitivk (Extract : G → B) (S : G) (hfn : B → B → F) (rivk : F) (a n : B) : B :=
+  Extract ((hfn a n + rivk) • S)
+
+omit [Field B] in
+/-- Algebraic core: two openings of the same `Commitivk` value force their Pedersen scalars to be
+equal or negatives. Uses the `Extract` ±-property and injectivity of `·•S` for `S ≠ 0` (`G` is an
+`F`-vector space). This is the deterministic content the whole key-binding reduction rests on. -/
```
daira: This is correct but I will split the injectivity of `·•S` out into a lemma. (Don't confuse this with the argument that there is only one ivk for (g_d, pk_d), which would require a bound on ivk; we haven't got to that part yet.)

Zcash/Security/KeyBinding/Basic.lean
```diff
@@ -0,0 +1,369 @@
...
+    match w.qk, w.sk with
+    | some qk, _ => .ext qk (ak Extract w) w.nk
+    | none, some sk => .legacy sk
+    | none, none => .int w.rivk_ext (ak Extract w) w.nk
+  else
+    .int w.rivk_ext (ak Extract w) w.nk
+
+omit [Field B] in
+/-- **Final-random-oracle representation** (ZIP 2005 key-binding proof, "Final-random-oracle
+structure"): under the derivation constraints, `rivk` is the output of the combined final oracle at
```
daira, line 295: @TalDerei you were asking what `finalOracle` referred to; it is in the ZIP. It's a slightly odd proof because the "final oracle" can correspond to one of three different hashes, but that's how it works and I think the complexity is inherent, not accidental.

daira (commented):
Claude knocks it out of the park again (this is not even Fable, it's Opus 4.8). There are some things I'd like to clean up and fill in, so I will leave it in draft until I can get time in front of a keyboard. Note that this is not yet connected to the high-level security properties; my Zcon3 presentation handwaved that and so did ZIP 2005, so it'll require some original work.

Zcash/Security/Common/Birthday.lean
TalDerei, line 1: grouping a few unrelated observations here:

- key binding should be in a dedicated folder,
- birthday logic should probably be moved to `common/`, alongside the random oracle per https://github.com/zcash/ironwood/pull/44#discussion_r3566369356,
- there isn't a `TrustBoundary` module, so the axiom claims aren't build enforced,

Zcash/Security/KeyBinding.lean
```diff
@@ -0,0 +1,369 @@
...
+where `rivk` is an oracle output) is per-instantiation, as with `NoteCommitBreak`. -/
+def commitCollision
+    (Extract : G → B) (S : G) (hfn : B → B → F)
+    (hExt : ∀ P Q : G, Extract P = Extract Q ↔ P = Q ∨ P = -Q) (hS : S ≠ 0)
+    {w₁ w₂ : Witness G F B SK QK} (hbrk : OpeningBreak Extract S hfn w₁ w₂) :
+    RandomOracle.CollisionUpToSign (pedersenScalar hfn) where
+  q₁ := (ak Extract w₁, w₁.nk, w₁.rivk)
+  q₂ := (ak Extract w₂, w₂.nk, w₂.rivk)
+  ne := (openingBreak_scalar_pm Extract S hfn hExt hS hbrk).1
+  pm := by simpa [pedersenScalar] using (openingBreak_scalar_pm Extract S hfn hExt hS hbrk).2
```
TalDerei: firstly as an aside, `commitCollision` isn't currently used anywhere, and i'd expect the def to live alongside its consumer.
  
moreover, tracing previous instances of our "breaks-as-computed-data" style, this seems to violate that shape in an interesting way.

`commitCollision` returns a "collision" of a map that anyone can collide on purpose (since `pedersenScalar` is affine in its r coordinate, you can architect a collision), so the collision in some sense doesn't mean anything. it's vacuous in a way our other break objects aren't, and those are self-certifying: holding one implies something hard happened. for instance,

  - holding a `NontrivialRelation` implies a break of discrete-log-relation hardness
  - holding a Merkle `Collision` H implies someone broke collision resistance of H
  - BUT holding a `CollisionUpToSign` (`pedersenScalar`) implies nothing, because you can fabricate it

in that context, i think we should harden our definition of "breaks-as-computed-data". currently we claim:
  1. Break events are structures carrying the breaking data
  2. Reductions are plain computable defs producing them

that's fine and clear, but both clauses police how the data is produced, but neither says what constitutes breaking data. i'd add a third clause:

3. Every break event type must have a nameable hardness assumption under which inhabiting it is infeasible. If no such assumption can be stated, it isn't a break event.

`commitCollision` satisfies 1 and 2 perfectly and fails 3.

Zcash/Security/KeyBinding/Basic.lean
```diff
@@ -0,0 +1,369 @@
...
+    (Hask : SK → F) (Hnk : SK → B) (Hrivk_legacy : SK → F)
+    (Hrivk_ext : QK → B → B → F) (Hrivk_int : F → B → B → F)
+    {w₁ w₂ : Witness G F B SK QK}
+    (hbrk : Break Extract S hfn Ggen Hask Hnk Hrivk_legacy Hrivk_ext Hrivk_int w₁ w₂) :
+    hfn (ak Extract w₁) w₁.nk + (finalQueryOf Extract w₁).eval Hrivk_legacy Hrivk_ext Hrivk_int =
+      hfn (ak Extract w₂) w₂.nk + (finalQueryOf Extract w₂).eval Hrivk_legacy Hrivk_ext Hrivk_int ∨
+    hfn (ak Extract w₁) w₁.nk + (finalQueryOf Extract w₁).eval Hrivk_legacy Hrivk_ext Hrivk_int =
+      -(hfn (ak Extract w₂) w₂.nk + (finalQueryOf Extract w₂).eval Hrivk_legacy Hrivk_ext Hrivk_int) :=
+  sameIvk_finalOracle_pm Extract S hfn Ggen hExt hS Hask Hnk Hrivk_legacy Hrivk_ext Hrivk_int
+    hbrk.1.1 hbrk.2.1.1 hbrk.2.2.1 hbrk.1.2 hbrk.2.1.2
```
TalDerei: `break_finalOracle_pm` proves "the two oracle outputs are equal", but not that the two queries differ?

Zcash/Security.lean
TalDerei, line 1: can we do a pass through this PR with my doc skill to refine comments further?

Zcash/Security/KeyBinding.lean
```diff
@@ -0,0 +1,369 @@
...
+  refine ⟨h₁, h₂, hivk, fun heq => hne ?_⟩
+  simpa using congrArg (fun t => t.1) heq
+
+end Full
+
+section Onward
+variable [AddCommGroup G] [Field F] [Field B] [Module F G] [DecidableEq F]
+
+/-- The final query of a witness: which final oracle produces its `rivk`, and at what input.
+Selected by the external/internal ivk choice (`rivk = rivk_ext`?) and then the qk/sk branch. -/
```
TalDerei:
```suggestion
  /-- The final query of a witness: which final oracle produces its `rivk`, and at what input.
  Selected by the external/internal ivk choice (`rivk = rivk_ext`?) and then the qk/sk branch.
  The choice is decoded from the fields, not carried as witness data: at a fixpoint
  `Hrivk_int rivk_ext ak nk = rivk_ext`, an internally-derived witness decodes as external —
  harmless for `rivk_eq_finalOracle` (which holds either way), but the birthday accounting must
  partition query pairs by this decode, not by the ZIP's derivation-based four cases. -/

**Review by TalDerei — COMMENTED** [review]
left some observations that need triage.

File: Zcash/Security/KeyBinding/Basic.lean
Comment (TalDerei) [inline Zcash/Security/KeyBinding/Basic.lean:1]: obviously the _entire_ the ROM key-binding theorem is not formalized which is the accepted scope here:

```ts
break_finalOracle_pm proves only the algebraic equation (G_1=\pm G_2). 
Separately, Birthday.lean counts independent uniform pairs and rearranges (\binom q2·3/|F|).

Nothing connects them. Missing from Lean are:
      - a random-oracle/adversary/query-transcript model;
      - independence and uniformity of the three final oracles;
      - the claimed non-querying independence of hfn;
      - distinct-final-query or residual-collision case analysis;
      - mapping queries to at most (\binom q2) witness pairs;
      - the union bound itself.
```

my intention is to fully implement the first bullet as part of #24 (fully model RO semantics possibly after [VCVio](https://github.com/Verified-zkEVM/VCVio) / [ArkLib](https://github.com/Verified-zkEVM/ArkLib)), then we can iterate on closing the remaning gaps here. let's file an issue to track these.

Zcash/Security.lean
TalDerei, line 1: also filed https://github.com/zcash/ironwood/issues/55, which might be an interesting follow-up to visually trace the interactions in the construction.

Zcash/Security/Common/Birthday.lean
daira: Got it, thanks.

Zcash/Security/KeyBinding.lean
daira: You're right, thanks. This observation drove a significant rework. A `CollisionUpToSign` of `pedersenScalar` can indeed be fabricated standalone (the map is affine in `rivk`), so *holding* one certifies nothing. It isn't vacuous, though; it's a step toward the overall reduction, and we need to be able to express those for modularity.

In any case, clause 3 as stated is too strong. At this abstraction level even `OpeningBreak` is inhabitable: `rivk` is a free witness field, so the same affine solving can be used to efficiently fabricate two valid openings of one `ivk`. A per-type "nameable hardness assumption" test would reject most of the deterministic layer. Strictly speaking, it would even reject `Collision H` for a *fixed* keyless compressing hash: collisions of such a hash provably exist, and no cryptographic hardness assumption forbids holding one. What makes exhibiting one meaningful is that plausibly no-one *knows* such a collision (Phillip Rogaway, *Formalizing Human Ignorance: Collision-Resistant Hashing without the Keys*, <https://eprint.iacr.org/2006/281>). Incidentally that paper is the ancestor of our breaks-as-computed-data convention: state theorems as explicitly-given reductions. (I'm pretty sure I had read it and was vaguely remembering it.)

What I've adopted instead, judged per-instantiation rather than per-type:

* A **terminal break event** must, where it is instantiated, name the assumption or model under which producing an inhabitant is infeasible. These events are where a security argument is concluded. For example: DLR hardness for `NontrivialRelation`, or for this argument the birthday bound over the shifted final oracle.
* An **intermediate certificate** —the output of a partial reduction— only has to be computed data. It is consumed by a further reduction, and never named or presented as a break by itself.

Concretely, in this PR (to be pushed next): `commitCollision` and `openingBreak_scalar_pm` will be folded into a single def `CollisionUpToSign.ofOpeningBreak`, documented as an intermediate certificate. This also resolves your unused/ live-with-its-consumer aside. The layer now has a genuine terminal object, `CollisionUpToSign.ofBreak` (see the other thread). The convention-text hardening (terminal/intermediate + the Rogaway citation) belongs in the book's formal-verification page.

Zcash/Security/KeyBinding/Basic.lean
daira: Good question! The answer turned out to be: not always, and the exception is the interesting case. Final-query distinctness is *not* derivable in general; two internally-derived witnesses can share `(ak, nk, rivk_ext, rivk)` and differ only in the qk/sk branch, giving a genuine `Break` with equal final queries. But in that case `rivk_ext₁ = Hrivk_ext(qk, ak, nk) = Hrivk_legacy(sk) = rivk_ext₂` is itself a collision of the combined oracle at *distinct* `rivk_ext`-derivation queries, and the shifts agree. So the collision relocates rather than disappears.

This is now computed rather than asserted (addressed in the branch, will push next): `CollisionUpToSign.ofBreak` returns a ±-collision of the shifted final oracle at distinct queries in every case, with `residual_of_finalQuery_eq` as the deterministic case analysis. That closes the "distinct-final-query or residual-collision case analysis" bullet from your scope list. `break_finalOracle_pm` remains as the bare ±-equation the def builds on.

Zcash/Security/KeyBinding/Basic.lean
daira: Agreed with the list and with modelling RO semantics in #24. One update: the "distinct-final-query or residual-collision case analysis" bullet is now deterministic and formalized (`CollisionUpToSign.ofBreak` / `residual_of_finalQuery_eq`; will push next), and the union-bound side stays Layer C as named hypotheses (`Birthday.lean` has the closed form). A useful consequence of that factoring: the load-bearing ROM use is isolated to a single statement — ±-colliding the *shifted* final oracle (`H^*` offset by the non-querying `hfn`, which the derivation constraints make a function of the query alone) at distinct queries is hard. Yes, let's file the tracking issue.

Zcash/Security/Common/Birthday.lean
daira: The TrustBoundary point is addressed (will push next): there is now a single top-level `Zcash/TrustBoundary.lean` census, enforced by the library root import. The key-binding computed reductions are checked by a new `assert_computable` macro (plain `def`, axiom tier bounded by `propext`/`Quot.sound`, with `+choice` marking the reductions where `Classical.choice` enters via erased `Prop` fields). The theorems are checked by `assert_axioms`. I've folded in the Ledger and BindingSignature per-module TrustBoundary files. The module organization questions (dedicated key-binding directory; birthday alongside the random oracle in `common/`) are still open.

Zcash/Security/KeyBinding.lean
daira: Applied, adapted to the current shape of the file (will push next). The witness now carries a `qk_or_sk : Branch QK SK` sum field, so the ⊥-structure half of the decode is type-enforced and only the external/internal choice remains a decode of the value. The fixpoint caveat is stated on `finalQueryOf` essentially as you wrote it. Thanks.

Zcash/Security/Common/Birthday.lean
TalDerei: great! references https://github.com/zcash/ironwood/issues/49.

Zcash/Security/KeyBinding.lean
TalDerei: ACK directionally

TalDerei (approved):

TalDerei:
ACK on review comments, optimistically approving.

Zcash/Security/KeyBinding/Basic.lean
daira: Filed #68 to track the remaining ROM-model work, with the list updated for what the deterministic layer now covers.

Zcash/Security/Common/Birthday.lean
daira: The module moves are in now too (addressed, will push shortly): `Zcash/Security/Common/` holds `RandomOracle` and `Birthday`, and `Zcash/Security/KeyBinding/` holds `Basic` and `Instance`. Namespaces are unchanged. Together with the trust-boundary census earlier in this thread, that resolves all three observations.

Zcash/Security/KeyBinding.lean
daira: Addressed (will push shortly): `commit_scalar_pm` now uses Mathlib's `smul_left_injective` instead of deriving the injectivity inline.

Zcash/Security/KeyBinding.lean
daira: Reworded (will push shortly): the docstring no longer claims the instantiation — the development is entirely abstract, with CompElliptic as the intended concrete route.

Zcash/Security.lean
daira: Done — we ran the pass with the proof-comment-style skill over the branch's files (docs-only commit, will push shortly). Besides general tightening, it caught two stale claims — the residual sub-case was still described as probabilistic after `residual_of_finalQuery_eq` made it deterministic, and the `RandomOracle` module docstring still had the old birthday constant — and restructured the Basic module docstring into a route map.

daira: A readability follow-up on the same theme, now pushed: the key-binding conditions (`KBOpening`, `KBDerivation`, `KB`, `OpeningBreak`, `Break`) are Prop structures with named, individually documented fields, the break projection is a named `BreakProj` structure (so the pinning proofs say `congrArg BreakProj.nk` instead of `fun t => t.2.2.1`), and `EqUpToSign` gained the scoped notation `a =± b` used across the ±-equations and the Birthday filters.

PR #56, opened 2026-07-14
feat(fv): FS adversary

### LINKED ISSUE #11: Discharge the Fiat-Shamir assumption (random-oracle model + forking/rewinding)

TalDerei:
**Human Summary:**

Closes #11.

PR #56 makes #28’s AGM reduction usable by a real Fiat–Shamir adversary: it models bounded queries and grinding, computes fork certificates and AGM instances, and derives the relation/DL bound. [VCVio](https://github.com/Verified-zkEVM/VCVio) was a foundational influence in this modeling the core of the adversary (query machine) and some of the bounds. 

----------

**AI Summary:**

This PR models a bounded-query algebraic Fiat–Shamir adversary against the deployed transcript. Specifically it connects the bounded-query Fiat–Shamir adversary to the representation-carrying AGM interface defined by #28. Unlike the existential probability wrappers in #24, it computes the fork certificate and deployed AGM instance from explicit oracle and extractor coins.

## What Is Proven

- **Adversary model:** supports arbitrary transcript queries and finite private coins while preserving the deployed Fiat–Shamir game.
- **Recursive extraction:** computes a valid `AlgebraicDForkCert` and packages a `DeployedAlgebraicForkingInstance` from standard AGM data.
- **Binding reduction:** turns every binding mismatch into an explicit relation, either directly or from a clean-opening commitment collision.
- **Probability accounting:** bounds binding probability by query/extraction loss, the `z = 0` slice, and `|basis|` times the discrete-log advantage.
- **Runtime:** proves totality and an efficient expected-run bound under fork spread.

## Remaining Floor

Blake2b and generator hash-to-curve are modeled as random oracles, and Vesta discrete log remains the computational assumption. The unconditional sparse-fork/grinding composition needed to remove the fork-spread runtime condition is not yet formalized.

## Issue Scope Check

This completes the operational producer left to #56 by #28. The distinct opening and constraint results from #24 remain as proof-level lemmas.


daira (approved):
Reviewed at branch head `707159f` (15 commits over #28), with the full stack built green locally.

## Verdict

No showstoppers. This PR closes the floor that #24's own docs named as the missing piece — a real bounded-query Fiat–Shamir adversary, with query loss charged and the fork certificate computed rather than posited — and lands it as an executable chain from the adversary machine to per-family DL-hardness endpoints. The efficiency gap that remains (no field-independent polynomial call bound) is stated accurately in the code, the book, and the glossary. Recommend merging in stack order after #28.

## What it establishes

- **The adversary model** (`Adversary/OracleComp`, ~1600 lines): a free-monad query machine `OracleComp T F α` with whole-table eager RO semantics — the oracle is a uniform total function, reprogramming is `Function.update`, and repeat queries are handled by deduplication. Query loss is charged only at fresh points, justified by the update-blind measure lemmas rather than assumed. The structural `QueryBound` is credited to VCVio's design.
- **The `3/|F|` constant is proven, not conventional**: the ladder-escape analysis in `Forking/Tree` shows an accepted run either extracts or draws some round's challenge from a set of at most three (zero plus two extractable-transitions), giving `fsAdvantage_le : ≤ (Q + k)·(3/|F|)`, with a local forking bound (the `ε² − ε/N` shape) in both counting and measure form in `Forking/Probability`.
- **The executable extractor** (`Adversary/Recursive`): `recursiveAlgebraicForkFrom` reprograms at round transcript points, rejects changed trunks, and requires three distinct nonzero challenges; `DForkCert.treeRuns_eq` pins the certificate at exactly `3^k` accepting leaves; `algebraicForkCertAttempt_valid` proves every returned certificate satisfies #24's `DeployedForkValid`.
- **The trunk-stability crux is discharged, not assumed.** The generic recursive layer takes `stable`/`stable_update` as parameters, but the deployed instantiation (`Adversary/Algebraic`) proves them: the pre-IPA transcript is pinned across rewinds, and a reprogrammed round point can never alias one of the eleven pre-IPA squeeze points, by `preIpaTranscript_inj` and the positional length lemmas added to `Forking/Ordering`. This is where the `preIpaLen` arithmetic earns its keep.
- **Provenance is derived from the deployed MSM**: `multiopenMsm_points_mem` proves the `AssemblyPoint` enumeration by peeling the deployed `assembleOpening`, so the every-point-has-a-representation premiss of `AlgebraicWfProof` is a theorem about the real verifier, not an assumption about it.
- **The probability composition is complete**: `AGM/ProbabilityCoins` extends #28's slot accounting to private extractor coins with the same counting and coin-bijection discipline; `Adversary/DomainReduction` transfers the bound from bounded transcripts to arbitrary domains via the finite reachable-support split; `Adversary/Adaptive` removes junk queries at a `(Q+1)/|F|` zero-slice cost.
- **The endpoints state the intended bound verbatim**: acceptance without a clean opening (and dually, the binding event) has probability at most `(Q + k)·(3/|Fp|) + (Q + 1)·(1/|Fp|) + |AugmentedIndex(2^k)|·ε` under `DiscreteLogRelationHardFor` — deliberately a *per-family* hardness statement, since a family-universal form would range over computationally unbounded solvers; PPT-ness of the family remains the documented external premiss. The generator-RO variants close the loop with #28's `orchard_uniformURSIdentification_of_generatorRO` (with `hB : B ≠ 0` carried).
- **Binding is the right event**: the computable mismatch between the effective accepted value —including the `z⁻¹·(multiU + ξ·sU)` shift from the adversary's declared `U` coefficients— and the carried aggregate opening; never nonexistence of an opening, which is vacuous at prime order. A clean opening on a mismatch run converts to a commitment-collision relation rather than being discarded.
- **The efficiency floor is accurate**: the unconditional call bound is `(2·|F|+1)^k` (`reductionEfficient_exponential`); the geometric `(6·|F|/(σ₀−1))^k` expected-run bound is conditional on the `ForkSpread` density hypothesis and is not plumbed into the endpoints; `ExpectedRunsPoly` holds proven node-level pieces of a polynomial accounting but no global theorem. All three facts are stated where a reader will meet them.
- **Trust boundary**: 169 `assert_no_sorry` checks plus guarded `#print axioms` pins over the whole computed chain, recording exactly which endpoints inherit the Vesta point-count native axiom (`bindingWin_unbounded_measure_le`, `runToSnark`, and the concrete probability endpoints). The diff adds no `sorry`, no `Classical.choose`, and no new `native_decide` anywhere; the one `noncomputable` exception (`globalReachSet` / `splitFamilyRand`, a `Fintype.ofFinite` materialization) is documented in the book.
- **The legacy surface is demoted cleanly**: the eight Vesta forking rungs and the three Rewind soundness wrappers are renamed `legacy_*` with pointers to the computed path, and the module docs that described the missing adversary experiment now describe the layer boundary instead.

## Candidate follow-up issues

1. **The polynomial AFK call bound** is now the single load-bearing efficiency gap: the endpoints are gated on `ReductionEfficient R`, dischargeable today only at the exponential level, with the `ForkSpread`-conditional geometric bound proven but unplumbed and `ExpectedRunsPoly` off the soundness DAG. Worth its own tracking issue so the floor does not live only in docstrings.
2. **Legacy cleanup once the stack lands**: the `legacy_*` ladder (Vesta and Rewind) has no consumers, and the single-point `reprogram` family from #24 remains unused. One deletion-or-wiring pass over the demoted surface would shrink the audit area considerably.
3. **Unused new lemmas** (repo-wide reference count 1, verify before deleting): `adaptive_fork_mem_le`, `schedule_fork_bound` and its three private helpers, `fsAdvantage_le_of_forcing` and the `escapesDuring` path superseded by `escapesDuringC`, `recursiveAlgebraicFork_runs_le_instance_pow`, `fsAdvantage_pure`, and `truncateTranscript` with its two lemmas (guarded in the trust boundary but unconsumed).

## Nits (no action needed)

- The Blake2b field-conversion bias is explicitly excluded from the bound (`Forking/Oracle` module doc) — same residue as #24's follow-up; no new issue needed if that one is filed.
- `Adversary/ExpectedRunsPoly` is guarded by the trust boundary but consumed by nothing; fine as parked work given the docstrings say so.

## Rebase note

Inherits #24's `Soundness/Main.lean` conflict transitively; #56's own touches there are docstring-level and sit near the same region, so the resolution guidance from #24 (keep main's restated forms) covers it.

TalDerei:
@daira folded up the cross-PR follow-ups in https://github.com/zcash/ironwood/issues/70.

PR #57, opened 2026-07-15
chore(fv): check concrete Post-NU6.3 verifier MSMs

### LINKED ISSUE #54: CI generated autogen
After upstream fixture generation merges, we'll have CI check out upstream halo2/orchard crates and checks that they regenerate the same fixtures.

When an upstream pin changes intentionally, the regenerated fixture changes should be reviewed and committed with the pin update. The upstream feature-gated Rust generators remain the source of truth; Ironwood does not need a separate verifier implementation.

TalDerei:
Closes https://github.com/zcash/ironwood/issues/54. Consumes upstream changes in https://github.com/zcash/halo2/pull/914 and https://github.com/zcash/orchard/pull/531 in a fixture yml file.

The previous fingerprinting implementation had a few issues this resolves:
- ironwood fixtures used a real Halo2 Orchard circuit and real proofs, but wrong circuit version for NU6.3 deployment,
- Lean fingerprints represented vesta points as opaque natural numbers (hashmap mapping from points to natural numbers for some lean kernel performance optimization reasons i thought were clever early on which just aren't) rather than real curve points and and didn't check point equality which raised some injectivity concerns there, 
- didn't use a `PinnedVerifierKey` (Halo2's canonical identity for a verifying key), and although VK came from actual Rust VK, it wasn't independently checked against canonical pinned VK,
- we had top-level capstones that modeled fingerprint acceptance as MSM = 0, but we didn't actual compute the MSM and check that equality (we assumed it based on some symbolic representation), so this now actually computes it.

These are all fixed and documented now!


ebfull (approved):
Optimistic ACK.

TalDerei:
cc @daira for visibility.

PR #58, opened 2026-07-16
fv: bump CompElliptic to the unconditional Pasta order; drop the Vesta Hasse assumption

daira:
Updates the CompElliptic dependency to the merged unconditional-Pasta-order version (daira/CompElliptic#6) and adapts the Vesta soundness layer, removing its last curve assumption.

CompElliptic now proves the Pasta group orders with **no assumption** — an elementary point-count bound (`#E(F) ≤ 2·#F + 1`) stands in for Hasse, which Mathlib lacks. So on the ironwood side:

- `vestaOrder` and the `Fact VestaOrder` instance are now unconditional (from `Pasta.Vesta.card_eq`, no hypothesis);
- the four Vesta capstones drop their `[Fact (HasseBound Vesta.curve)]` binder;
- the checked axiom pin for `ofUnopenedForkVesta` shrinks from four CompElliptic `native_decide` facts to one (the `≈2^254` prime-order witness), since `Gpt`/`Gpt_ne_zero` upstream now discharge by kernel `decide`.

The dependency bump (`lakefile.toml` + `lake-manifest.json`) and the Vesta adaptation are one commit because they are atomically coupled — the old `Vesta.lean` uses the now-removed `HasseBound`.

Builds green (`lake build`, 8383 jobs).


daira (commented):
Self-ACK

Issue #60, opened 2026-07-16
Unsatisfiable ∀ a quantification in the deployed-verifier reduction theorems


ItsViolet:
The constraint-side hypotheses in `orchard_verifier_deployed_opening_of_forked` and `orchard_verifier_deployed_constraint_of_forked` (`Soundness/Main.lean:320-378`) are quantified over every mathematical opening `a : Fin (2^k) → Fp` of the commitment `(P, b, v)`:

```lean
(hcirc : ∀ a, IpaRelation urs fs.openedCommitment b (multiopenValue vk ps ch) a → circuitSat a)
(hquot : ∀ a, IpaRelation ... a → quotientCheck ...)
(hgood : ∀ a, IpaRelation ... a → ...)
```

At a prime-order curve the set of such `a` is an affine subspace of dimension `≥ 2^k − 2` (two linear conditions on `2^k` coefficients), so any `circuitSat` that genuinely reads the witness fails on almost all of it. **The premises are unsatisfiable as the theorems carry no gate-level
content.**

The comment at [Main.lean:280-294](https://github.com/zcash/ironwood/blob/c2174c4679b04affa113a774f690ce4c6fa0b694/Zcash/Snark/Soundness/Main.lean#L280-L294) already documents this:

> The implication: the `_of_forked` theorems below currently carry no gate-level content — they conclude `S` only for `circuitSat` instantiations that ignore the witness, so no soundness for the deployed circuit's constraints follows from them yet.

## Proposed cause

The proof extracts a single `a` via `ipaRelation_of_acceptV` (line 332) but the hypotheses are
stated over **all** `a`:

```lean
obtain ⟨a, hrel⟩ := ipaRelation_of_acceptV urs b fs.openedCommitment
  (multiopenValue vk ps ch) (projTree fs.tree) hclean
exact hencodes a ⟨hrel, hcirc a hrel⟩
```

Only this one `a` is ever used. `∀ a` is both unnecessary and unsound.

## Impact

- `circuitSatViaGates_of_check` derives `circuitSat` for *one* `a` from that `a`'s own gate check — but the `∀ a`-quantified hypotheses demand it for all openings, so the derivation is unusable.
- The constraint-side capstones (`_of_forked`) are semantically vacuous: they conclude `S` only for `circuitSat` instantiations that ignore the witness.
- Issue #18 is blocked in the sense that the tracked decode gap cannot be meaningfully closed while the premises are unsatisfiable.

## Proposed fix

[...]

PR #61, opened 2026-07-16
Fixing unsatisfiable ∀ a quantification in the deployed-verifier reduction theorems

viabhinav:
This is a fix for #60 

### What changed

- `hcirc` changed from `∀ a, IpaRelation ... a → circuitSat a` → `circuitSat a`
- `hquot` changed from `∀ a, IpaRelation ... a → quotientCheck ...` → `quotientCheck ...`
- `hgood` changed from `∀ a, IpaRelation ... a → ...` → `...`
- `hclean` parameter dropped (it only fed `ipaRelation_of_acceptV`; caller now passes `hrel` directly)
- `hrel : IpaRelation ... a` added as explicit parameter
- `a : Fin (2 ^ urs.k) → Fp` added as explicit binder

### What stayed

- `hencodes` stays `∀ a`-quantified — `S` is a free `Prop` and the caller says "for ANY witness satisfying the SNARK relation, conclude `S`". This was not the problem.

# Explaining the fix

After the fix, `circuitSatViaGates_of_check` (KnowledgeSoundness.lean:84) can meaningfully derive `circuitSat` from the verifier's gate check for the one extracted witness `a`. The hypotheses `hquot`/`hgood` now constrain exactly that single `a`, no longer an unsatisfiable`∀` quantifier over an affine subspace.

The tracked decode gap (linking `decodeAdvice`/`decodeInstance` to the multiopen via `batch_open_soundV`) remains separate work and is no longer blocked by this unsatisfiable premise.

### DIFF
--- a/Zcash/Snark/Soundness/Main.lean
+++ b/Zcash/Snark/Soundness/Main.lean
@@ -313,25 +313,20 @@ def NontrivialRelation.ofUnopenedFork [DecidableEq G] [Inhabited G] {shape : Sha
 cleanly accepted, `ipa_soundV` extracts the opening witness for the declared opening
 `fs.openedCommitment` — the blinded opening of the pinned commitment,
 `deployedCommitment = ⟨a, g⟩ + [pU]u + [pW]w` with `⟨a, b⟩ = multiopenValue`; the circuit side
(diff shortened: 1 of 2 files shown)


TalDerei:
ACK thanks!

Issue #68, opened 2026-07-20
Model ROM semantics for the key-binding theorem's probabilistic layer


daira:
Tracking the gap between the deterministic key-binding layer (#50) and the full ZIP 2005 key-binding theorem (ROM). The deterministic layer ends at a computed break event: `CollisionUpToSign.ofBreak` turns any key-binding `Break` into a ±-collision of the shifted combined final oracle at distinct queries, and `collision_mem_shifted_pm` places its output pair in the set `Birthday.lean` counts (per-pair fraction ≤ 2/|F|, closed form ε_kb ≤ q(q−1)/r). What remains is the probabilistic model that turns those counting facts into a probability statement:

- [x] a random-oracle / adversary / query-transcript model — realized as the whole-table eager ROM (a uniform `PMF` over `FinalQuery → F` tables) with the adversary as an `OracleComp` machine and its query list as the transcript; #24's transcript model (possibly via VCVio / ArkLib) can later refine this;
- [x] uniformity and independence of the combined final oracle's outputs at distinct queries — supplied by the whole-table measure itself (fresh-read conditioning: `uniformOfFintype_cond_point`, `uniformOfFintype_point_mem_blind_le`) rather than consumed as a named hypothesis;
- [x] the conditioning step for the shifts — the shift is data of the query alone (`shiftOf`, with `hfn` non-querying), and the uniform point measure is shift-blind (`uniformOfFintype_point_mem_blind_le`, feeding `pair_shifted_collision_measure_le`);
- [x] averaging over the adversary's private randomness and the `H^ask`/`H^nk` tables — `toOuterMeasure_bind_le` / `break_measure_le_mixture`;
- [x] the wlog that the adversary queries every derivation input its witnesses use — a construction rather than an assumption (`OracleComp.completing`), giving the hypothesis-free capstone `break_measure_le_of_queryBound` at (n+4)(n+3)/|F|;
- [x] the union bound over the C(q,2) query pairs — static `finset_shifted_collision_measure_le` (q(q−1)/|F|) and adaptive `break_measure_le_adaptive` (n(n−1)/|F| over the recorded trace);
- [x] stating the five oracle tables as independent uniform draws (the product model over `H^ask`, `H^nk`, and the three rivk-derivation oracles) — the mixture capstone conditions on `H^ask`/`H^nk` under an arbitrary joint distribution; the product form was stated in #90 (`break_measure_le_product`, ZIP 2005's probability space in full);
[...]

PR #73, opened 2026-07-20
fv: key-binding birthday bound in the whole-table ROM

daira:
The probabilistic layer of the key-binding formalization (tracking issue #68). It turns the deterministic layer's computed collision event (#50) and the `Birthday.lean` counting facts into probability bounds, in two stages — a static (non-adaptive) model and the full adaptive bounded-query model.

**Model.** The combined final `rivk`-derivation oracle is sampled as one uniform table `O : FinalQuery QK SK B F → F` (`Fintype` via the constructor partition `finalQueryEquiv`). The three constituent oracles are its constructor restrictions (`eval_restrict`). A uniform table has independent uniform outputs at every query, and the shift `shiftOf` reads only the query and the fixed parameters, so a shifted pair read at distinct queries is still uniform. The uniform-table and conditioning lemmas come from `Snark.Soundness.Forking.Probability` (#56) — generic content imported across the Security/Snark boundary. Relocating it to a shared home is possible follow-up.

**Non-adaptive results** (the query set is fixed before the table is sampled):
- `pair_shifted_collision_measure_le` — shifted ±-collision at one distinct pair: ≤ `2/|F|`. Stated for an arbitrary finite query domain and arbitrary table-independent shift, for reuse by the Spendability `H^rcm` collision.
- `finset_shifted_collision_measure_le` — union over the `C(q,2)` unordered pairs of a query set of size ≤ q: ≤ `q·(q−1)/|F|` (exact, via `powersetCard 2`).
- `break_measure_le` — an adversary whose witnesses' derivation queries lie in the static set produces a key-binding `Break` with probability ≤ `q·(q−1)/|F|` — ZIP 2005's ε_kb at `|F| = r`.

**Adaptive results** (the adversary is a bounded-query `OracleComp` machine choosing each query after seeing earlier answers):
- `escapesWithHistory_measure_le` — a history-indexed escape bound: per-step escape probability `f(history length)` gives `≤ Σ_{i<n} f(m+i)` for a cache-avoiding n-query machine. This generalizes `escapesDuringC_measure_le` (the constant-budget case); #74 tracks unifying them once the stack's in-flight work has landed.
- `queries_pair_collision_measure_le` — at the `badList` sets (at most two candidate answers per recorded pair, one per sign) the triangle sum recovers the non-adaptive constant exactly: a trace ±-collision has probability ≤ `n·(n−1)/|F|`.
- `break_measure_le_adaptive` — the key-binding break, for a machine that queries its output witnesses' derivation inputs: ≤ `n·(n−1)/|F|` (ZIP 2005's accounting convention).
- `break_measure_le_of_queryBound` — hypothesis-free, for any n-query machine: ≤ `(n+4)·(n+3)/|F|`, via `OracleComp.completing` appending the four derivation queries — the textbook "wlog the adversary queries what it outputs" as a construction rather than an assumption.

The adaptive argument follows the Davies–Meyer collision analysis (Boneh–Shoup, *A Graduate Course in Applied Cryptography* v0.6, §8.5.3, Theorem 8.4). Their "reasonable adversary" normalization is the completion; their event Z is the trace collision, with the same per-pair factor 2 (from ⊕ there, ± here). The corresponding prose for ZIP 2005 is drafted in the working notes.

**Trust boundary.** `Zcash/TrustBoundary.lean` extends the census over the whole new surface (`assert_computable` for the data definitions, `assert_axioms` at the standard tier for the theorems). Full build green.

**Known deferrals** (per #68):
- The triple-of-oracles ↔ single-table uniformity transport as a stated `PMF` equality (currently the identification is definitional via `eval_restrict`).
- Distinguishing the domains of `ak`/`nk`/`rivk` etc. (not all `B` in practice).
- Wiring the bounds into the ledger-games layer.

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Security/KeyBinding/Probability.lean
+++ b/Zcash/Security/KeyBinding/Probability.lean
@@ -1,0 +1,249 @@
+import Zcash.Security.KeyBinding.Basic
+import Zcash.Security.Common.Birthday
+import Zcash.Snark.Soundness.Forking.Probability
+
+/-!
+# Key binding in the whole-table random-oracle model (Layer C)
+
+The deterministic key-binding layer ends at a computed event: `CollisionUpToSign.ofBreak` turns
+any key-binding `Break` into a ±-collision of the shifted combined final oracle at two distinct
+queries, and `collision_mem_shifted_pm` places its output pair in the set `Birthday.lean` counts.
+This module adds the probabilistic model that turns those counting facts into a probability
+bound.
+
+The model samples the combined final `rivk`-derivation oracle `H^*` as one uniform table
+`O : FinalQuery QK SK B F → F`. A uniform table is the same as independent uniform outputs at
+every query — the product structure of the uniform measure. (In Markov-category terms `PMF` is a
+commutative monad, so samples reorder freely; see Fritz [eprint arXiv:1908.07021] and
+`Mathlib.CategoryTheory.MarkovCategory` for the abstract setting.) The three constituent oracles
+are the table's restrictions along the `FinalQuery` constructors (`FinalQuery.eval_restrict`).
+
+The remaining data (`Extract`, the bases, `hfn`, `Hask`, `Hnk`) are fixed parameters. The shift
+`shiftOf` reads only the query and these parameters, never the sampled table — so a shifted read
+of the table at distinct queries is still a uniform pair. (The shift is a deterministic, a.k.a.
+copy-preserving, map in the Markov-category sense.) The pair and union lemmas below are stated
+for an arbitrary finite query domain and arbitrary shift function, for reuse by the Spendability
+`H^rcm` collision.
+
+Results (milestone 1 — a *static* query set, fixed before the table is sampled):
+
+* `pair_shifted_collision_measure_le` — a shifted ±-collision at one distinct query pair has
+  probability at most `2/|F|`.
+* `finset_shifted_collision_measure_le` — union bound over the `C(q,2)` unordered pairs of a
+  query set of size at most `q`: probability at most `q·(q−1)/|F|`.
+* `break_measure_le` — the key-binding capstone: an adversary whose witnesses' derivation
+  queries land in the static set produces a `Break` with probability at most `q·(q−1)/|F|`
+  (ZIP 2005's `ε_kb` at `|F| = r`).
+
+The more realistic adaptive attack model —a bounded-query machine choosing queries after seeing
+earlier answers— is the next milestone (issue #68). The fresh-read conditioning it needs is in
+`Snark.Soundness.Forking.Probability`, which this module already imports for the uniform-table
+lemmas.
+-/
+
+namespace Zcash.Security.KeyBinding
+
+open Finset Zcash.Security.RandomOracle Zcash.Security.Birthday Zcash.Snark
+open scoped ENNReal
+
+variable {G F B SK QK : Type*}
+
+/-! ## Finiteness of the query space -/
+
+/-- `FinalQuery` as the sum of its constructors' data. -/
+def finalQueryEquiv : FinalQuery QK SK B F ≃ (QK × B × B) ⊕ SK ⊕ (F × B × B) where
+  toFun
+    | .ext qk ak nk => .inl (qk, ak, nk)
+    | .legacy sk => .inr (.inl sk)
+    | .int rivk_ext ak nk => .inr (.inr (rivk_ext, ak, nk))
+  invFun
+    | .inl (qk, ak, nk) => .ext qk ak nk
+    | .inr (.inl sk) => .legacy sk
+    | .inr (.inr (rivk_ext, ak, nk)) => .int rivk_ext ak nk
+  left_inv q := by cases q <;> rfl
+  right_inv x := by rcases x with ⟨qk, ak, nk⟩ | sk | ⟨rivk_ext, ak, nk⟩ <;> rfl
+
+instance [Fintype QK] [Fintype SK] [Fintype B] [Fintype F] :
+    Fintype (FinalQuery QK SK B F) :=
+  Fintype.ofEquiv _ finalQueryEquiv.symm
+
+/-- Assembling the combined final oracle from a table's constructor restrictions gives back the
+table. -/
+theorem eval_restrict (O : FinalQuery QK SK B F → F) :
+    FinalQuery.eval (fun sk => O (.legacy sk)) (fun qk ak nk => O (.ext qk ak nk))
+      (fun rivk_ext ak nk => O (.int rivk_ext ak nk)) = O := by
+  funext q; cases q <;> rfl
+
+/-! ## The per-pair and union bounds, over an arbitrary finite query domain -/
+
+section Generic
+
+variable {Q : Type*} [Fintype Q] [DecidableEq Q] [Field F] [Fintype F] [DecidableEq F]
+
+/-- **Non-adaptive per-pair bound.** For distinct queries `a ≠ b` and any shift `s` fixed
+independently of the table, a uniform table's shifted outputs at `a` and `b` ±-collide with
+probability at most `2/|F|`: the pair read is uniform on `F × F`
+(`uniformOfFintype_map_precomp_injective`), and the shifted collision set has at most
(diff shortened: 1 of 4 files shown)


daira (commented):
Self-ACK. Some of the proofs are long; I'll see if I can improve that. Ready for review as-is, though.

Zcash/Security/KeyBinding/Probability.lean
```diff
@@ -0,0 +1,644 @@
...
+its collision. -/
+def derivQueries (Extract : G → B)
+    (w : Witness G F B SK QK × Witness G F B SK QK) : Fin 4 → FinalQuery QK SK B F :=
+  ![finalQueryOf Extract w.1, finalQueryOf Extract w.2,
+    extQueryOf Extract w.1, extQueryOf Extract w.2]
+
+/-- **Adaptive key-binding bound (ZIP 2005 accounting).** An `n`-query machine that queries its
+output witnesses' derivation inputs produces a key-binding `Break` with probability at most
+`n·(n−1)/|F|` — ZIP 2005's `ε_kb` at `|F| = r`, with the adversary's queries chosen adaptively. -/
+theorem break_measure_le_adaptive (Extract : G → B) (S : G) (hfn : B → B → F) (Ggen : G)
```
TalDerei, line 577: current ZIP 2005 theorem (https://zips.z.cash/zip-2005#thm-key-binding-rom) doesn't yet seem to state this bound, so this a spec divergence / bug?

Zcash/Security/KeyBinding/Probability.lean
```diff
@@ -0,0 +1,644 @@
...
+
+omit [Fintype Q] [DecidableEq Q] [Field F] [Fintype F] [DecidableEq F] in
+/-- Queries of a sequenced machine: the first machine's, then the continuation's on its
+output. -/
+theorem queries_bind {α β : Type*} (A : OracleComp Q F α) (f : α → OracleComp Q F β)
+    (O : Q → F) :
+    (A.bind f).queries O = A.queries O ++ ((f (A.run O)).queries O) := by
+  induction A with
+  | pure a => simp [OracleComp.bind, OracleComp.queries]
+  | query t k ih => simp [OracleComp.bind, OracleComp.queries, ih]
```
TalDerei: restatement of `OracleComp.queries_bind` (in `Adversary/OracleComp.lean`).

Zcash/Security/KeyBinding/Probability.lean
```diff
@@ -0,0 +1,644 @@
...
+output witnesses' derivation inputs produces a key-binding `Break` with probability at most
+`n·(n−1)/|F|` — ZIP 2005's `ε_kb` at `|F| = r`, with the adversary's queries chosen adaptively. -/
+theorem break_measure_le_adaptive (Extract : G → B) (S : G) (hfn : B → B → F) (Ggen : G)
+    (hExt : ∀ P R : G, Extract P = Extract R ↔ P =± R) (hS : S ≠ 0)
+    (Hask : SK → F) (Hnk : SK → B)
+    {A : OracleComp (FinalQuery QK SK B F) F (Witness G F B SK QK × Witness G F B SK QK)}
+    {n : ℕ} (hQ : A.QueryBound n)
+    (hqueries : ∀ (O : FinalQuery QK SK B F → F) (i : Fin 4),
+      derivQueries Extract (A.run O) i ∈ A.queries O) :
+    (PMF.uniformOfFintype (FinalQuery QK SK B F → F)).toOuterMeasure
```
TalDerei, line 584: `break_measure_le_adaptive` and `break_measure_le_of_queryBound` prove probability only over:
  - one uniform table for the three final `rivk` oracles,
  - a deterministic `OracleComp`,
  - fixed `Hask` and `Hnk`

that's clear. but a review flagged a soundness boundary gap: 

"That is not yet ZIP’s probability over adversary coins and all five independent oracle tables. The pointwise result should lift by conditioning and averaging, but no such theorem exists. The stated triple-oracle transport deferral does not
  cover private coins or random Hask/Hnk.

This is a missing bridge, not an inconsistency in the Lean theorem. Either add the product/mixture capstone or scope the PR description to “deterministic adversary, conditioned on Hask/Hnk.”

TalDerei:
Reviewed the probabilistic layer end-to-end (statement-level against `KeyBinding/Basic.lean`, `Birthday.lean`, and the `OracleComp`/conditioning scaffolding). The core argument checks out — conditioning at fresh coordinates via `applyUpdates`-blindness, the pair-to-trace reduction's case analysis, the Gauss-sum accounting, and the residual-case coverage via all four `derivQueries` are all sound. A few non-blocking notes, mostly about making model boundaries explicit:

1. **Deterministic adversary — add to deferrals.** `OracleComp` machines carry no private coins. Information-theoretically fine (a randomized adversary is a mixture of deterministic ones and the bound is uniform over machines), and the FS layer already has the averaging pattern (`Adversary.Algebraic`), but the KB capstones state nothing for randomized adversaries. Worth a line in the deferral list so the Layer-B wiring doesn't silently assume it.

2. **`hfn`/`Hask`/`Hnk` as adversary-known fixed functions — make the independence assumption explicit.** Treating them as fixed parameters rather than sampled oracles is a strictly stronger adversary model (free full tables, zero query charge), so the bound is conservative — but only under Sinsemilla-vs-PRF^expand independence in the ROM. If `hfn`'s oracle were correlated with the `rivk` table, "fixed shift" would be wrong. Belongs next to the whole-table deferral note.

3. **Steer consumers to `break_measure_le_of_queryBound`.** `break_measure_le_adaptive`'s `hqueries` is `∀ O` — the machine must query its outputs' derivation inputs on *every* table, not just breaking ones, which is realistically discharged only by construction (`completing`). A docstring sentence on the adaptive capstone would prevent someone from trying to prove `hqueries` directly for a natural adversary.

4. **Placement nit:** `eqUpToSign_symm` is generic `EqUpToSign` theory and belongs in `Security/Common/RandomOracle.lean` beside the definition, rather than in the `Generic` section here (where it needs `omit` gymnastics). Same follow-up bucket as the Security→Snark import relocation already flagged in the PR body.

5. **Accounting caveat (fine as-is, just noting):** `n` counts only combined-final-oracle queries, while ZIP 2005's `q` presumably counts all RO queries — so `n ≤ q` and the bound transfers. But when Layer B composes multiple oracle families, the query budgets must be partitioned once, not double-charged.

🤖 with the help of Claude Fable 5

TalDerei:
i think every review for these PRs should be directly checked against ZIP-2005 for spec, soundness, and impl-correctness bugs to prevent any divergence like some of the minor ones flagged here.

Zcash/Security/KeyBinding/Probability.lean
daira: It is not a bug, the formalization just proves a stronger bound. I'm tracking this as a pending update to ZIP 2005: https://github.com/zcash/zips/issues/1335

TalDerei (approved):
Optimistically approving since everything was relatively minor here. mergeable after they're addressed.

Zcash/Security/KeyBinding/Probability.lean
daira: Right — addressed by fff814e.

Zcash/Security/KeyBinding/Probability.lean
daira: Agreed that the pointwise capstones condition on the adversary's private randomness and the $\mathsf{H^{ask}}$ / $\mathsf{H^{nk}}$ tables. Addressed by cda05d6, taking the "add the product/mixture capstone" branch:

* `toOuterMeasure_bind_le` shows that a slice-uniform measure bound survives `PMF.bind` averaging.
* `break_measure_le_mixture` applies it with an index `ι` bundling the private coins and both outer tables under an arbitrary joint distribution, keeping the $(n+4)(n+3)/|\mathbb{F}|$ bound.

Stating the five oracle tables as independent uniform draws (and the triple <-> table transport) remains tracked in #68.

TalDerei:
this looks good!

TalDerei (approved):

TalDerei (approved):
ACK

daira (commented):
Self re-ACK

PR #76, opened 2026-07-21
chore(docs): document lean verification in the README

### LINKED ISSUE #63: lake build in the readme
It'd be nice to have instructions on formally verifying everything with a single command communicated in the readme.

TalDerei:
Add a "Verifying the proofs" section explaining that `lake build` is the single command that re-checks the whole formalization, with the one-time `elan` install and `lake exe cache get` prerequisites, and noting what the build additionally enforces (no-sorry/axiom trust boundaries and the `native_decide` fixture verifier run).

Closes #63.


daira (approved):
ACK, doc-only.

Issue #36, opened 2026-07-07
Fold into permutation + lookup arguments


TalDerei:
The current open stack of PRs doesn't account for https://github.com/zcash/ironwood/pull/21, which landed after the https://github.com/zcash/ironwood/pull/9 that the stack builds on. This was deliberate, keeping everything framed in terms of simpler gate checks makes the integration with @mitschabaude easier to reason about. It'll need to be folded in as follow-up work once this suite of PRs lands. cc @daira

TalDerei:
## Consolidated scope

This comment supersedes the earlier discussion on this issue.

### The modelling gap

The deployed verifier folds

```text
allExpressions = gateE ++ permE ++ lookupE
```

into `expectedHEval`, while the current constraint capstone's `gates` parameter models only
`gateE`. Consequently, `hfp` currently equates a gates-only Horner fold with the deployed
gates-plus-permutation-plus-lookup fold and cannot hold at `gates := vk.gates`.

This is a constraint-representation gap, not a capture gap. All required inputs are already present:

- `vk.gates`, `vk.lookupInputExprs`, and `vk.lookupTableExprs`;
- `vk.permutationChunks`, `vk.delta`, and `vk.chunkLen`;
- `ps.permutationSetEvals`, `ps.permutationCommonEvals`, and `ps.lookupEvals`; and
- the challenge record.

No fixture re-emission or fingerprint recapture is required.

### Work owned by #36

- Extend the folded constraint representation used by `combineGates`,
  `SnarkRelationWithMemberColumns`, and the terminal capstones to cover `gateE ++ permE ++ lookupE`.
  This is a representation change rather than an attempt to force `permE` and `lookupE` into the
  existing `Expr` shape.
- Discharge `hfp` as a per-instance expression-list equality. `hfp_of_expressions_eq` reduces the
  fold equality to that list equality, which should be checked in `Fixtures/*/Faithfulness.lean`
  beside `vk_expression_refs_in_range`, using `native_decide`.
- Restate `hfold_of_expectedHEval_binding`, `hfold_of_vanishing_slot_binding`,
  `hfold_of_member_budget`, and the terminal capstone wrapper over the extended folded object.
- Restate `hgood`'s antecedent over the extended `combineGates` representation.

The other `hfold` inputs are already discharged: `hbind` comes from the good branch of
`deployed_member_budget` at the routed vanishing slot, and `hxn` follows from deployed acceptance
[...]

PR #78, opened 2026-07-21
doc(fv): add human-readable guide; fix overclaims

### LINKED ISSUE #47: Human-readable guide to the theorems
The proof map shows the full structure of the verifier-soundness development, but almost all of it is internal detail — lemmas the machine generates and checks, which no human needs to read. What a human auditor *does* need to read is the trusted surface: the top-level theorem statements, the definitions those statements are phrased in, and the assumptions they are conditional on.

## Remaining work

**1. An entry page for the trusted surface.** One page ordered by what an auditor must decide, linking into the pages above rather than restating them. A reader arriving at the book should reach, in one hop, the list of everything they must personally check.

**2. A consolidated assumption register.** The assumptions are individually disclosed today, but spread across Oracle/Model, AdaptiveStatementModel, ProbabilityVesta, the security definitions, the glossary, and the proof journey. Consolidate them into one linked table separating:

- *terminal model assumptions* — online AGM, Blake2b-as-ROM, Vesta DLOG hardness, and the fixed-URS / hash-to-curve heuristic;
- *concrete deployment bindings* — transcript and VK byte encoding, and the Challenge255 `PMFEventBiasLE` instantiation;
- *tracked engineering and proof residuals* — resource certification and circuit-to-ledger composition.

## Non-goals

Proof mechanics, the internal lemma DAG (the proof map already covers navigation), and per-file API docs.

## Done when

A reader can, from the book alone, identify everything they must personally check to trust the result — and understand what each of those things says — reaching it from one entry page, with the assumptions in one register rather than distributed across the chapter.

cc @ebfull

TalDerei:
Closes #47.

The book gains a Guide to the Ironwood Formalization page (`book/src/formal-verification/guide.md`): what a reader should and should not conclude from this formalization — the trusted surface and the assumptions, stated in plain language — rather than the internal proof mechanics, which live in the proof map. The register of what you are trusting covers the discrete-log assessments for both Pasta curves (Vesta for the SNARK's knowledge soundness; Pallas for the Action circuit and the ledger properties built on it), the idealizations (random oracle, algebraic adversary, and byte-level encodings — stated as tractability choices of this formalization, with their trade-off), the sampled-bases caveat, the Fiat–Shamir byte layer, and the compiler-trusted native facts together with the fast native arithmetic their evaluations run.

Alongside the guide:

- the ledger-scope section reflects the delivered group-hash indifferentiability (#199), pointing at that chapter and at #188 for the remaining oracle-access wiring, and states the strength of the games-to-circuit gap (no definition is yet shared between the two developments);
- the two Clean-integration design notes (`clean-boundary.md` and `lawfulness-and-certificate-elimination.md`) move out of the source tree into a Working Notes subsection of the book, edited out of their session-notes register and checked against the tree: the boundary note records the one current violation of its rule (`VkMatch.lean`), and the lawfulness note describes the arc as it stands after #145, with #202 as the work in progress;
- the hash's name is written BLAKE2b throughout prose and doc comments, as the spec and RFC 7693 write it (Rust identifiers and the digest-pinned fixture captures keep their existing bytes);
- the extraction passage states that the prover runs once — the deployed route extracts straight-line;
- the Ledger Security Games page draws the games-to-circuit hand-off as the open gap it is: the diagram's bridge edge is now a red dashed "not yet formalized" edge with its own legend entry, and the prose states that the witness-supply requirement is a modelling assumption of the games until the #147/#155 glue lands.

The guide's drafting history is squashed into the opening commit: the branch is doc-only and long-lived, and its intermediate texts and messages described development states — the decode gap, the missing copy and lookup constraints — that are already fixed in the code this lands on. The branch is rebased onto main; the contract-layer commits that belonged to #191 dropped out when it merged.

🤖 Claude Fable 5


book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,223 @@
...
+This is deliberately *not* a tour of the proof mechanics, the internal lemma
+DAG, or the per-declaration API. Those live in the [proof map](proof-map.md),
+the [proof journey](proof-journey.md), and the Lean source itself. Here we stay
+on the *trusted surface*: the top-level statements, the definitions you need to
+read them, and the boundary between what Lean proves and what you are trusting.
+New terms are collected in the [glossary](glossary.md); the two development-wide
+conventions (["breaks as computed data"](../formal-verification.md#breaks-as-computed-data)
+and the ["trust discipline"](../formal-verification.md#trust-discipline)) are on
+the [Formal Verification](../formal-verification.md) landing page.
+
```
daira:
```suggestion

This page focuses on claims relating to the knowledge soundness of the
Action circuit with respect to its intended statement. For a map of how that
will be used to prove some of the high-level security properties of the Zcash
protocol, see the [Security Definitions](security-definitions.md) page.

```

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,223 @@
...
+## What is being claimed, in one paragraph
+
+For the deployed Halo 2 verifier that Orchard/Ironwood runs over the Vesta
+curve, a proof the verifier *accepts* can be turned — by an efficient
+extractor — into a witness that both **opens the IPA commitment** and
+**satisfies the circuit's gates**. The only escape hatch is that the prover
```
daira:
```suggestion
**satisfies the circuit's constraints**. The only escape hatch is that the prover
```
Satisfying only the gate constraints without copy and lookup constraints would be meaningless. Tracked in #36.

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,223 @@
...
+The soundness statements come in a deliberate ladder, from a fully abstract
+*conditional* form down to the concrete deployed verifier at Vesta. Reading them
+in order shows exactly what each layer adds.
+
+### Conditional soundness — `orchard_verifier_sound_conditional`
+
+The top of the ladder (`Zcash/Snark/Soundness/Main.lean`) takes acceptance as an
+**opaque `Prop`**. It says: *if* an accepting proof yields extraction data (the
+hypothesis `ExtractableFromAcceptance` — a consistent transcript, an
+`IpaRelation` opening, and a circuit witness), *and if* gate satisfaction
```
daira:
```suggestion
`IpaRelation` opening, and a circuit witness), *and if* constraint satisfaction
```

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,223 @@
...
+  relation.
+* **An ideal random oracle** for Blake2b and for field-element challenge
+  conversion — and, on the generator-RO endpoints, for the hash-to-curve URS
+  derivation. Identifying the deployed hash with a random oracle is external to
+  Lean (see below).
+* **The algebraic group model (AGM)** for the computed reduction: the adversary
+  declares algebraic representations of the group elements it outputs.
+* **CompElliptic's Vesta point-count.** A closed numeric fact discharged by
+  `native_decide`, which adds one compiler-trust axiom. Concrete-curve endpoints
+  inherit it; the per-theorem `#print axioms` pins record exactly where.
```
daira: This is not the only `native_decide` axiom, so if we want to mention compiler-trust axioms at all, we shouldn't single out the curve order.

We rigorously prove the curve order, we just don't prove it depending only on the Lean kernel. In practice it's vanishingly unlikely that the order I [originally computed in Sagemath](https://github.com/zcash/pasta/blob/f0f7068552a3565786cb338448cb58bc36a8314a/amicable.sage#L184) by a completely different method, just happens to be wrong in a way that lines up with a Lean compiler bug (and also in a way that wouldn't have been noticed immediately when implementing the curve arithmetic).

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,223 @@
...
+  relation among the fixed generators $(g, U, W)$. This is where every "unclean
+  fork" and "no clean opening" branch is discharged. Commitment binding is the
+  same assumption in another guise: two openings of one commitment *is* a
+  relation.
+* **An ideal random oracle** for Blake2b and for field-element challenge
+  conversion — and, on the generator-RO endpoints, for the hash-to-curve URS
+  derivation. Identifying the deployed hash with a random oracle is external to
+  Lean (see below).
+* **The algebraic group model (AGM)** for the computed reduction: the adversary
+  declares algebraic representations of the group elements it outputs.
```
daira: These are not computational hardness assumptions. They are restricted-adversary heuristics.

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,223 @@
...
+* **Verifying-key correctness** (input side): that the `VerifyingKey` handed to
+  the verifier faithfully encodes the real deployed circuit. Assumed, not proven
+  — see the checklist.
+
+Two model limits are worth stating plainly. Reduction **efficiency** is the one
+property Lean does not express: it is argued by inspection (the reductions are
+straight-line manipulations of their inputs), and a *field-independent
+polynomial* bound on the extractor's adversary-run count is still open — the
+present bound is the exponential $(2\cdot|F|+1)^k$ fallback, with a polynomial
+$(6/\delta)^k$ available only under a fork-spread hypothesis. And adversary
```
daira: This is likely impenetrable jargon to most readers.

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,223 @@
...
+
+1. **The computational assumptions hold.** Discrete-log relation hardness on
+   Vesta, the random-oracle idealization of Blake2b and challenge conversion
+   (and hash-to-curve for the URS), and the AGM for the computed reduction.
+2. **The `native_decide` oracles are correct.** The fingerprint match
+   `fingerprint_matches` and CompElliptic's Vesta point-count. Both are closed,
+   re-checkable numeric facts; a miscompiled or buggy oracle would be caught by
+   an independent recomputation disagreeing. The
+   [trust discipline](../formal-verification.md#trust-discipline) pins them at
+   build time via `assert_no_sorry` and a `#guard_msgs`-frozen `#print axioms`.
```
daira:
```suggestion
   build time.
```

daira (commented):
I think this needs quite a bit of work. There's way too much jargon and it's not clear who the intended audience is. To understand it currently they'd need to be experts on formal verification and security analysis (and it arguably falls between two stools, being insufficiently precise for the experts). Something that is claimed to be "a human-readable guide to the theorems" should probably assume less background.

TalDerei:
> There's way too much jargon and it's not clear who the intended audience is

this is fair, and I'd like to do another pass to make it as simple to understand as possible. I wasn't really thinking about an "intended audience" beyond wanting it to contain everything a human needs to check technically. the human-readable form here should be the subset of the proof map that someone actually has to read.

TalDerei:
@daira did another pass in much simpler prose, with the jargon removed. I don't feel strongly about this version, so feel free to defer reviews / push your own changes.

LucidSamuel:
@TalDerei took a stab at this and did a rewrite on `[theorem-guide-v3](https://github.com/zcash/ironwood/compare/main...LucidSamuel:ironwood:theorem-guide-v3)` mainly addressing the following nits: 

- #91/ #100 added permutation and lookup arguments, closing #36. #30 tied the IPA witness to the
  verifier columns, and the page's decode-gap anchor, `batch_open_soundV`, is now gone.
- `c5c6013` removed the conditional-rewind ladder, along with the old exponential extractor bounds so the extraction section now describes a route that no longer exists.
- Lean also now derives the verifying key and certifies it against the capture.
- and twelve of the anchor table's nineteen rows no longer resolve:- thirteen dead symbols, plus two
  dead paths.

there is also one substantive correction rather than drift, `specPost_to_ledger`
(`Zcash/Security/Ledger/Bridge.lean:783`) already gets from `SpecPost` to either an explicit
Sinsemilla escape or a valid Orchard ledger action. and lastly, the ledger witness is existential in `Prop`, so it cannot be passed as data into the next reduction. which is
the problem tracked by #147.

the rewrite also changes the presentation rather than just resyncing names. It names the intended
reader, defines commitments and openings before using them. I kept Lean names out of the prose and put them in one
anchor table at the end.

TalDerei:
> @TalDerei took a stab at this and did a rewrite on `[theorem-guide-v3](https://github.com/zcash/ironwood/compare/main...LucidSamuel:ironwood:theorem-guide-v3)` mainly addressing the following nits:

these are great suggestions, integrating thanks!

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,210 @@
...
+covers the coined terms, [Security Models](security-models.md) states the models in full, and
+the [Knowledge-Soundness Contract](knowledge-contract.md) reads the theorem field by field.
+This page is about the Action circuit; for the protocol properties built on top of it, see
+[Ledger Security Games](ledger-security-games.md).
+
+## The claim, in one sentence
+
+> If the deployed verifier accepts a proof, then whoever produced that proof could have
+> handed you the secret data it is a proof *about* — unless they solved a problem the field
+> believes is infeasible, or hit an event the theorem shows is vanishingly unlikely.
```
daira: This has several problems. As a first try:

```suggestion
> handed you the secret data it is a proof *about* — unless they solved a problem that
> most cryptographers believe has been infeasibly difficult (although quantum computers
> could change this in a few years), or hit an event the theorem shows is vanishingly unlikely.
```

Alternatively, eliding less detail which I think may be preferable here:
```suggestion
> handed you the secret data it is a proof *about* — unless they solved a discrete
> logarithm problem (which most cryptographers believe is infeasibly difficult in the
> absence of a working quantum computer), or hit an event the theorem shows is
> vanishingly unlikely.
```

There's also a more fundamental difficulty which is that I think that statements of this kind are inherently overclaims: they don't acknowledge modelling gaps. If there is a knowledge-soundness-related problem with the protocol despite the formalization, it's at least as likely to be due to a modelling gap as a discrete logarithm break.

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,212 @@
...
+have to be assessed by other means.
+
+It is not a walkthrough of the proofs. The [proof map](proof-map.md) shows how the results
+connect, the [source map](source-map.md) indexes the tree, [definitions](definitions.md)
+covers the coined terms, [Security Models](security-models.md) states the models in full, and
+the [Knowledge-Soundness Contract](knowledge-contract.md) reads the theorem field by field.
+This page is about the Action circuit; for the protocol properties built on top of it, see
+[Ledger Security Games](ledger-security-games.md).
+
+## The claim, in one sentence
```
daira:
```suggestion
## The claim
```
I don't believe it's possible (or advisable to try) to write the claim in one sentence.

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,212 @@
+# Guide to the Theorems
+
+This page is for anyone deciding how much to trust Ironwood's verifier-soundness proof. It assumes one
+idea from cryptography — that a proof system lets someone convince you of a claim without
+showing you the data behind it — and nothing else: not Lean, nor Halo 2, nor any background in
+proof systems. It tries to explain the modelling assumptions and heuristics the verifier
+soundness claim rests on that are not established by a formal theorem, and therefore
+have to be assessed by other means.
```
daira: There's something about this prose that feels very AI-generated. Needs to be less "I'm going to tell you a thing very plainly" followed by something that is actually quite imprecise.

```suggestion
# Guide to the Ironwood Formalization

This page is for anyone deciding what to conclude, and what not to conclude, from this
formalization. This section is general; later sections of the page focus on the verifier
knowledge-soundness proof.

## The place of formalization in security assurance

A protocol formalization such as this one is not a "proof of security". It is a set of
claims that constrain the difficulty of various classes of attacks, relative to breaking
the cryptographic problems on which the protocol's security was intended to rest.
The classes of attacks it rules out are not exhaustive; they are relative to a particular
model of what adversaries are able to do.

Sometimes a formalization is presented as something that *has* to be complete in order
to be valuable. This idea might come from the fact that the formal logic in which it is
grounded has a "[principle of explosion](https://en.wikipedia.org/wiki/Principle_of_explosion)",
such that if we make a mistake then it is possible to prove "false". In that case, it would
be possible to prove that a protocol is secure when it is not. But this does not take into
account how protocol formalization actually works. How it works is that we progressively
extend the model to cover ways in which the protocol could be vulnerable. If a flaw lies
in a particular piece, and we try to formalize that piece within the framework we've set,
then we won't be able to do it, and then we are very likely to find the flaw. This can only
really give confidence once the framework is mostly complete —which it is— because we need
to know that we are proving the right kinds of thing, that will eventually be composable.

Formalization cannot replace other forms of security assurance: informal security arguments;
conservative cryptographic design; detailed and complete specification; incident response
processes; careful software engineering practices; unoptimized reference implementations;
thorough code reviews; internal and third party audits; automated and manual testing;
supply chain integrity checks, etc. On the contrary, it should be interwoven with them, for
instance by checking that the formalized objects are consistent with the specification and
(where possible) that they match the same test vectors as the production implementation and
the unoptimized reference.

So, modelling gaps and the long-term task of closing them, gaining more and more confidence
as we go —and then maintaining the formalization in sync with the spec as the protocol
evolves— are not a side-issue; that process is the entire point. It is a process that finds bugs
just like any other assurance technique, but it is deeply more thorough than conventional
auditing or testing. If, as a protocol designer, you're having sleepless nights about a class of
potential security problems in a specific part of the protocol, then competently formalizing
that part gives you a very meaningful improvement in assurance. And it *is* feasible to be
thorough enough in coverage to *approach* what we might naively expect from "proven security"
more closely than by other methods.

On the flip side, there have been cases where formalization was used as an excuse to skip
other forms of assurance, which is positively harmful. See for example Nadim Kobeissi's paper
[Verification Theatre: False Assurance in Formally Verified Cryptographic Libraries](https://eprint.iacr.org/2026/192).
Even when we don't make the kind of mistakes described in Kobeissi's paper, it's easy to be
speaking in good faith and still end up overclaiming — as we have seen a number of times
during this formalization effort. It takes substantial labour to educate readers about what
they should *not* conclude from a formalization, and that is as much a matter of good
technical writing and teaching as of proof.

In particular, it turns out that AIs are very good at writing a lot of Lean code (and
comments) quickly. That is what made this development possible, but it also presents a
huge challenge to reining in complexity, verbosity, and overclaims. The path of least
resistance is to let the AI produce more output than is humanly reviewable. Using a
proof-oriented language like Lean makes this less terrible than it could be, but it is still
terrible: not only does it hide the wood for the trees; it introduces some biases in the
direction of "it's complicated, but trust us" that we need to actively resist. If this
formalization oversimplifies or fails to properly explain something, that's a deficiency
and needs to be fixed. We are still in the process of removing a lot of redundancy and
rewriting excessively jargon-laden documentation. This page aims to help by focussing
attention on the things most important for a human reader to assess.

# Scope

There are three foci for this formalization effort: the Action circuit, the Halo 2 SNARK,
and the ledger layer (the high-level protocol security properties). Ideally they should
fit together. They don't quite do so yet.

## Scope at the ledger layer

The formalization of *Spendability* and *Spend authority* properties is not yet
finished. It needs significant attention to the way honest parties are modelled: the
properties need to be strengthened by giving the adversary oracles that allow it to
control how the honest parties create transactions, generate keys (without telling the
adversary those keys), etc. That is how these properties were originally defined at
Zcon3 in [Understanding the Security of Zcash](https://raw.githubusercontent.com/daira/zcash-security/main/zcash-security.pdf).

The formalization of Balance integrity is in much better shape, and all the important pieces
have been completed. Modelling honest parties is not an issue for Balance properties; the
adversary merely has to exhibit an unbalanced consensus-valid ledger. There are still some
potential gaps, however, described below and on the [Ledger Security Games](ledger-security-games.md)
page.

For tractability, the modelled ledger also abstracts away from the real protocol in
several directions:

* The ledger security arguments are not yet properly tied into the Action circuit knowledge
  soundness proof end-to-end ([#147](https://github.com/zcash/ironwood/issues/147)
  and [#155](https://github.com/zcash/ironwood/issues/155)).
* The real protocol has, at the time of writing, six chain value pools: the transparent pool;
  the lockbox (also transparent), and four shielded pools for Sprout, Sapling, Orchard, and
  Ironwood. The modelled ledger has one transparent pool, and one Orchard-protocol shielded
  pool (similar to either Orchard or Ironwood). We argue that this simplification is enough
  to capture realistic attacks, because the interaction of all other pools with a given
  shielded pool is as though the other pools are transparent. (It is a bit of an
  oversimplification for Orchard and Ironwood because they share addresses and key material;
  we may decide to either model this explicitly or justify more rigorously why that isn't
  needed.)
* Modelling of the hash-to-curve is not complete. This also particularly affects the
  formalization of the Spendability and Spend authority properties, because they depend
  on diversified address generation using $\mathsf{DiversifyHash}$, and so a realistic
  adversary has to be able to invoke the hash-to-curve. There are some technical issues
  involved in this modelling that have not been fully resolved, but we have a plan for how
  it will work, starting with proving indifferentiability for the concrete hash-to-curve
  construction used by Orchard ([CompElliptic#25](https://github.com/daira/CompElliptic/issues/25)).

# Circuit and SNARK layers

## Guide to the Circuit and SNARK Theorems

The verifier knowledge-soundness proof for the Orchard protocol (and therefore the
Ironwood pool) assumes one idea from cryptography — that a proof system lets someone
convince you of a claim without showing you the data behind it. This idea can be understood
without background in Lean, Halo 2, or any background in either formal or succinct proof
systems. The rest of this page tries to explain the modelling assumptions and heuristics
the verifier soundness claim rests on that are not established by a formal theorem, and
therefore have to be assessed by other means.
```

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,212 @@
...
+hashing everything sent so far, the **Fiat–Shamir transform**. The prover can compute them
+too, but only after fixing the earlier messages they derive from.
+
+**Knowing, not merely existing.** A weaker claim would say a witness *exists*. Not enough for
+a payment system: the statement behind an Orchard action can be true while the person proving
+it is not entitled to spend — the note is real and somebody holds the key, just not them. So
+the claim is that *this* prover holds a witness, made precise by demanding an **extractor**, a
+procedure that produces the witness given access to the prover.
+
+Proving that outright would settle open problems in complexity theory. What can be proven is a
```
daira: This is presumably referring to P-vs-NP, but that's not relevant. If P ≠ NP were proven tomorrow, it wouldn't necessarily lead to practical cryptosystems, and more to-the-point, if it did then Orchard still would not be using those cryptosystems. If you meant a proof of classical hardness for discrete log, that would both be astoundingly unlikely and still would not resolve the quantum threat.

```suggestion
It is vanishingly unlikely that the hardness of the problems on which Orchard's security
depends could be proven outright. What can be proven is a
```

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,212 @@
...
+trade: any prover that cheats can be turned, cheaply and mechanically, into a solver for a
+problem believed infeasible. That trade is a **reduction**, and every theorem here has that
```
daira:
```suggestion
problem believed (for now) infeasible. That trade is a **reduction**, and every theorem here has that
```

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,212 @@
...
+shape. The infeasible problem is always the same — given the fixed list of curve points, find
+multipliers, not all zero, that cancel them to zero, a **nontrivial discrete-log relation**.
```
daira: We can't accurately say this, because we are working in the AGM+RO model. I know it explains that later, but we have to refer to it in order to avoid over-claiming.
```suggestion
shape. Within the algebraic-adversary model that we consider, the assumed-infeasible
problem is always the same — given the fixed list of curve points, find multipliers,
not all zero, that cancel them to zero: a **nontrivial discrete-log relation**.
```

book/src/formal-verification/theorem-guide.md
```diff
@@ -0,0 +1,212 @@
...
+### The acceptance test
+
+Lean assembles the weighted sum in the same order Halo 2's Rust verifier does. Acceptance
+means the assembly succeeds, the structural checks pass, and the sum is zero. A first proof
+step rewrites that compact test into the explicit equation the rest of the argument consumes.
+
+### Classifying one accepting proof
+
+An equation holding says nothing about who knew what. What closes the gap is the
+algebraic-adversary assumption below: the prover must show its work, declaring for every curve
```
daira: The AGM+RO is required by *this* formalization, it is not a necessity in general. There are standard-model formalizations of nontrivial protocols.

TalDerei:
ACK changes (reviewed your comments last night), good to merge after rebasing and resolving merge conflicts and we can iterate.

daira (approved):
Self-ACK

daira (approved):
Self-ACK

Issue #71, opened 2026-07-20
Consolidate trust-boundary checks onto the AxiomCheck macros and update the book's description


daira:
Main now has `Zcash/Meta/AxiomCheck.lean` (the `assert_axioms` / `assert_computable` macros, with `+choice` / `+native` opt-ins) and the consolidated top-level census in `Zcash/TrustBoundary.lean` (#50). The Snark-side trust-boundary files (`Soundness/TrustBoundary.lean`, `Soundness/Deployed/TrustBoundary.lean`, `Soundness/AGM/TrustBoundary.lean` on the in-flight stack, and `Fixtures/*/TrustBoundary.lean`) still use the raw `#guard_msgs`-pinned `#print axioms` idiom together with `assert_no_sorry`.

Two related pieces of work, deliberately deferred while the #24–#56 stack is in flight and best done after it lands:

1. **Migrate the pins to the macros.** `assert_axioms` states the exact expected axiom set in one line and is robust to `#print axioms` formatting drift; `assert_computable` additionally enforces that the computed-data producers stay computable (the raw idiom checks sorry-freedom and the axiom set, but nothing re-checks that a `def` has not silently become `noncomputable`). Whether to also fold the Snark pins into the top-level `Zcash/TrustBoundary.lean` census, as was done for the Ledger and BindingSignature pins, can be decided per module.

2. **Update `book/src/formal-verification.md` accordingly.** The trust-boundary section still describes the raw `#guard_msgs`-pinned idiom as the mechanism and cites `Zcash.Security.Ledger.TrustBoundary` and `Zcash.Security.BindingSignature.TrustBoundary` as separate modules; those were consolidated into `Zcash/TrustBoundary.lean` by #50. The historical `Lean.ofReduceBool` contrast sentence is accurate and can stay. (Check the `Zcash.Snark.Fingerprint.TrustBoundary` module name while there.) Note #56 edits adjacent text in the same file, so this should be sequenced after the stack merges.

🤖 Claude Fable 5

PR #77, opened 2026-07-21
chore(docs): add a security definitions map to the book

TalDerei:
Closes #55.

The existing "Proof Map" book entry covers verifier soundness (`Zcash/Snark/`). This adds its companion for the protocol security properties under `Zcash/Security/` — binding-signature *balance*, *key binding*, and the *ledger-model games* — showing how each reduces, as computed data, to a primitive hardness assumption and how it connects to the verifier-soundness spine.

PR #79, opened 2026-07-21
chore(fv): consolidate the trust boundary onto a single file and the `AxiomCheck` macros

### LINKED ISSUE #49: Single top-level `TrustBoundary` file

TalDerei:
Closes #49 and #71.

Consolidate the per-directory trust-boundary files into a single top-level census, then migrate the checks onto the `Zcash.Meta.AxiomCheck` macros. 

It additionally enforces things about transitive dep chains that `--wfail` cannot:

  - No unexpected axioms or unauthorized native_decide.
  - Computed reductions remain actual computable defs, not noncomputable.
  - Classical.choice appears only where explicitly permitted.
  - Transitive problems in cached/imported .olean dependencies are still detected.

---------------------------

## `ea30657` — #49: single top-level `TrustBoundary`
- Folds the three `Zcash`-library Snark `TrustBoundary` files (`Soundness`, `Soundness/Deployed`, `Soundness/AGM`) into the single top-level `Zcash/TrustBoundary.lean`, segmented into sections; drops their imports from `Zcash/Snark.lean`; deletes the old files.
- `Zcash/Snark/Fixtures/SingleAction/TrustBoundary.lean` stays put — it belongs to the separate `FixtureCheck` lake target that `lake build Zcash` omits by design (the captures are large and slow), so absorbing it would pull the fixtures into every default build.

## `cc241da` — #71: migrate onto the AxiomCheck macros
- Replaces the raw `assert_no_sorry` + `#guard_msgs`-pinned `#print axioms` idiom with `assert_axioms` / `assert_computable`:
  - Computed break reductions → `assert_computable … +choice` (`+native` for the Vesta producers), additionally enforcing they stay plain computable `def`s.
  - Soundness theorems / probability bounds / run-time lemmas → `assert_axioms` (`+native` on the Vesta-instantiated endpoints).
- The `SingleAction` fixture check moves its `assert_no_sorry` calls onto `assert_axioms`, keeping the `#guard_msgs`-pinned `#print axioms` for the `native_decide` fixtures — the one target where documenting the exact `native_decide` axiom is itself the claim.

## `3fb9196` — #71 docs
- Rewrites the trust-discipline section of `book/src/formal-verification.md` to describe the single consolidated census and the `assert_axioms` / `assert_computable` mechanism, replacing the now-removed per-directory module references.


daira (commented):
Reviewed at branch head `3fb9196`, at the *mechanism* level. Line-by-line review deferred, since main has moved under the branch (#77, #80, #84) and the rebase will reshape the diff. CI was fully green at the last push; the current conflicts are against the merges since.

## Verdict

The consolidation is faithful and the migration matches the discipline #71 asked for. All 191 pinned declarations from the three folded Snark census files — the 181 `assert_no_sorry`/same-line `#print axioms` targets plus the 10 wrapped `#print axioms` targets — reappear in the consolidated census, alongside the pre-existing top-level entries; none are dropped. Tier assignments follow the stated rule: computed break reductions at `assert_computable +choice` (with `+native` only on the Vesta-instantiated producers), theorems and probability bounds at `assert_axioms` (`+native` on the Vesta endpoints). Keeping `Fixtures/SingleAction/TrustBoundary.lean` in the separate `FixtureCheck` target is right (absorbing it would pull the captures into every default build), as is retaining the `#guard_msgs`-pinned `#print axioms` there — the one place where the exact `native_decide` axiom identity is itself the claim. `Zcash/Meta/AxiomCheck.lean` is unchanged by this PR, so no CompElliptic-side sync is triggered.

The asks below are mostly rebase-folds; one migration gap and one design question.

## Rebase-folds (main moved under this PR)

1. **#84's census tier upgrade.** The four reach-set/split-family producers (`ComputedAlgebraicFSFamilyUnbounded{,Rand}.globalReachSet` / `.splitFamilyRand`) became plain defs in #84, and the AGM census now holds them at `assert_computable +choice`. The consolidated census has them at `assert_axioms` — correct at this branch's base, where they were still `noncomputable`. The conflict resolution on the AGM file needs to carry the upgraded tier into the consolidated census, or the computability guarantee #84 installed is silently lost.
2. **The book's scoped-exception paragraph.** The rewritten trust-discipline section still contains "One scoped exception: the arbitrary-domain adversary reduction (`globalReachSet`, `splitFamilyRand`) is `noncomputable` ..." — #84 removed that paragraph from main (the exception no longer exists). Drop it during the rebase.
3. **#80's new modules.** `Forking/KnowledgeError.lean` and `Deployed/ConcreteBounds.lean` landed with no census coverage (flagged in the #80 review); this PR is the natural place for their entries.
4. **#73 contention (if it merges first).** It appends a whole-table key-binding section to the same top-level census; the conflict is append-vs-append, textual only.

## Migration gap

`Zcash/Snark/Fixtures/MultiAction/TrustBoundary.lean` (present at this branch's base) is untouched: still `assert_no_sorry` plus unpinned `#print axioms`, while its `SingleAction` sibling was migrated to `assert_axioms` plus pinned `#print axioms`. Either migrate it the same way or note why it is out of scope — the PR description names only `SingleAction`.

## Design question: origin of native trust

The old `#guard_msgs`-pinned idiom documented exactly *which* `native_decide` witnesses a Vesta endpoint trusts. `assert_axioms +native` admits any number of native axioms from any origin, so a new `native_decide` introduced deep in a dependency would pass the census unnoticed. The advantages of the macro —much more concise syntax, and tolerance of changes to the native axiom names chosen by the toolchain— are significant. Currently exact pins survive where they matter most (the fixtures). An alternative that keeps the best of both worlds: alongside bare `+native` (all native axioms allowed), accept an origin-restricted form `+native(decl)` that allows only the native axioms for `decl` — matching axiom names of the form `decl.*.native_decide.*`, where `*` is any number of name components. That pins where compiler trust enters while staying stable across the toolchain-dependent tail, at the cost of some macro complexity. Fine as a follow-up issue rather than in this PR; it would also need mirroring in CompElliptic's `AxiomCheck.lean` when done.

TalDerei:
@daira the migration gaps it notes in the fixtures were explicit design decisions, which i didn't feel strongly about. now that i think about it, not sure why keeping the trust boundaries separate for the fixtures seemed like a good idea so will consolidate this.

Zcash/TrustBoundary.lean
```diff
@@ -133,6 +133,16 @@ open Zcash.Snark
...
 
+/-! ### Verifier-soundness capstones -/
+
+assert_axioms deployedAccepts_verifierEq
+assert_axioms orchard_verifier_sound_conditional
+assert_axioms orchard_verifier_deployed_opening_of_forked
+assert_axioms orchard_verifier_deployed_constraint_of_forked
+assert_axioms orchard_verifier_sound_vesta_conditional +native
+assert_axioms orchard_verifier_vesta_opening_of_forked +native
+assert_axioms orchard_verifier_vesta_constraint_of_forked +native
```
TalDerei, line 144: we missed these capstone pins even prior to this PR.

daira (approved):
Re-reviewed at the rebase. All the rebase-fold asks from the mechanism review are addressed: the four reach-set/split-family producers carry `assert_computable … +choice` in the consolidated census, the book's scoped-exception paragraph is gone, the fork-tree knowledge-error modules are covered, the MultiAction fixture check is migrated onto the macros, and the whole-table key-binding section from #73 is preserved. The warning-strict CI with the tree-wide lint cleanup is a welcome extra.

On folding the fixture censuses in: the reason to keep them separate stands — they belong to the `FixtureCheck` target that `lake build Zcash` deliberately omits, so absorbing them into the top-level census would pull the captures into every default build.

The `+native(decl)` origin-pin remains a follow-up (it needs the CompElliptic `AxiomCheck.lean` mirror when done).

daira:
One more rebase-fold before merge: #90 landed after this branch's rebase, and its seven census entries are absent from the consolidated file — `Extractor.card_ivk_ge`, `uniformOfFintype_prod`, `evalEquiv` (flagless `assert_computable`), `uniform_triple_eval`, `break_measure_le_product`, `toInterface_break_measure_le` (`assert_axioms`), and `nfOldEqOrBreak` (`assert_computable +choice`). Tiers as on main; without the fold the merge drops them silently.

PR #81, opened 2026-07-21
chore(docs): document the Lean module layout

### LINKED ISSUE #64: Create documentation index of Lean modules
It'd be nice of we can have something in the docs that describes what each subfolders does, for instance: "Soundness/Forking/ is the reusable forking kernel: random-oracle primitives, fork-tree/certificate definitions, probability lemmas, transcript ordering/assembly, and the certificate-to-opening-or-relation machinery."

TalDerei:
Closes https://github.com/zcash/ironwood/issues/64, and should merge after #96.

-------

**AI Summary:**

Adds a **Source Map** page to the book's Formal Verification section — a directory-by-directory index of the Lean development under `Zcash/`, drawn from each module's own docstring, alongside the **Proof Map**, **Security Definitions**, and **Glossary**. Rebased onto current `main` and refreshed for the 116 commits of drift since it was drafted. Three of the draft's claims were wrong rather than merely incomplete:

- **Consolidated trust boundaries.** The draft described `TrustBoundary` files under `Soundness/`, `Soundness/AGM/`, and `Soundness/Deployed/`. All three were deleted by the SNARK trust-boundary consolidation; the census is now `Zcash/TrustBoundary.lean` plus the two per-capture fixture boundaries.
- **Misattributed fixture checks.** It credited `SingleAction/` with the faithfulness and negative fixtures, which live only in `MultiAction/`, and scoped the checked `TrustBoundary` to the single-action capture, though both captures carry one.
- **Wrong target scope.** It attached the `FixtureCheck` lake target to one subtree, where the glob is all of `Zcash.Snark.Fixtures.+`.

Also: the framing moves from two layers to three, since `Zcash/Circuits/` (13 directories) was absent entirely; `Soundness/Composition/` and `Soundness/Multiopen/` are added; the `Security/Ledger/` entry now covers Balance, Spendability, and Spend Authority as modules rather than an interface the page pointed at; and ~14 new `Soundness/` modules are placed.

## Review notes

- The page stays at the directory level by design, naming notable modules as entry points; the per-module detail stays in the docstrings, which is what the descriptions are drawn from.
- Every module, type, and directory name the page cites resolves against the tree, and all 34 directories under `Zcash/` are covered.
- The rebase drops the old `Merge branch 'main'` commit, so this is a single commit. The one conflict was `SUMMARY.md`, resolved to keep `main`'s Security Definitions entry.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -12,5 +12,6 @@
 - [Formal Verification](formal-verification.md)
   - [Proof Journey](formal-verification/proof-journey.md)
   - [Proof Map](formal-verification/proof-map.md)
+  - [Source Map](formal-verification/source-map.md)
   - [Security Definitions](formal-verification/security-definitions.md)
   - [Glossary](formal-verification/glossary.md)
--- a/book/src/formal-verification/source-map.md
+++ b/book/src/formal-verification/source-map.md
@@ -1,0 +1,207 @@
+# Source Map
+
+A directory-by-directory index of the Lean development under [`Zcash/`](https://github.com/zcash/ironwood/tree/main/Zcash):
+what each subtree contains and where to start reading. It is the source-tree companion
+to the [proof map](proof-map.md) (which traces how the results connect), the
+[security definitions](security-definitions.md) (which state the properties being proven),
+and the [glossary](glossary.md) (which defines the coined terms).
+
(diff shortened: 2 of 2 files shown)


TalDerei:
@daira would you prefer this be a readme under `Zcash/`?

daira (approved):
ACK

daira:
> @daira would you prefer this be a readme under `Zcash/`?

No it's fine here.

PR #82, opened 2026-07-21
chore(fv): proof map fixes

TalDerei:
The proof-structure map crowned the legacy propositional family (legacy_orchard_verifier_vesta_*) as the top-level 'verifier soundness' box. Retitle that box to the current computed capstone (knowledgeSoundness_under_DL / binding_under_DL) and rewrite its description to match; the corner 'DL reduction bound' box stays the proven sub-bound it packages.


daira (approved):
ACK (doc-only) modulo merge conflicts.

daira (approved):
ACK

PR #84, opened 2026-07-21
fv: computable Fintype for Vesta; drop the noncomputable exception

### LINKED ISSUE #83: Make use of the computable `Fintype` for Vesta once CompElliptic#10 lands
https://zcash.github.io/ironwood/formal-verification.html says:

> A structure with data fields cannot be inhabited by proof-irrelevant existence, and a plain `def` cannot conjure the data from mere existence via choice — the compiler enforces this, so `noncomputable` is not permitted for these definitions. One scoped exception: the arbitrary-domain adversary reduction (`globalReachSet`, `splitFamilyRand`) is noncomputable, solely because it materializes a `Fintype` instance for the Vesta curve group (`Fintype.ofFinite`); the witness data those endpoints produce still flows through the same computable relation kernels.

Once daira/CompElliptic#10 is fixed: bump the CompElliptic dependency, import the `Fintype` instance from there, mark these definitions as computable, update the corresponding `TrustBoundary` checks, and remove the above exception from the book.

daira:
Fixes #83, consuming daira/CompElliptic#11.

The CompElliptic side derives `Fintype` for curve points computably (a filter of `F × F` under the decidable `Valid` predicate). This PR bumps the dependency to CompElliptic main (`a549e4555b17bbbe3c8d48aa7d5b493912ccfbe9`, the daira/CompElliptic#11 merge), removes the nine `Fintype.ofFinite` materializations, and makes `globalReachSet` / `splitFamilyRand` (both the unbounded and randomized variants) plain `def`s. The trust-boundary guards pass unchanged — the pinned axiom sets were already accounting for `Classical.choice` from erased positions. The reach-set and split-family pins are upgraded from `assert_no_sorry` to `assert_computable +choice`. The book's scoped-exception paragraph is removed.

Verification: full build green (8580 jobs, rebased over the #77/#80 merges) with the guards intact; the defs compile without `noncomputable`, so the compiler certifies the data path.

🤖 Claude Fable 5


daira (commented):
Self-ACK

TalDerei (approved):
ACK

Issue #65, opened 2026-07-19
Check instance-commitment derivation in Lean


TalDerei:
The fingerprint fixture currently imports Rust-computed instance commitments as points. Therefore, `fingerprint_matches` checks MSM assembly given those points, but not:

```text
public inputs → instance commitments
```

Derive the commitments from public inputs in Lean and check them against the captured Rust commitments for both the single- and multi-action fixtures.

TalDerei:
Currently, Ironwood's `VerifyingKey` is effectively "circuit VK plus an opaque statement commitment."

| | Current model | Desired model |
|---|---|---|
| Circuit-fixed data | Domain constants, gates, query layouts, fixed commitments, permutation commitments/chunks, and lookup expressions | Remains in `VerifyingKey` |
| Public instances | Not present in Lean | Supplied separately per proof/action |
| Instance commitment | Caller-supplied `vk.instanceCommitment : ... → G` | Computed from the supplied public instances |
| Verification | Checks openings against the opaque commitment | Checks openings against the derived commitment |

Currently, `vk.instanceCommitment p` is taken directly from `capturedInstanceCommitments` and used by [`assembleQueries`](https://github.com/zcash/ironwood/blob/9911199dae1049984112840fc809e7842672e511/Zcash/Snark/Verifier/Assemble.lean#L129-L142). Lean never sees the underlying anchor, nullifier, `rk`, and other public inputs. It therefore proves that the proof's `instanceEvals` open consistently against the supplied point, but not that the point represents the actual public instance.

Conceptually, the desired interface is:

```text
verify :
  CircuitVerifyingKey →
  PublicInstances →
  ProofString →
  ...

let instanceCommitment :=
  commitLagrange urs (zeroPad publicInstances)
```

The derived commitment must then be used in both:

- the Fiat-Shamir transcript;
- the instance opening queries.

It could remain cached inside a larger `VerifierContext`, but it should not be an unconstrained field of the circuit VK. At minimum, the current field needs an accompanying proof:

```text
instanceCommitment =
  commitPublicInstances urs publicInstances
```

The intended separation is: the VK determines the circuit; the public instances determine $C_{\mathrm{inst}}$; verification combines them.

TalDerei:
superseded by tracking in https://github.com/zcash/ironwood/issues/86.

PR #85, opened 2026-07-22
feat(fv): derive instance commitments

TalDerei:
Consumes upstream https://github.com/zcash/halo2/pull/918 and https://github.com/zcash/orchard/pull/538. Step towards closing the public instance gap referenced in https://github.com/zcash/ironwood/issues/65 and tracked in https://github.com/zcash/ironwood/issues/86.

Invokes lean equivalence checks between public inputs and derived instance commitments (see generated fixture changes), and launders the instance commitments through dozens of callsites.


.github/workflows/fixtures.yml
```diff
@@ -37,9 +37,11 @@ concurrency:
...
 
 env:
-  # Orchard release tag shipping zcash/orchard#531 ("Add Post-NU6.3 verifier fingerprint fixtures").
-  ORCHARD_REPO: zcash/orchard
-  ORCHARD_REF: 0.15.1
+  # Orchard PR zcash/orchard#538 ("Derive instance commitments from public inputs in the fingerprint
+  # fixture export"), which threads the public instances into the capture. Pinned to a commit SHA;
+  # Halo2 is supplied by that branch's `[patch.crates-io]`, which points at zcash/halo2#918.
+  ORCHARD_REPO: TalDerei/orchard
+  ORCHARD_REF: 65ee845f7933010149b52e447840bfa92548066e
```
TalDerei: yml workflow will need to be updated with release tag.

TalDerei: https://github.com/zcash/ironwood/pull/85/changes/fc2fdd71cf8b4cd9da79c46fe5b06829bbeeb880

daira (commented):
Reviewed at branch head `fc2fdd7` (6 commits over merge base `4dbd0ac`, 18 files, +594/−491), CI fully green including the fixture regeneration. Upstream anchors verified: zcash/halo2#918 and zcash/orchard#538 both merged 2026-07-22, consumed here as halo2_proofs 0.3.4 and the `ORCHARD_REF 0.15.3` capture pin.

## Verdict

No showstoppers; this is a clean narrowing of the public-instance gap (#65), and the trust movement is exactly right. What was previously an opaque trusted point — the VK's `instanceCommitment` field, set in the fixtures straight from the capture — is now (a) removed from the VK (which correctly holds circuit-fixed data only), (b) threaded as an explicit, universally-quantified argument through the verifier and every soundness capstone, and (c) *proven* in the fixtures to equal the commitment derived from the captured public inputs. The soundness theorems become strictly more general — every `−` line in the Main/Vesta diff is an old-arity restatement, with no hypothesis or check removed — and the derivation matches halo2: `commitLagrange` is the MSM `Σᵢ instanceᵢ • gLagrangeᵢ + w`, i.e. `Params::commit_lagrange` with `Blind::default() = 1`, over the URS Lagrange bases, with `capturedPublicInstances_within_lagrange` guarding the `getD` prefix so a missing generator cannot silently contribute the identity.

The core new fact is `instance_commitments_derived` (per fixture): the captured instance commitments equal the Lean-computed commitments of the captured public inputs. Its `native_decide` tier is appropriate — it is an instance-level capture check, the same tier as the pre-existing `fingerprint_matches`.

## Asks

1. **Fixture census pins for the new trust surface.** The per-fixture `TrustBoundary.lean` files still pin only `fingerprint_matches` / `capturedMsm_eval_eq_zero`. The two new `native_decide` theorems per fixture (`instance_commitments_derived`, `capturedPublicInstances_within_lagrange`) are precisely the new trust-bearing facts and should get the same pinned-`#print axioms` treatment; the new defs (`commitLagrange`, `derivedInstanceCommitment`, `capturedPublicInstances`, `capturedUrsGLagrange`) want `assert_axioms` pins alongside. (Post-#79, the fixture censuses deliberately stay separate from the top-level file, which is unaffected.)

## Notes (no action needed)

- The soundness layer deliberately does not pin `instanceCommitment` to the derivation — the pinning lives in the fixtures, and the parametric wiring is #86's scope. The quantifier direction makes this sound: the capstones hold for *every* commitment function, so instantiating at the derived one is free.
- MaxShape's commId-disjointness change is pure arity threading; the instance-column slot keeps its `CommitmentId.instanceCol p` tag, so cross-sub-proof disjointness is unchanged.
- Rebase note: main has moved (#79 consolidation, #90, #89); the contention is limited to the deleted per-directory Snark census files and `Zcash/Snark.lean` imports — the fixture files this PR touches are outside #79's consolidation.

daira (approved):
The census ask is addressed: `instance_commitments_derived` and `capturedPublicInstances_within_lagrange` are pinned alongside the new defs, with the docstring stating exactly what trust the fixture buys (the derivation in place of opaque commitments). One note, no action needed: these use `assert_axioms +native` (an upper bound) where the fixture censuses use exact `#print` pins for `fingerprint_matches`; the `+native(decl)` origin-pin follow-up from the #79 review would be the natural tightening for both. Approve.

PR #87, opened 2026-07-22
Remove unneeded `classical` and `open Classical`

daira:
Removes `classical` tactics and `open Classical in` modifiers that rederive instances the sections already provide, across the Forking/AGM stack (per the instance-hygiene convention: assume the section's instances; rederiving them just lengthens proofs).

Full build green under the warning-strict linting from #79.

🤖 Claude Fable 5


daira (commented):
Self-ACK

TalDerei (approved):
nice ACK

PR #90, opened 2026-07-23
fv: complete the key-binding ROM model

### LINKED ISSUE #68: Model ROM semantics for the key-binding theorem's probabilistic layer

daira:
Fixes #68 — the four remaining items of the key-binding ROM model.

* **Triple↔table transport, as a stated PMF equality.** `uniformOfFintype_prod`, `evalEquiv`, and `uniform_triple_eval`: independent uniform draws of the three rivk-derivation oracles, combined with `FinalQuery.eval`, are exactly the uniform whole-table draw.

* **Five independent oracle tables.** `break_measure_le_product` states ZIP 2005's probability space in full — adversary private randomness under an arbitrary distribution, and the five oracle tables drawn independently, the tables uniformly — keeping the hypothesis-free `(n+4)·(n+3)/|F|` bound.

* **Interface delivery.** `toInterface_break_measure_le` phrases the break event through the games' `KeyBindingInterface`, so the games' `∨ kv.Break` branches inherit the bound directly.

* **Role-typed variables.** The shared base-field variable splits by role, following the spec typing of each quantity: the key-binding layer gets `IVK`/`AK`/`NK`, and the games layer's `Primitives` / `Note` / `ActionInstance` / `ActionWitness` split over `IVK`, `NK`, `RHO`, `PSI`, `CMX`, and `RT` (nullifiers share `RHO`, forced by ρ-uniqueness). All concretely `ZMod q` for Orchard; the roles are documentary and Sapling-ready.

* **Scalar role types.** `RIVK` (rivk values and the rivk-derivation oracle outputs; `G` is an `RIVK`-module and the capstone denominators read `|RIVK| = r`) splits from `ASK` (`H^ask` outputs, acting on `G` via `SMul` only); the generic birthday layer keeps its own field `F`, instantiated at `RIVK` by the capstones. Type-parameter order is canonical across the layer — `G, IVK, AK, NK, RIVK, ASK, QK, SK`, with `QK` before `SK` matching the ZIP's tagged-union order — and the `Witness`/`BreakProj` fields and the instance arguments follow it; ZIP 2005's witness and break-projection presentations are to adopt the same order.

* **`IVK` sizing, checked.** `Extractor.card_ivk_ge`: `toIVK_pm` makes `toIVK` at most 2-to-1, so `|G| ≤ 2·|IVK|` — an `IVK` small enough to find collisions in admits no `Extractor`, making undersized instantiations vacuous rather than falsely secure. This records why the birthday bounds involve only `|RIVK|`; a hash-derived `ivk` (Sapling's `CRH^ivk`) does not satisfy `toIVK_pm` and would need a collision-resistance term in the bound instead.

Two signature refactors ride along: the five oracles bundle as an `Oracles` structure, and the two extract maps as an `Extractor` structure (`toIVK`, carrying the ±-property as the `toIVK_pm` field, and `toAK`; both instantiate to `Extract_P` concretely). A wording clarification in the `badList` docstring leads the branch.

Verification: full build green at the tip, and every commit in the range builds green individually; census entries added for all new declarations, with `evalEquiv` at the flagless `assert_computable` tier.

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Security/KeyBinding/Probability.lean
+++ b/Zcash/Security/KeyBinding/Probability.lean
@@ -13,7 +13,7 @@ This module adds the probabilistic model that turns those counting facts into a 
 bound.
 
 The model samples the combined final `rivk`-derivation oracle `H^*` as one uniform table
-`O : FinalQuery QK SK B F → F`. A uniform table is the same as independent uniform outputs at
+`O : FinalQuery QK SK AK NK F → F`. A uniform table is the same as independent uniform outputs at
 every query — the product structure of the uniform measure. (In Markov-category terms `PMF` is a
 commutative monad, so samples reorder freely; see Fritz [eprint arXiv:1908.07021] and
 `Mathlib.CategoryTheory.MarkovCategory` for the abstract setting.) The three constituent oracles
@@ -59,12 +59,12 @@ namespace Zcash.Security.KeyBinding
 open Finset Zcash.Security.RandomOracle Zcash.Security.Birthday Zcash.Snark
 open scoped ENNReal
 
-variable {G F B SK QK : Type*}
+variable {G F AK NK SK QK : Type*}
 
 /-! ## Finiteness of the query space -/
 
 /-- `FinalQuery` as the sum of its constructors' data. -/
-def finalQueryEquiv : FinalQuery QK SK B F ≃ (QK × B × B) ⊕ SK ⊕ (F × B × B) where
+def finalQueryEquiv : FinalQuery QK SK AK NK F ≃ (QK × AK × NK) ⊕ SK ⊕ (F × AK × NK) where
   toFun
     | .ext qk ak nk => .inl (qk, ak, nk)
     | .legacy sk => .inr (.inl sk)
@@ -76,13 +76,13 @@ def finalQueryEquiv : FinalQuery QK SK B F ≃ (QK × B × B) ⊕ SK ⊕ (F × B
   left_inv q := by cases q <;> rfl
   right_inv x := by rcases x with ⟨qk, ak, nk⟩ | sk | ⟨rivk_ext, ak, nk⟩ <;> rfl
 
-instance [Fintype QK] [Fintype SK] [Fintype B] [Fintype F] :
-    Fintype (FinalQuery QK SK B F) :=
+instance [Fintype QK] [Fintype SK] [Fintype AK] [Fintype NK] [Fintype F] :
+    Fintype (FinalQuery QK SK AK NK F) :=
   Fintype.ofEquiv _ finalQueryEquiv.symm
 
 /-- Assembling the combined final oracle from a table's constructor restrictions gives back the
 table. -/
-theorem eval_restrict (O : FinalQuery QK SK B F → F) :
+theorem eval_restrict (O : FinalQuery QK SK AK NK F → F) :
     FinalQuery.eval (fun sk => O (.legacy sk)) (fun qk ak nk => O (.ext qk ak nk))
       (fun rivk_ext ak nk => O (.int rivk_ext ak nk)) = O := by
   funext q; cases q <;> rfl
@@ -502,20 +502,20 @@ theorem toOuterMeasure_bind_le {α β : Type*} (p : PMF α) (f : α → PMF β) 
 
 section Capstone
 
-variable [AddCommGroup G] [Field F] [Field B] [Module F G] [NoZeroSMulDivisors F G]
-variable [Fintype QK] [Fintype SK] [Fintype B] [Fintype F]
-variable [DecidableEq QK] [DecidableEq SK] [DecidableEq B] [DecidableEq F]
+variable [AddCommGroup G] [Field F] [Field AK] [Module F G] [NoZeroSMulDivisors F G]
+variable [Fintype QK] [Fintype SK] [Fintype AK] [Fintype NK] [Fintype F]
+variable [DecidableEq QK] [DecidableEq SK] [DecidableEq AK] [DecidableEq NK] [DecidableEq F]
 
-omit [Fintype QK] [Fintype SK] [Fintype B] [Fintype F] in
+omit [Fintype QK] [Fintype SK] [Fintype AK] [Fintype NK] [Fintype F] in
 /-- The queries at which `CollisionUpToSign.ofBreak` exhibits its collision are the witnesses'
 derivation queries: the `rivk_ext`-derivation pair in the residual case, the final-query pair
 otherwise. Stated over any `c` equal to the computed collision (`hc`), so that a consumer's
 `set c := CollisionUpToSign.ofBreak ... with hc` supplies `hc` directly. -/
-theorem ofBreak_queries {Extract : G → B} {S : G} {hfn : B → B → F} {Ggen : G}
+theorem ofBreak_queries {Extract : G → AK} {S : G} {hfn : AK → NK → F} {Ggen : G}
     {hExt : ∀ P R : G, Extract P = Extract R ↔ P =± R} {hS : S ≠ 0}
-    {Hask : SK → F} {Hnk : SK → B} {Hrivk_legacy : SK → F}
-    {Hrivk_ext : QK → B → B → F} {Hrivk_int : F → B → B → F}
-    {w₁ w₂ : Witness G F B SK QK}
+    {Hask : SK → F} {Hnk : SK → NK} {Hrivk_legacy : SK → F}
+    {Hrivk_ext : QK → AK → NK → F} {Hrivk_int : F → AK → NK → F}
+    {w₁ w₂ : Witness G F AK NK SK QK}
     {hbrk : Break Extract S hfn Ggen Hask Hnk Hrivk_legacy Hrivk_ext Hrivk_int w₁ w₂}
     {c : RandomOracle.CollisionUpToSign
       (shiftedFinalOracle Extract Ggen hfn Hask Hnk Hrivk_legacy Hrivk_ext Hrivk_int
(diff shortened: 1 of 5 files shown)


daira (commented):
Self-ACK

TalDerei:
Audit notes (statement-level review). Tip CI green, no `sorry`s, census complete; the deterministic layer and birthday machinery are faithful and correct. Findings, most severe first:

### 1. The five-table capstone's adversary is blind to `H^ask`/`H^nk` — Medium

In `break_measure_le_product` (`Probability.lean:718`) and `toInterface_break_measure_le` (`Instance.lean:43`), `i` is drawn independently of the tables and the machine's only oracle is the combined rivk table — so the adversary can neither query nor depend on `H^ask`/`H^nk`. ZIP 2005's `q_kb` counts queries to all five oracles, and sk-branch witnesses need `H^ask(sk)` to satisfy `akP = H.ask sk • Ggen`; that attack surface is outside the modeled class. (`break_measure_le_mixture` covers table-dependent adversaries via `ι`, but isn't what the interface delivers.) Cheap, proof-preserving fix: `A : ι → (SK → ASK) → (SK → NK) → OracleComp …` — `break_measure_le_of_queryBound` is already uniform in both tables.

### 2. The proven bound is tighter than ZIP 2005's, but docstrings claim identity — Low

ZIP: ε_kb ≤ 3·q(q−1)/(2·r). Here: q(q−1)/r, by relocating the residual case to the `rivk_ext`-derivation pair the union bound already counts. Sound (the formal bound implies the ZIP's), but "This is ZIP 2005's `ε_kb`" (`Basic.lean:24`, `Probability.lean:37/49`) isn't literally true until zcash/zips#1338 — which adopts exactly this argument — merges.

### 3. `Extractor.toAK` carries no axiom — Low

The split kept the ± property only on `toIVK`. All proofs go through, and a coarser `toAK` only shrinks `Break`, but `ak_pinned`'s "pinned up to y-sign" silently degrades to nothing under a coarser instantiation (a constant `toAK` type-checks). Require the ± property for `toAK` too, or document the caveat on `ak_pinned`.

### 4. Instantiation obligations to record — Info

- `toIVK_pm` at the identity needs "no Pallas point has x = 0" (x = 0 forces y² = 5, a non-square) — the one nontrivial obligation when the CompElliptic instantiation lands.
- `KBOpening.nonzero` is faithful (ZIP requires `ivk ∉ {0, ⊥}`) but unused by any proof; its discharge falls to the future statement instantiation.

### 5. Nits

`break_measure_le_product`'s docstring says `/|F|` where the convention is now `|RIVK|`; `uniformOfFintype_prod` is a general PMF lemma whose natural home is `Forking/Probability.lean` (or Mathlib).

---

**Verified clean:** `card_ivk_ge`, `commit_scalar_pm`, the `finalQueryOf` fixpoint decode and all four cases of `residual_of_finalQuery_eq`, `CollisionUpToSign.ofBreak`, `evalEquiv`/`uniform_triple_eval`, the `+4` completion arithmetic, the `Oracles`/`Extractor` bundling and reordering (no role-type swaps; oracle inputs, break projection, and `ρ`/nullifier `RHO`-sharing match the ZIP verbatim), the `Statement.lean` role-typing, the definitional `Break` hand-off in `toInterface`, and census tiers. The per-commit-green claim was not re-verified (tip only).

🤖 With the help of Claude Fable 5

Zcash/Security/KeyBinding/Probability.lean
```diff
@@ -637,22 +692,62 @@ pointwise capstone conditions on the adversary's private randomness and the `H^a
...
-  exact break_measure_le_of_queryBound Extract S hfn Ggen hExt hS (Hask i) (Hnk i) (hQ i)
+  exact break_measure_le_of_queryBound Extract S hfn Ggen hS (Hask i) (Hnk i) (hQ i)
+
+
+/-- **Adaptive key-binding bound over five independent oracle tables** — ZIP 2005's
+probability space in full: the adversary's private randomness (any distribution `p`) and
+the five oracle tables — `H^ask`, `H^nk`, and the three final `rivk`-derivation oracles —
+drawn independently, the tables uniformly. The machine runs on the combined table, and the
+bound is the hypothesis-free `(n+4)·(n+3)/|F|`. -/
+theorem break_measure_le_product {ι : Type*} [Fintype ASK] [Nonempty NK] [Nonempty ASK]
```
TalDerei, line 725: it says this samples `H^ask`, `H^nk`,  but `A` doesn't receive them. we need to parameterize `A` by `H^ask` and `H^nk` per the ZIP defintion, otherwise this isn't a full five-oracle adversary from what i can tell. 

and just conjecturing but `toInterface_break_measure_le` also needs to propagate this because it also mentions the five oracle tables in its doc comments?

TalDerei (approved):
Optimistic ACK after https://github.com/zcash/zips/pull/1338 lands (so the stronger claims we make here match the upstream ZIP) and `break_measure_le_product` is fixed.

Otherwise, i understand this PR to prove that with five independent hash oracles, an attackers chance of breaking key binding stays below some birthday style bound, and then we expose that bound to high level security games and also clean up some types in the process.

daira:
3 is not a valid finding. We don't state a similar `toAK_pm` because it doesn't matter for what we are trying to prove at the layer of ZIP 2005's security arguments. `Extract.toAK` and `Extract.toIVK` both have to correspond to the actual $\mathsf{Extract}_{\mathbb{P}}$ in the protocol anyway. When we prove *concrete* bounds on the probability of breaking Spendability or Spend authority, we'll need the cardinality of `AK`, but that is not this layer.

Zcash/Security/KeyBinding/Probability.lean
daira: Confirmed — the family was indexed by the coin space alone, which is a weaker class than the ZIP's five-oracle adversary. Fixed in 5c76ebe: `A` is now parameterized by both sampled tables (receiving the whole table dominates any query strategy against `H^ask`/`H^nk`, and the query budget stays on the rivk oracles, which is conservative since only `H^*` queries can complete a collision). The mixture form already covered the general case through its index; the product form now matches it. And yes — `toInterface_break_measure_le` propagates the same parameterization.

daira:
All addressed:

1. Fixed in 5c76ebe — `A` is parameterized by both sampled tables (receiving the whole table dominates any query strategy against `H^ask`/`H^nk`, and the query budget stays on the rivk oracles, conservative since only `H^*` queries can complete a collision). `toInterface_break_measure_le` propagates the same parameterization.
2. Fixed in 1242469 — the `ε_kb` claims now say "as sharpened in zcash/zips#1338", which adopts exactly this argument.
3. As agreed, not a finding: the pinning guarantee is inherently "up to the extractor's fibres", and the consuming game demands the fidelity.
4. Recorded in 1242469 as docstrings on the declarations the instantiation will meet: `toIVK_pm`'s x = 0 obligation, and `KBOpening.nonzero`'s discharge belonging to the statement instantiation.
5. The `/|F|` docstring is fixed (1242469); relocating `uniformOfFintype_prod` to `Forking/Probability.lean` (or Mathlib) is recorded as a follow-up alongside the existing generic-lemma relocation note.

daira (commented):
Self-ACK

TalDerei (approved):
ACK

Issue #67, opened 2026-07-19
Compose forking extraction with the decoded capstones: retire ExtractableFromAcceptance


TalDerei:
Follow-up to #11, sitting at the apex of the #56 / #30 stack.

#11's random-oracle / forking side is proven (#24 → #28 → #56): the computed extraction pipeline (`recursiveAlgebraicFork` → `runToSnark`) yields, on an accepting run, a clean opening or a computed DL relation, with quantitative failure bounds across all adversary models (`snarkFailure_prob_le_of_*`, including privately-randomized, arbitrary-domain, and generator-RO). #30 (closing #18) binds the extracted IPA witness to the verifier's real columns: the decoded capstones consume columns recovered from the batched multiopen rather than a free decode.

The two results are stacked but do not yet touch: #56 ends at "accepting ⟹ clean opening"; #30 ends at "opening decodes to columns satisfying the gates". The deployed capstones (`Soundness/Main.lean`, `Soundness/Vesta.lean`) still consume the assumed bridge `ExtractableFromAcceptance`.

```text
        #56 (fs-adversary)              #30 (decode-witness)
   accepting ⟹ clean opening    opening ⟹ decoded columns ⊨ gates
                     \                    /
                      [this issue: composition]
        accepting ⟹ SnarkRelation, quantitatively bounded
```

**Scope**
1. Composition theorem: run `runToSnark`'s clean-opening branch through the decoded capstones (`Multiopen/Deployed`) to conclude the deployed `SnarkRelation` for the extracted witness.
2. Re-instantiate the `Main.lean` / `Vesta.lean` soundness theorems on the computed path, inheriting the bound (recursive query loss + `z = 0` slice + `|basis| · ε` DL advantage); demote `ExtractableFromAcceptance` and the current conditional family to legacy/derived status.
   Target `circuitSatViaConstraints` (#91), not `circuitSatViaGates`, so the re-instantiated capstones assert the whole constraint system rather than the gates alone. This is the same edit either way; doing it at the gate predicate means redoing it later.
3. Repoint the remaining deployed capstones to the same predicate — the `circuitSatViaGates` instantiations left in `VestaBudget.lean` and `Composition/Bridge.lean`. `snarkRelation_constraints` (#91) is the projection: it pairs the IPA opening those capstones already compute with the constraint identity, which `hfold_of_constraint_polys` plus `hgood` supplies at that point.
4. Pin the new endpoints in `TrustBoundary`.

[...]

PR #91, opened 2026-07-23
feat(fv): wiring the permutation and lookup arguments

### LINKED ISSUE #36: Fold into permutation + lookup arguments

TalDerei:
**Human Summary:**

Builds on #30, and closes [zcash/ironwood#36](https://github.com/zcash/ironwood/issues/36).

The soundness stack was written about gate checks only. This folds in the permutation and lookup arguments from #21 and carries them through the capstone chain (introducing full-constraint capstones). So the circuit-satisfaction relation (the predicate) now includes gates, permutation, and lookup.

----------

**AI Summary:**

This PR puts the permutation and lookup arguments into the constraint model, then proves both of them from checks the verifier already runs.

## What Is Proven

- **One set of builders:** the constraint builders never divide, so the same code produces the verifier values and the polynomials the proof is about.
- **Fingerprint:** the fold the verifier computes and the fold the proof uses are now one function run twice, so `hfp` is proved rather than assumed.
- **One check into many:** the verifier folds every constraint by `y` and checks a single value; a good `y` splits that back into one identity per constraint.
- **Rows:** read at a row, the permutation step rule *is* the running-product recurrence — both of its folds are running products.
- **Telescoping:** multiplying that recurrence across the rows, and across chunks through the chaining rule, leaves one product over every cell.
- **Multisets:** two root counts, one per challenge, carry that product to the multiset identity consumed by the #21 kernel — pairs for the permutation, and the two columns separately for the lookup.
- **Endpoints:** cells sharing a permutation cycle hold equal values; every lookup input value is a table value.

## Integration Status

The links are now chained through the deployed full constraint list. `eval_combineConstraints_deployed` identifies the verifier fold with the polynomial constraint fold; `hfold_of_constraint_polys`, `SnarkRelationWithMemberConstraints`, and the terminal wrappers carry that full numerator through the capstone chain. Circuit satisfaction is stated by `circuitSatViaConstraints`, covering gates, permutation, and lookup rather than gates alone.

The remaining top-level work belongs to #67 / PR #96: retain the per-proof decoder indexing, tie extraction to the actual batch and canonical member decodes, construct the adaptive probability coupling, consume the priced good-challenge event, and transfer the resulting composite bound to deployed acceptance.

Two hypotheses are conditions on the concrete key, written as ordinary binders rather than axioms: that the `δ`-powers below the column count lie in distinct cosets of `⟨ω⟩` (what key generation picks `δ` for), and that the Lagrange terms are nonzero at the rows used. Both are decidable for a given key. The case where the running product ends at zero is priced, not assumed away: every result ends in *either* the identity *or* an explicit vanishing factor.

## Issue Scope Check

#36 asked for #21 to be folded into a stack framed around gate checks. The three steps listed by #21 as still to come — grand-product telescoping, the Schwartz–Zippel bridge, and final lookup assembly — are proved here, and `perm_copy_constraints` and `run_structure` are reached from the deployed step rule. The expression-model and fingerprint leg is complete here; the final deployed-acceptance probability composition belongs to #67 / PR #96.


TalDerei:
@daira this is ready for review now.

TalDerei:
this needs rebasing over #30.

daira (approved):
Reviewed at branch head `bdc2af0`, CI green (the two skipped jobs are path-conditional). The merge base is the #30 head, now merged, so the diff is exactly the PR's own 17 files (+3775/−72). This closes #36 the way the issue asked: the #21 kernels are consumed, not restated, and the constraint model now covers what the verifier actually checks.

## Verdict

Approve, with one small doc fix and one forward-looking note below. The two places Daira-Emma was most worried about — the zero-knowledge row adjustments and the permutation grand product spanning column chunks — are both handled faithfully and, in the chunked case, elegantly.

What I verified, in the order the argument runs:

1. **One set of builders, and the fingerprint is a theorem.** The pre-existing verifier expression builders (already fingerprint-validated against Rust) are generalized from `Field` to `CommRing`, so the same code runs at claimed evaluations and at column polynomials; the `Expr.map` / ring-hom commutation lemmas then make `eval_combineConstraints_deployed` an equation between two runs of the same builder, with `allExpressions_eq` identifying the verifier's own list with the generic one. `hfp` stops being an assumption.
2. **The constraint shapes match halo2.** `permutationExpressions` carries the start rule `l₀·(1 − z)` on the first set only, the end rule `(z² − z)·l_last` on the last set only, the chaining rule `(z_{c+1} − lastEval_c)·l₀` between consecutive sets, and the step rule switched off by `1 − (l_last + l_blind)`; `lookupExpressions_eq` writes out the lookup's five rules including the `ω⁻¹`-rotated previous-row read. The ZK adjustments are exactly right: every row-level conclusion is gated on the active-row factor being nonzero (`hactive`, a decidable condition on the concrete key), blinding rows are excluded throughout, and the last-row `z ∈ {0, 1}` relaxation is handled by `prod_eq_or_factor_eq_zero`, which surfaces the `z`-ends-at-zero case as an explicit vanishing-factor branch — probability-bounded in `ChallengePricing` — rather than assuming it away.
3. **The chunked grand product.** This is the subtle one and it is done well: per-chunk telescoping plus the chaining rule make the chunk products one more running product (`perm_copy_constraints_of_chunk_products`), so the single-chunk argument covers the whole table, across chunks of different widths (Orchard's 7/7/1). The identity names use the global column index `c·chunkLen + j` — halo2's offsetting, including the short last chunk — and `chunkName_injective_of_coset` discharges name distinctness from the keygen's δ-coset condition without ever dividing.
4. **The bridge from one field check to the multiset identity.** The two-variable Schwartz–Zippel is layered correctly: a good γ gives the `value + β·name` multisets, a good β through the coefficients of `pairProdDiff` gives the pairs, and `pairProdDiff_ne_zero` (from #21's `prod_pair_inj`) is what makes the β bad set a genuine root set. The lookup variant separates input from table columns via the γ-leading coefficient, and the θ-decompression recovers row tuples from the compressed membership.
5. **Nothing about the checks is assumed.** Every rule is *located* inside the deployed constraint list (`…_mem_permutationExpressions`, `lookup_rules_dvd_of_identity`) and its vanishing is read off the single folded identity via `FoldSplit` — the `y`-split bounded per coefficient, with out-of-range witnesses proven zero so the `∃ j` events are finite unions.
6. **The probability bounds are complete and honest.** `ChallengePricing` bounds each new surface (`y`: `n·|cs|/p`; γ: `|cells|/p`; β: `(|cells|+1)·|cells|/p`; the vanishing-factor escapes: factor count over `p`; θ: `N²·r/p`), and the module doc plus TODO say plainly that combining these into the terminal knowledge-error bound awaits the same sequential-coupling hook as `hgood` — the #67 / PR #96 seam, consistent with the integration design's ownership rule: the bad events are explicit premisses or explicit branches here, and #96 combines their bounds into the composite.
7. **The capstone payload retains its provenance.** `SnarkRelationWithMemberConstraints` keeps `opens`, `batchOpenings`, the advice/instance layouts, `memberDecode`, `quotCommitted`, and `satisfiesCircuit` — the fields the `hencodes` bridge needs — with `.toSnarkRelation` and the terminal wrappers threading `circuitSatViaConstraints` through the deployed chain, and `ConstraintRelations` reading the copy constraints (via `PermConstruction.build` + `build_correct`) and the lookup inclusion back out of the predicate.
8. **Census.** 79 new `assert_axioms` entries cover the new modules' surface.

## Ask

1. `Constraints.lean`'s module doc still ends with "It does not connect the permutation and lookup terms to the circuit-level copy and lookup constraints …; that is separate work. So 'circuit satisfaction' on this path means gate satisfaction." That was true before this PR and is now contradicted by the same file's own "permutation and lookup arguments at the polynomial level" section and by `ConstraintRelations`. Reword to present state: the gates-only path (`combineGates`/`circuitSatViaGates`) is the compatibility shape, and the full path is `combineConstraints`/`circuitSatViaConstraints`.

## Note for the handoff (no change needed in this PR)

The planned `hencodes` bridge theorem concludes with a proposition (`Action.BundleStatement … ∨ …`), but the ledger-model security games one layer up consume *witnesses*: the high-level witness record with the Merkle position and path, the note fields, and the key material. The data is already in this PR's payload — the decoded member columns are the full assignment, and `SnarkRelationWithMemberConstraints` retains them — so the request is only that the hencodes bridge's factoring keep it available (the work staged on Gregor's fork; of the tracking issues, #93 is nearest): a witness-yielding lemma (`TopLevelAssignment` → the high-level witness record and its facts) before the final proposition-level statement. This is issue #97's extraction-friendly-statement point applied to the bridge; it is cheap to provide while the bridge is being designed and awkward to retrofit.

daira:
A post-merge note for the record, from the review pass: how the running product ending at zero is handled, and what remains for the composition.

1. **The deterministic layer never assumes it away.** `RunningProduct.prod_eq_or_factor_eq_zero` telescopes `z_m·∏B = z_0·∏A`; with the boundary rules giving `z_0 = 1` and `z_m ∈ {0, 1}`, the `z_m = 0` case forces `∏A = 0` and the proof surfaces the concrete vanishing factor. That disjunct survives, as data, into every endpoint — `… ∨ ∃ cell, value + β·name + γ = 0` for the permutation, `… ∨ ∃ t, (input + β)·(table + γ) = 0` for the lookup — all the way through `circuitSatViaConstraints`'s read-out theorems. (This mirrors why halo2 constrains `z² − z` at the last row rather than `z − 1`: an honest prover whose factor genuinely hits zero must still have a valid proof, so completeness needs the 0 branch and soundness must bound it.)

2. **The probability bound is proven, and it does not assume an honest prover.** `ChallengePricing.escape_measure_le` bounds the event that some listed factor `v + challenge` vanishes by `#factors / p` over a uniform challenge, for arbitrary committed values — no honesty used. The dependence structure is also non-circular and ready for the transcript coupling: the β bad sets depend only on data committed before β is squeezed, and the γ bad sets and the escape events depend on (data, β), with γ squeezed after β — matching the transcript order.

3. **The remaining step is the declared composition, not a silent assumption.** The per-surface lemma says "for a uniform challenge the escape is negligible"; turning that into "the escape term appears in the deployed game's final ε against an adaptive prover" needs the sequential-conditioning step — the challenge is uniform conditioned on the transcript prefix that pinned the factor values. That is the same coupling hook `hgood` awaits, and `ChallengePricing`'s module doc and TODO assign it to the #67 / PR #96 composition. When PR #96 stages, the check-item is that the vanishing-factor escapes and the `y`/β/γ/θ exclusions all appear in its composite bound.

PR #92, opened 2026-07-23
Fix all Lean linter warnings

mitschabaude:
fixes all warnings in `lake build`, they are a tax on our agents' context


TalDerei (approved):
ACK

daira (commented):
Post-hoc ACK

Issue #97, opened 2026-07-23
Write circuit specs in an extraction-friendly style (Bundle.SpecBase existentials)


daira:
From a post-merge review of #89, focused on whether `Bundle.SpecBase` reflects §4.17.4 with *knowledge* soundness.

In principle, the knowledge soundness property of Halo 2 should allow us to extract the entire witness matrix from the prover of a valid proof. Then, knowing the circuit definition we should be able to extract a high-level witness from that matrix. The arguments for high-level security properties will depend on being able to do so.

The three existentials in `SpecBase` potentially complicate this picture:

* `∃ ivk` — fine: in the non-break branch `ivk` is pinned to `(bp + rivk • commitIvkR).x` with `bp` the deterministic Sinsemilla hash-to-point of the extracted `(akP.x, nk)` chunks, so it is a function of extracted data. The break branch exhibits break data instead.
* `∃ root` — fine, but only by a case-specific argument: `vOld·(root − anchor) = 0` makes `root` equal to the public anchor whenever `vOld ≠ 0`, and the Merkle check is intentionally vacuous for `vOld = 0` (a dummy spend). So `root` is either known or irrelevant.
* **The Merkle path — not fine by any local argument**: `MerkleRoot`/`MerkleStep` existentially quantify all path data (`∃ mid`, `∃ lv rv`), and `SpecBase` never constrains the extracted `wit.merklePath` cells, so the exported statement forgets the path, and in particular the leaf position. The position would be sufficient for any security argument because the tree is public. However, as currently stated, `SpecBase` does not certify the position bits. (I think that the Orchard security arguments happen not to need it, but the Sapling ones do.)

It is possible to `obtain` the witness of the existential, but this is not consistent with "breaks as computed data" and would not allow us to derive a computable reduction.

Proposal: prefer a spec style that guarantees high-level-witness extraction, reserving `∃` for values that are provably and *locally* determined by extracted or public data. This is cheaper than deciding per-existential, per-consumer, whether existence happens to suffice.

Filed using AI (Claude Fable 5).

daira:
[...]

PR #98, opened 2026-07-24
Bridge circuit spec and action statement

gio54321:
Goes in the direction described by #97 and provides a bridge from the circuit spec to the action statement, with a clean split between the circuit-exported specification and the security reduction.

### Description

This PR builds the refinement layer between the deployed Action circuit's postcondition (`Circuits.Action.Circuit.SpecPost`) and the games-facing ledger statement (`Security.Ledger.ActionSatisfied`), and establishes the layering rule for the Sinsemilla exceptional cases: **circuit-exported specs speak only about concrete properties of the partial hash function (the spec's literal ⊥-model); all break statements live in the security layer, where the reduction to break data is a compiler-checked computation.**

The end-to-end result (`Ledger/Bridge.lean`):

```lean
theorem circuit_soundness_to_ledger ... :
    ActionBreak (extract cfg input i₀ env) ∨
      ∃ inst w, PublicProjection (extract cfg input i₀ env) inst ∧
        ActionSatisfied Pool.primitives Pool.keyBinding inst w ∧
        CrossAddressSatisfied (extract cfg input i₀ env) w ∧
        EnableFlagsSatisfied (extract cfg input i₀ env) w
```

A satisfying circuit assignment yields either a fully satisfied ledger action with matching public inputs, or a break — where every `ActionBreak` constructor (Commit^ivk, both note commitments, Merkle) carries a `hashToPointB … = .inr br` equation on the **witness's own extracted chunks**, never a witness-independent escape.

### The spec/reduction split

The circuit specs previously exported or-break disjunctions (`SpecOrBreak`, `MerkleRootStrict ∨ MerkleBreakAt`). These are gone:

- **Guarded exports.** `SpecBase`'s Sinsemilla conjuncts, `CommitIvk.Main.Spec`, `NoteCommit.Main.Spec`, and the Merkle `CalculateRoot` spec are stated with the new `HashGuarded` predicate — `∀ B, hashToPoint … = some B → P B`, the literal `∈ {…, ⊥}` escape shape of §4.17.4/§5.4.1.9 — plus `ExactMerklePathData` for the path. Reviewability against the protocol spec text no longer routes through break vocabulary.
- **The removal is provably not a weakening.** The named equivalence `specOrBreak_hashToPointB_iff_guarded` shows that, given only fixed-base validity facts, the or-break form was equivalent to the guarded form: its break branch is discharged by `validBreak_of_inr` alone and never constrained the witness. The untied `MerkleBreakAt` disjunct (an existential escape related to nothing in `ActionData`) is deleted outright, together with the legacy `MerkleRoot` conjunct and the strict/break-window machinery.
- **The reduction is a computation.** `Bridge.classifyAction : ActionData → Option ActionBreakData` re-evaluates the witness's four exact query families — `ivkQuery` / `noteOldQuery` / `noteNewQuery` / `merkleQuery`, defined once and shared by the break vocabulary, the classifier, and every proof, consuming the raw 255-bit encodings verbatim — with the Σ-refined chain `hashToPointB`, and returns the first escape as Type-valued data (site + `BreakData`). `actionBreak_of_classify` and `classify_none_defined` are its correctness theorems; `successes_or_break` and `specPost_to_ledger` route through it, so the Prop-level refinement and the computed data cannot diverge.
- **Computability is asserted, with an honest Choice caveat.** `BridgeTests` asserts `assert_computable classifyAction +choice`: a plain compiled `def`, so `Classical.choice` provably cannot contribute to the computed break data. Literal choice-freedom is unattainable over `Fp` — even the deployed point literals (`Action.ivkQ`) depend on `Classical.choice` through erased `Prop` proof fields of Mathlib's `ZMod` numeral/arithmetic instances — and the plain-`def` check is exactly the guarantee that choice stays erased.

### Changes by layer

- **Circuit exports** (`Circuits/Action/Bundle.lean`, `Circuits/Sinsemilla/Merkle.lean`): `ActionData` gains the literal per-layer 255-bit child encodings (reconstructed from the decomposition pieces, not the reduced nodes) and the cond-swap position bits; the `ExactMerklePathData` conjunct in `SpecBase` certifies them, threaded through the soundness/completeness proofs of both 16-layer Merkle folds. This addresses the third bullet of #97: the authentication path and leaf position are extraction-friendly functions of witnessed cells.
- **Ledger Merkle interface** (`Security/Ledger/{Merkle,Statement}.lean`): replaces `merkleCRH : B × B → B` with `MerklePrimitives` — a separate raw-encoding type `E` (a field element can have several accepted 255-bit representatives), level-personalized **partial** compression `Fin depth → E × E → Option B`, and a `Merkle.Path` whose layer equations also assert each compression succeeded (`Path.compress_isSome` recovers definedness from the path fact alone). `Merkle.Collision` becomes a success-only `RandomOracle.DefinedCollision`, so escaped queries can never constitute a collision — this keeps the future collision reduction provable against `ActionSatisfied` alone, with no bridge-local invariants.
- **Concrete instantiation** (`Ledger/Pool.lean`, `KeyBinding/Pool.lean`, `Concrete/PallasGroup.lean`): the deployed pool over Pallas (`PallasGroup` wrapper providing the `Fq`-module structure and `NoZeroSMulDivisors`), the real Sinsemilla domain points, and the deployed circuit's key-binding interface (bare `SinsemillaHash(ak.x, nk) + [rivk]·CommitIvkR` opening plus `ivk ≠ 0`, derived from on-curve facts) — intentionally separate from the Recovery Statement adapter.
- **Side conditions preserved with exact field semantics** (no booleanity assumed): `CrossAddressSatisfied` fires for *any* nonzero `disableCrossAddress`; `EnableFlagsSatisfied` forces the corresponding note value to zero unless the flag is `1`. The signed-magnitude ↔ integer-subtraction no-wrap argument for value commitments is proved from the 64-bit bounds.
- **Regression checks** (`Ledger/BridgeTests.lean`): both value-sign directions, non-canonical 255-bit zero encodings, dummy-spend vacuity, three-valued cross-address flag, per-layer definedness, disabled-flag consequences, the or-break ↔ guarded equivalence instantiated at the deployed `Commit^ivk` domain point, `assert_no_sorry` and axiom audits on the exported theorems, and `assert_computable` on the classifier.

### Relation to #97

The Merkle path — the one existential #97 flags as unrecoverable "by any local argument" — is certified against extracted cells, and the bridge consumes only that exact data. The two residuals flagged in earlier revisions of this PR are now resolved in-branch: the legacy `MerkleRoot` and `MerkleRootStrict ∨ MerkleBreakAt` conjuncts are removed (reshaping the circuit soundness proofs turned out to be a pure deletion), and the refinement now carries a Type-valued break classifier over the extracted `ActionData`. The deterministic conversion from a classified `BreakData` to the games-facing Sinsemilla-DLR break object is resolved.

### Verification

Full `lake build` (8754 jobs) — zero errors, zero warnings, no sorries. Break vocabulary and bridge shape cross-reviewed against #97; the or-break removal cross-validated with GPT-5.6 Sol (codex) before implementation.

Assisted by Claude Fable 5


Zcash/Circuits/Action/Bundle.lean
```diff
@@ -231,6 +245,27 @@ def extract (cfg : Config) (_ : Var PrivateInputs Fp) (i₀ : RegionIndex)
...
   pkdNew := ⟨cellRead env (i₀ + 348) 0 cfg.eccConfig.witnessPoint.x,
              cellRead env (i₀ + 348) 0 cfg.eccConfig.witnessPoint.y⟩
+  leftEncoding := fun j =>
+    let k := j.val
+    let mcfg := if k < 16 then cfg.merkle1 else cfg.merkle2
+    let base := if k < 16 then i₀ + 8 + 8 * k else i₀ + 136 + 8 * (k - 16)
+    let a := (cellRead env (base + 1) 0 mcfg.sinsemilla.witnessPieces).val
+    let b := (cellRead env (base + 4) 0 mcfg.sinsemilla.witnessPieces).val
+    let b1 := (cellRead env (base + 2) 0 cfg.lookupConfig.runningSum).val
+    a / 2 ^ 10 + 2 ^ 240 * (b % 2 ^ 10 + 2 ^ 10 * b1)
```
mitschabaude, line 255: nice

Zcash/Circuits/Action/Bundle.lean
```diff
@@ -273,30 +320,27 @@ def SpecBase (G : Generators) (B : Bases) (wit : ActionData) : Prop :=
...
-      (noteScalars wit.gdNew wit.pkdNew wit.vNew wit.nfOld wit.psiNew).chunks) ∧
-  -- Merkle path validity, tied through the `q_orchard` anchor check. The strict-or-break
-  -- refinement exhibits any Sinsemilla escape as data, as requested by zcash/ironwood#45.
+  -- `Extract(NoteCommit(…)) ∈ {cmx, ⊥}` (guarded ⊥-model)
+  HashGuarded G.S B.noteQ
+    (noteScalars wit.gdNew wit.pkdNew wit.vNew wit.nfOld wit.psiNew).chunks
+    (fun bp => wit.cmx = (bp + wit.rcmNew.2 • B.noteCommitR).x) ∧
+  -- Merkle path validity, tied through the `q_orchard` anchor check: the exact
+  -- raw-encoding chain of the witnessed path cells (zcash/ironwood#97), each layer's
+  -- hash in the guarded ⊥-model.
```
mitschabaude, line 340: interesting i was pushing my models in the opposite direction (to use SpecOrBreak) but if it's equivalent doesn't matter I guess

Zcash/Circuits/Action/RealBases.lean
```diff
@@ -69,19 +69,4 @@ def orchardBases : Circuit.Bases where
...
-`Unconstrained` hint pattern: the input is the witness programs, the verifier value is
-erased, the prover value is the evaluated `WitnessData`). Soundness is knowledge
-soundness at the constructive extractor: for ANY assignment satisfying the
-constraints, the `ActionData` read off the cells satisfies the §4.17.4 spec; the
-input matters only for completeness (the honest prover running those programs
-satisfies the constraints). -/
-def orchardActionCircuit :
-    FormalCircuit Fp Unit Circuit.Config Circuit.PrivateInputs unit :=
-  Circuit.circuit Specs.Sinsemilla.orchardGenerators orchardBases
-
```
mitschabaude, line 86: why delete this? I've been using it

Zcash/Circuits/Sinsemilla/Merkle.lean
```diff
@@ -1965,29 +2029,39 @@ def circuit :
...
   Assumptions _ := True
 
-  -- the 32 per-layer (sibling, swap) cell readings, at 8-region stride
-  Witness := fun F => ℕ → F × F
-  extract := fun (ccfg, _, _) _ i₀ env => fun j =>
-    (eval env (AssignedCell.of (i₀ + 8 * j) 0 ccfg.b : Var field Fp),
-     eval env (AssignedCell.of (i₀ + 8 * j) 0 ccfg.swap : Var field Fp))
+  -- The complete per-layer extraction: cond-swap readings plus the literal
+  -- decomposition encoding used by the following seven-region hash layer.
+  Witness := fun _ => ℕ → Layer.Witness
```
mitschabaude, line 2034: 👍🏻

Zcash/Security/Concrete/PallasGroup.lean
mitschabaude, line 1: I feel like we should already have "`SWPoint` instantiated at `Pallas.curve`" somewhere.. maybe in `CompElliptic`?

Zcash/Security/Ledger/Bridge.lean
```diff
@@ -0,0 +1,919 @@
...
+-/
+
+namespace Zcash.Security.Ledger.Bridge
+
+open Zcash.Circuits
+open Zcash.Circuits.Action
+open Zcash.Circuits.Action.Circuit
+open Zcash.Circuits.Specs.Sinsemilla
+open Zcash.Security.Concrete
+open Zcash.Security.Ledger
```
mitschabaude: nit: last open is redundant (llms never get Lean namespaces)
```suggestion
namespace Zcash.Security.Ledger.Bridge

open Zcash.Circuits
open Zcash.Circuits.Action
open Zcash.Circuits.Action.Circuit
open Zcash.Circuits.Specs.Sinsemilla
open Zcash.Security.Concrete
```

Zcash/Security/Ledger/Bridge.lean
```diff
@@ -0,0 +1,919 @@
...
+    (hconstraints : Constraints env.place env.env
+      ((mainPost orchardGenerators orchardBases cfg input).operations i₀) i₀) :
+    ActionBreak (extract cfg input i₀ env) ∨
+      ∃ inst w, PublicProjection (extract cfg input i₀ env) inst ∧
+        ActionSatisfied Pool.primitives Pool.keyBinding inst w ∧
+        CrossAddressSatisfied (extract cfg input i₀ env) w ∧
+        EnableFlagsSatisfied (extract cfg input i₀ env) w := by
+  have hpost := soundnessPost orchardGenerators orchardBases cfg i₀ env input
+    henv trivial hconstraints
+  exact specPost_to_ledger (by simpa using hpost)
```
mitschabaude: awesome!

mitschabaude (commented):
looks great, probably one for @daira to review

Zcash/Circuits/Action/Bundle.lean
TalDerei, line 168: stale docs; there's now ten public input rows for the NU6.3 circuit implementation https://github.com/zcash/orchard/pull/504.

Zcash/Security/Ledger/BridgeTests.lean
TalDerei, line 1: `BridgeTests.lean` this isn't compiled as part of the CI and not imported by `Security.lean`, so assertions are missing from CI.

Zcash/Security/Ledger/Bridge.lean
```diff
@@ -0,0 +1,918 @@
...
+          dsimp [w]
+          simp [hz]
+        exact (sub_eq_zero.mp ((mul_eq_zero.mp heo).resolve_left hvNew0)).symm
+  · exact Or.inl hbreak
+
+/-- End-to-end refinement for a satisfying run of the deployed post-NU6.3
+Action circuit.  This is the direct composition of the circuit soundness
+contract with `specPost_to_ledger`; it deliberately leaves the exceptional
+Sinsemilla branch explicit and reports the enable gates as `EnableFlagsSatisfied`. -/
+theorem circuit_soundness_to_ledger (cfg : Config) (i₀ : RegionIndex)
```
TalDerei: `circuit_soundness_to_ledger` and internal caller `soundnessPost` need to be checked in top-level `Zcash/TrustBoundary.lean`, otherwise it's a trust policy hole.

Zcash/Security/Ledger/Statement.lean
```diff
@@ -116,15 +125,14 @@ satisfy this interface (the latter enforces strictly more).
...
     (kv : KeyBindingInterface KW G IVK NK) (inst : ActionInstance G RT RHO CMX)
-    (w : ActionWitness KW F G RHO PSI RT P.depth) : Prop where
+    (w : ActionWitness KW F G RHO PSI E P.merkle.depth) : Prop where
   /-- Spend-side commitment integrity: `cm_old` opens `note_old` with `rcm_old`. -/
   commit_old : P.noteCommit w.rcm_old w.note_old = some w.cm_old
   /-- Merkle path validity for nonzero-valued spends. -/
   merkle_path : w.note_old.v ≠ 0 →
-    Merkle.pathRoot P.merkleCRH P.depth w.pos w.path
-      (P.leafOf (P.extract w.cm_old) w.note_old.ρ) = inst.rt
+    Merkle.Path P.merkle (P.leafOf (P.extract w.cm_old) w.note_old.ρ) inst.rt w.path w.side
```
TalDerei: #98 modified `ActionSatisfied`, but seems to leave the original TODO unchanged?

there's a few things open here; we need to:
- convert `ActionBreakData` into the games-facing DLR break,
- turn two conflicting valid merkle paths into a concrete hash collision

so break reduction per https://zcash.github.io/ironwood/formal-verification.html remains incomplete. 

also `SpecPost` is the circuit statement, and `ActionSatisfied` is the security-game statement (the abstract ledger meaning of a successful action, introduced in https://github.com/zcash/ironwood/pull/44). `SpecPost` and `ActionSatisfied` should remain separate, but we should verify that every `SpecPost` case becomes either `ActionSatisfied` or `ActionBreak`.

Zcash/Security/Ledger/Merkle.lean
```diff
@@ -132,3 +132,84 @@ theorem pathRoot_authPath (H : B × B → B) (d : ℕ) (f : (Fin d → Bool) →
...
+      simp [Fin.val_succ, Fin.val_last, heq]
+    have hroot := h.2
+    rw [← hlast, hnode] at hroot
+    exact ⟨_, hroot⟩
+
+/-- A Merkle collision records the layer whose personalization was collided, and
+demands *successful* evaluations on both queries (a `DefinedCollision`): an escaped
+layer, whose compression is `none`, can therefore never constitute a collision. -/
+abbrev Collision (P : MerklePrimitives B E) :=
+  Σ i : Fin P.depth, RandomOracle.DefinedCollision (P.compress i)
```
TalDerei: where is the collision reduction for this new partial merkle interface? afaict the existing `collisionOfWrongLeaf` only supports the old total hash interface and can't produce this `DefinedCollision`. Is that planned here or tracked separately?

Zcash/Security/Ledger/Bridge.lean
```diff
@@ -0,0 +1,918 @@
...
+    (hconstraints : Constraints env.place env.env
+      ((mainPost orchardGenerators orchardBases cfg input).operations i₀) i₀) :
+    ActionBreak (extract cfg input i₀ env) ∨
+      ∃ inst w, PublicProjection (extract cfg input i₀ env) inst ∧
+        ActionSatisfied Pool.primitives Pool.keyBinding inst w ∧
+        CrossAddressSatisfied (extract cfg input i₀ env) w ∧
+        EnableFlagsSatisfied (extract cfg input i₀ env) w := by
+  have hpost := soundnessPost orchardGenerators orchardBases cfg i₀ env input
+    henv trivial hconstraints
+  exact specPost_to_ledger (by simpa using hpost)
```
TalDerei: should `EnvAssumptions` remain an assumption here? assuming gadget-level theorems can keep it as explicit assumptions, but deployed capstones should discharge them which they currently don't? in that context, doesn't `circuit_soundness_to_ledger` remain conditional? not sure what exactly what be in `EnvAssumptions` scope though.

Zcash/Security/Ledger/Statement.lean
TalDerei: should `ActionInstance` be packaged with the flags? like the `circuit::Instance` contains the five action values plus `enable_spend, enable_output, cross_address_disabled` (the  bundle-level flags are copied into every action’s circuit instance. 

reference: https://github.com/zcash/orchard/blob/8b9d310b106fc259705f915413728ab07d2ccf8a/src/circuit.rs#L1147-L1274 and https://github.com/zcash/orchard/blob/8b9d310b106fc259705f915413728ab07d2ccf8a/src/bundle.rs#L35-L50.

Zcash/Security/Ledger/Bridge.lean
```diff
@@ -0,0 +1,918 @@
...
+  | noteCommitNew
+  | merkle (i : Fin 32)
+deriving DecidableEq, Repr
+
+/-- A classified escape: the site, together with the escape data of the witness's
+own exact hash query at that site. -/
+structure ActionBreakData where
+  site : BreakSite
+  data : BreakData
+deriving Repr
```
TalDerei, line 392: can we add a TODO here?

```ts
  -- TODO: Add a computable conversion from a classified ActionBreakData
  -- into the games-facing Sinsemilla/DLR break object.
```

Zcash/Security/Ledger/Merkle.lean
TalDerei: can we add a TODO here?

```ts
  -- TODO: Add the computable position-binding reduction from conflicting
  -- valid paths for this partial interface to a DefinedCollision.
```

Zcash/Security/Ledger/Statement.lean
TalDerei, line 153: can we add a TODO below this?

```ts
  -- TODO: When the Balance and Spendability games are formalized, verify
  -- their exact Action interface, including whether the NU6.3 flag
  -- conditions belong in the game instance or a circuit-facing wrapper.
```

Zcash/Security/Ledger/Bridge.lean
TalDerei: like currently, `ActionBreakData` just records why `Sinsemilla` evaluation failed, but it doesn't do the step where you take `ActionBreakData` and have a plain def converts it into `NontrivialRelation`.

TalDerei (commented):
meta observation; this still needs work to model closer to the "[break-as-computed-data](https://zcash.github.io/ironwood/formal-verification.html)" pattern in https://github.com/zcash/ironwood/pull/46, which is incomplete here afaict (see feedback below)?

Zcash/Security/Ledger/Bridge.lean
mitschabaude: I got rid of all top level assumptions in #99, so when merged together this can be removed

TalDerei: should've assumed something like that gets handled in #99!

TalDerei:
more intermediate review feedback in https://github.com/zcash/ironwood/pull/103.

Zcash/Security/Ledger/BridgeTests.lean
TalDerei: fixed in https://github.com/zcash/ironwood/pull/103.

Zcash/Security/Ledger/Bridge.lean
TalDerei: fixed in https://github.com/zcash/ironwood/pull/103.

Zcash/Circuits/Action/Bundle.lean
daira: The equivalence is real, but it's worth spelling out exactly what it does and doesn't say.

What each form says: `HashGuarded S Q chunks P` is literally the spec's ⊥-model — if the Sinsemilla chain over these chunks is defined, the payload holds; escapes are unconstrained. `SpecOrBreak S Q P (hashToPointB … chunks)` evaluates the break-aware chain and demands: on `.inl B` the payload, and on `.inr br` that the escape data is a valid break.

Why they're equivalent: the forward direction is trivial. The backward direction (`breaksOfGuarded`) discharges the break arm by `validBreak_of_inr` using only fixed-base validity — facts about the deployed constants, independent of any circuit satisfaction. That's the crux: the break arm is fulfillable without additional information or hypotheses *about the witness*. Since it is witness-independent, carrying it inside each exported circuit spec adds no constraint that the guarded form lacks.

The differences that survive the equivalence — why the export surface standardizes on guarded:

1. **Reviewability.** Guarded exports are textually §4.17.4/§5.4.1.9's `∈ {…, ⊥}` shape, so auditing circuit specs against the protocol spec compares like with like, with no formalization-internal break vocabulary on the trusted statement surface.
2. **Provenance of break data.** In the or-break style, the "data" sits in `Prop` positions threaded through every composed proof — extracting it computably needs choice or a recomputation anyway. In the guarded style it is produced once, in the security layer, by re-running `hashToPointB` on the same witnessed chunks, as a compiled `def` under `assert_computable +choice`, with `actionBreak_of_classify` proving that the `Prop`-level refinement and the computed data coincide. That is the breaks-as-computed-data split: `Prop`s for specs, `def`s for reductions.
3. **Churn isolation.** When `BreakData` evolves (it will, when the games-facing Sinsemilla-DLR object lands), guarded exports do not move.
4. **The boundary where the equivalence would fail** — the reason this is a design decision rather than a coin flip: it holds *because* the break arm is dischargeable witness-independently. A spec whose break arm carried witness-entangled strength (say, an escape tied to specific constrained cells) would make or-break strictly stronger than guarded, and then it should be stated deliberately. The deleted `MerkleBreakAt` was the opposite cautionary case: an untied existential that constrained nothing. So the audit rule the lemma establishes: a break disjunct can either be discharged independently of the witness (keep the export guarded) or it can't (real strength — state it on purpose).

Practically for your models: keeping `SpecOrBreak` internally costs nothing — `breaksOfGuarded` converts at the export boundary in the cheap direction, given only base validity.

Zcash/Security/Ledger/Statement.lean
mitschabaude: > we should verify that every SpecPost case becomes either ActionSatisfied or ActionBreak.

Isn't that exactly what circuit_soundness_to_ledger in this PR establishes?

gio54321: > Isn't that exactly what circuit_soundness_to_ledger in this PR establishes?

Do we maybe want a more constructive reduction, i.e. a function rather than a theorem, so it is easier to inspect to be PPT?

TalDerei: > Isn't that exactly what circuit_soundness_to_ledger in this PR establishes?

you're right on that part, missed it. everything else in there though is still valid.

TalDerei:
Audit note: relationship to the soundness review of #99

I reviewed #99 for specification bugs, implementation-correctness bugs, and soundness gaps. The main findings were:

1. The #99 capstone places Sinsemilla `ValidBreak` / `MerkleBreakAt` cases inside the successful `BundleStatement` branch. Consequently, the final `BundleStatement ∨ URS relation` conclusion can discharge a generator-DLR failure through what is presented as the successful Action branch, rather than exposing it as the intended security-game break.
2. The #99 `PublicInputs` boundary consists of raw field elements. On degenerate witnesses, the flag rows need not be Boolean, and the boundary does not itself enforce a nonidentity `rk`. That is weaker than a typed Orchard transaction-instance interface.
3. The deployed instance-commitment theorem recomputes an `instanceKey`, while the certificate fact connecting the derived Lagrange prefix to the captured Rust/deployed family is not consumed by the terminal theorem. This leaves a deployment-faithfulness/transport obligation, even if the internal algebra is sound.

TalDerei:
Audit of the spec, implementation, and soundness of this PR, re-checked against #103 (which lands on this branch's head, 9c23b45) — findings that #103 already fixes are omitted, so what follows is what I believe is still open.

**Summary: the bridge architecture looks sound to me — I could not break the refinement logic, and every spec-conformance surface I could check externally (deployed domain-point constants, message layouts, layer personalization, signed-value arithmetic, nullifier derivation) checks out. With #103 applied I found nothing that affects a currently-proved theorem. Two latent issues remain: a break structure that is inhabitable without cryptographic content, which will block the reduction it was built for, and a missing lemma at the consumer boundary.**

I read the full diff plus the complete files it depends on (`Statement.lean`, `Merkle.lean`, `SinsemillaBreak.lean`, the `SpecBase`/`SpecPost`/`extract` regions of `Bundle.lean`, `AxiomCheck.lean`, `RealBases.lean`), then re-verified each finding against #103's head (49298c4).

## Findings

### 1. `NoteCommitBreak` is inhabitable by value-overflow aliasing, blocking the advertised onward reduction

`Note.v` is an unbounded `ℕ`; `Pool.noteCommit` feeds it through `(n.v : Fp)` and `noteScalars`' 64-bit truncation. `NoteCommitBreak` (`Statement.lean:204`) requires only distinct `(rcm, note)` tuples with extract-equal commitments and carries no value bounds — so `n₂ := n₁ with v := v + 2^64` yields a "break" with zero cryptographic content.

The structure predates this PR, but the concrete `Pool` instantiation added here makes the aliasing real, and the docs position "`NoteCommitBreak` → Sinsemilla/DLR relation" as the intended next step. That theorem is **false** as stated and will need `v < valueBound` fields on the structure. The rest of the note tuple is fine: `gd`/`pkd` are group elements with injective affine encoding, and `ρ`/`ψ` are field elements, so no other component aliases.

Nothing currently proved is unsound — the games only mint breaks from `ActionSatisfied` pairs, which carry `v_old_lt`/`v_new_lt` — but the bound has to travel with the break object rather than being recoverable only at the call site, or the reduction cannot be written against `NoteCommitBreak` alone.

## Soundness boundaries

### 1. Accepted proof → constraints

Already flagged in the [`EnvAssumptions` thread](https://github.com/zcash/ironwood/pull/98#discussion_r3647372998); explicit composition with the deployed verifier is owned by #99.

### 2. No named bridge from Prop-level `ActionBreak` to the computed classifier

`circuit_soundness_to_ledger`'s left arm is `ActionBreak (extract …)`, but the computable reduction path starts at `classifyAction`/`classifyRelation`. The implication `ActionBreak wit → (classifyAction wit).isSome` is derivable — via `classify_none_defined` plus `hashToPoint_of_inr` — but it is not stated, so a consumer landing in the break arm has to reconstruct the glue itself to reach the `ActionDLBreak` object. Given the PR's own design goal that "the Prop-level refinement and the computed data cannot diverge," an exported `actionBreak_iff_classify_isSome` would close that loop at the consumer boundary as well as internally.

The deterministic DLR and Merkle-collision pieces previously flagged are now implemented; item 2 is the remaining consumer-side bridge before those break objects can be connected downstream to the terminal security games and probability bound.

Minor nit in the same file: `signedMagnitude_eq_int_sub` proves `(-1 : Fp) ≠ 1` by `native_decide` (`Bridge.lean:86`) where `decide` or `norm_num` would do. #103 makes this inert in practice rather than removing it — by pinning `specPost_to_ledger` and the other exported endpoints at `+native`, it documents that they already sit at the `native_decide` tier via the fixed-base certificates, so the gratuitous one adds no tier. The value lemmas that stay in the standard tier (`value_positive`, `value_negative`, `value_equal`) route through `int_sub_eq_of_fp_sub_eq` instead, so they are unaffected either way.

## What I checked and found correct

**Soundness architecture.** The central design — circuits export only guarded ⊥-model statements (`HashGuarded`, `ExactMerklePathData`), and the security layer re-runs `hashToPointB` on the witness's own four query families — is airtight as far as I could push it. The classifier's queries are *the same terms* (shared `abbrev`s) as the guarded clauses' queries, so drift is impossible by construction. When any hash escapes, the guarded clauses go vacuous exactly where the classifier reports a break; when none does, every guarded clause fires with the classifier-supplied witness, because `successes_or_break` cases on `classifyAction` itself and so no witness can fall between the disjunction's arms. `specOrBreak_hashToPointB_iff_guarded` is the correct equivalence, and the removal claim is honest: the old `SpecOrBreak` break branch really was dischargeable from fixed-base validity alone, and the deleted `MerkleBreakAt` was an untied existential that constrained nothing. The new `ExactMerklePathData` is strictly *stronger* than the deleted `MerkleRoot` export, since it pins encodings and side bits to extracted cells.

**Merkle chain.** The `GuardedPath → Path` reconstruction is correct at both endpoints (leaf case `i = 0`, root case at height 31). Per-height definedness (`compress_isSome`) means escaped compressions can never masquerade as nodes or as collisions, since `DefinedCollision` is success-only. Non-canonical 255-bit representatives are handled correctly by the encoding-domain interface: sibling encodings are free, on-path encodings are pinned only to the decode fibre, which matches the deployed circuit, since it does not pay for a canonicity check. The encoding reconstruction formulas (`a/2^10 + 2^240·(b%2^10 + 2^10·b1)` and `b2 + 2^5·c`) match the 250∥10+5+5∥250 decomposition, and the cell offsets line up across all four extract layers — any drift there would fail elaboration.

**External spec conformance.** I decoded the three domain points from upstream `zcash/orchard`'s `constants/sinsemilla.rs` byte arrays: `Q_MERKLE_CRH`, `Q_COMMIT_IVK_M_GENERATOR`, and `Q_NOTE_COMMITMENT_M_GENERATOR` all match the Lean literals digit-for-digit, and `Pool`'s duplicates match `RealBases` with proofs that would fail if they diverged. `merkleChunks` (10+255+255 little-endian bits into 52 chunks) matches §5.4.1.3; `commitIvkChunks` (510 bits into 51 chunks, no padding) matches §5.4.8.4; leaf-up layer personalization 0..31 equals the spec's `MerkleDepth − 1 − layer`. `deriveNullifier` matches the deployed `extract_p(cm + [mod_r_p(PRF^nf + ψ)]K)`, sound because p < q, which is proved in-PR. The 2^64 no-wrap argument is valid, since 2^65 < p. `ivk ≠ 0` is *derived* rather than assumed — the identity has no on-curve affine encoding and `pk_d` is witnessed on-curve — matching how the deployed circuit enforces it. Enable and cross-address flags keep exact field semantics with no booleanity assumption and are correctly read off the instance column; their exclusion from `ActionSatisfied` is a documented TODO, not a hole.

One inherited divergence worth carrying forward: the final blinding additions in `noteCommit` and `commitIvk` are modeled as complete addition. That is faithful to deployed halo2_gadgets `CommitDomain::commit`, but the protocol spec's literal `SinsemillaCommit` uses incomplete ⸭, so any future lemma equating `Pool.noteCommit` with the spec-text function will need the negligible-escape caveat.

**DLR reduction.** `ofPoint_hashToPoint`'s generator combination is the correct Sinsemilla unrolling. Each `breakCoeffs` case produces the right relation with the right sign conventions, and nontriviality never depends on table-entry distinctness: every kind carrying a prefix also carries the `2^n` coefficient on `Q`, nonzero in odd-order `Fq`, and `genZero` is a genuine identity-generator event. `assert_computable`'s combination of a plain-`def` check and an axiom-closure check genuinely delivers the "choice stays erased" guarantee claimed.

**Trust inventory** as amended by #103: `native_decide` certificates only — the six fixed-base window tables, small `interval_cases` facts, and CompElliptic's Pallas point count. No declared axioms remain on this path; `pallas_natCard` is now a theorem delegating to `CompElliptic.Curves.Pasta.Pallas.card_eq`, which exists in the already-pinned CompElliptic revision and is backed by the `native_decide` witness `q_nsmul_Gpt`. The book text is appropriately honest that this swapped a bespoke axiom for a re-checkable compiler-trust certificate rather than for nothing. Separately, the auto-generated `orchardGenerators` table is proved on-curve in Lean but its values are pinned by being committed source, checked against the deployed VK by the fixture machinery rather than re-derived from `GroupHash` — that remains the one input certified by provenance rather than by proof.

**Scope caveat.** All of this is source-level review: I read the proofs and checked constants, layouts, import closure, and declaration existence. I did not run `lake build`, so I cannot confirm elaboration — and in particular #103 is the first time several of the `assert_*` checks execute at all, so that CI run is worth watching.

---
Review assisted by Claude Fable 5.

TalDerei (approved):
optimistic ACK after [findings](https://github.com/zcash/ironwood/pull/98#issuecomment-5083866128) below are addressed and #103 lands.

gio54321:
Findings should be resolved. Thanks for the review @TalDerei!

Zcash/Common/DiscreteLogRelation.lean
```diff
@@ -29,4 +29,15 @@ structure NontrivialRelation {n : ℕ} (g : Fin n → G) (U W : G) where
   nontrivial : a ≠ 0 ∨ α ≠ 0 ∨ β ≠ 0
   relation : (∑ i, a i • g i) + α • U + β • W = 0
 
+/-- The one-distinguished-generator variant: a nontrivial `F`-linear relation among the
+family `g` and a single distinguished generator `U`.  This is the shape of a Sinsemilla
+discrete-log break (protocol spec Theorem 5.4.4): a relation among the per-chunk generator
+table and the domain point `Q`, produced as data by the escape reduction
+(`Zcash.Security.Ledger.Bridge.relationOfBreakData`). -/
+structure NontrivialRelationOne {n : ℕ} (g : Fin n → G) (U : G) where
```
daira, line 37: This isn't strictly necessary because you could set `W` to the identity point and `β` to 1 in `NonTrivialRelation`. I don't particularly mind it though.

Zcash/Security/Concrete/PallasGroup.lean
daira: The examples in [CompElliptic.Curves.Pasta](https://github.com/daira/CompElliptic/blob/main/CompElliptic/Curves/Pasta.lean) just write `SWPoint Pallas.curve`. I don't really see why `PallasGroup` needs to be a wrapper structure; the intended usage is something like `abbrev PallasGroup = SWPoint Pallas.curve`. Then it's automatically an `[AddCommGroup]`, etc., and this whole file becomes just a few lines.

Anyway, I don't mind adding this now to avoid churn in this PR, and simplifying it later.

Zcash/Security/Concrete/PallasGroup.lean
```diff
@@ -0,0 +1,213 @@
...
+  intro hx
+  rcases p with ⟨x, y⟩
+  subst hx
+  exact Point.no_onCurve_of_x_zero y hp
+
+/-- The `Extract_ℙ` ±-fibre property: two group elements have equal extracted
+`x`-coordinates exactly when they agree up to sign.  The affine encoding of the
+identity is `(0, 0)` and no on-curve point has `x = 0`, so the identity is alone in
+its fibre. -/
+theorem toPoint_x_eq_iff (P Q : PallasGroup) :
```
daira, line 165: Moving this to CompElliptic is tracked in https://github.com/daira/CompElliptic/issues/12.

Zcash/Security/Ledger/BridgeTests.lean
```diff
@@ -0,0 +1,269 @@
...
+assert_axioms cross_address_flag_one
+assert_axioms cross_address_flag_arbitrary_nonzero
+assert_axioms enable_spend_disabled_forces_zero
+assert_axioms enable_output_disabled_forces_zero
+
+-- The onward reduction from classified break data to the games-facing
+-- discrete-log-relation object is likewise a computation: the coefficients are a
+-- plain compiled `def` over the break datum, with the relation and nontriviality
+-- facts in erased `Prop` fields (same `+choice` reading as the classifier above).
+assert_no_sorry ofPoint_hashToPoint
```
daira: There is basically no reason to use `assert_no_sorry` in this repo any more. Both `assert_axioms` and `assert_computable` imply `assert_no_sorry`, and we always want to make some assertion about the axioms.

Zcash/Security/Ledger/Pool.lean
```diff
@@ -0,0 +1,239 @@
...
+constants.
+-/
+
+namespace Zcash.Security.Ledger.Pool
+
+open Zcash.Circuits
+open Zcash.Circuits.Specs.Sinsemilla
+open Zcash.Security.Concrete
+
+def merkleQ : Point Fp :=
```
daira, line 40: Document where all of these constants come from, in the Rust code and spec. This can be a follow-up but it's necessary.

Zcash/Security/Ledger/Pool.lean
```diff
@@ -0,0 +1,239 @@
...
+def noteHash (n : Note PallasGroup Fp Fp) : Option (Point Fp) :=
+  hashToPoint orchardGenerators.S noteQ (noteScalars n).chunks
+
+/-- The deployed note commitment: the Sinsemilla hash of the note message, plus the
+blinding term `[rcm] NoteCommitR`.
+
+The final blinding addition is modeled as the *complete* group addition — faithful to
+the deployed `halo2_gadgets` `CommitDomain::commit`, whose Sinsemilla chain ends before
+the blinding term is added with complete addition.  The protocol spec's literal
+`SinsemillaCommit` (§5.4.8.4) composes the blinding with incomplete `⸭` instead, so a
```
daira: Blocking: that is not correct; §5.4.8.4 uses complete `+` to add the blinding.

Image

So does the `orchard` circuit implementation, delegated to `halo2_gadgets`:
https://github.com/zcash/halo2/blob/d5e9ee8ed8072e5155acfa230a97ecb82f173115/halo2_gadgets/src/sinsemilla.rs#L503

That is necessary for soundness; an adversary could choose an $r$ (in practice $\mathsf{rcm}$ or $\mathsf{rivk}$) that forces an exceptional case, for example $r = 0$.

Zcash/Security/Ledger/Pool.lean
```diff
@@ -0,0 +1,239 @@
...
+      (commitIvkChunks ak.val nk.val) = some p) (hp : p.Valid) :
+    commitIvkHash ak nk = some (PallasGroup.ofPoint p hp) := by
+  unfold commitIvkHash
+  rw [h, Option.bind_some]
+  exact PallasGroup.ofPoint?_eq_some p hp
+
+/-- The deployed circuit's key-binding interface: the bare `Commit^ivk` opening
+`Extract(hashPoint + [rivk] CommitIvkR)` plus `ivk ≠ 0`.  As with `noteCommit`, the
+blinding addition is the complete group addition (deployed-gadget-faithful), not the
+spec text's incomplete `⸭`. -/
```
daira: Ditto; the spec uses complete addition. Blocking.

daira (changes requested):
There are incorrect assertions about the spec using incomplete curve addition to add Sinsemilla blinding in Zcash/Security/Ledger/Pool.lean.

Zcash/Security/Ledger/SinsemillaDLR.lean
```diff
@@ -0,0 +1,485 @@
...
+  obtain ⟨hsplit, -, -⟩ := hashToPointB_inr h
+  have hb := siteQuery_bounded wit s
+  rw [hsplit] at hb
+  exact ⟨fun m hm => hb m (List.mem_append_left _ hm),
+    hb br.chunk (List.mem_append_right _ List.mem_cons_self)⟩
+
+/-- **The conversion requested by the break-reduction plan**: a plain `def` turning
+a classified `ActionBreakData` into the games-facing discrete-log-relation object at
+the escaped site's domain point.  The hypothesis ties the datum to the witness's own
+query; `classify_query_inr` discharges it for the classifier's output. -/
```
daira: Nit: it's not necessary to refer to development plans:

```suggestion
/-- **Reduction from a classified `ActionBreakData` into the games-facing discrete-log-relation**
at the escaped site's domain point. The hypothesis ties the datum to the witness's own
query; `classify_query_inr` discharges it for the classifier's output. -/
```

Zcash/Security/Ledger/Statement.lean
```diff
@@ -147,9 +152,17 @@ structure ActionWitness (KW F G RHO PSI MHASH MENC : Type*) (d : ℕ) where
...
+action — and it is deliberately separate from the circuit-facing `SpecPost`:
+`Bridge.specPost_to_ledger` verifies that every `SpecPost` case becomes either this
+statement or an exhibited `Bridge.ActionBreak`. The interface itself stays
+`Prop`-only; the breaks-as-computed-data pattern applies at the bridge
+(`Bridge.classifyAction`, reduced onward to the games-facing discrete-log-relation
+object by `Bridge.relationOfBreakData`), not here.
+
+TODO: Verify the exact Action interface the Balance and Spendability games consume,
+including whether the NU6.3 flag conditions belong in the game instance or a
+circuit-facing wrapper. -/
```
daira: They belong in the game instance, because the adversary must be allowed to submit transactions both before and after NU6.3.

daira (commented):
Reviewed the full diff at the current head. Verdict: this delivers the #97 witness-yielding shape end to end, and the layering (guarded circuit exports, breaks computed in the security layer) is the right split. Approve once the already-posted blocking items land (the complete-addition docstring corrections in `Pool.lean`, and the other open threads). The findings below are the residue not already covered by existing comments.

## What I verified

- `Bridge.lean`: `specPost_to_ledger` constructs all thirteen `ActionSatisfied` conjuncts from `SpecPost`, in the `∃ inst w` witness-yielding form, with `PublicProjection` tying the five public values. The dummy-spend gate (`vOld · (root − anchor) = 0`) correctly makes the Merkle conjunct conditional on nonzero value. The signed-magnitude ↔ integer-subtraction bridge rules out field wraparound from exactly the 64-bit bounds.
- The classifier discipline: `classifyAction` re-evaluates the same four query families the specs and break vocabulary are stated over; `actionBreak_iff_classify_isSome` pins the Prop/computation agreement at the consumer boundary.
- The strict/guarded Merkle split is lossless (`path_iff_guarded_defined`), and the bridge assembles the strict `Path` from the guarded export plus per-layer definedness rather than re-deriving the node chain.
- `SinsemillaDLR.lean`: the chain-combination lemma plus per-kind coefficients cover all five escape kinds, with computed data (`breakCoeffs`), correctness (`breakCoeffs_relation`), and nontriviality (power-of-two domain coefficient / indicator). This is Theorem 5.4.4 made computational, and it de-terminalizes the security map's SDLR node — we'll restructure the map when this merges (the red retreats to the independence-of-bases edge and the DLP terminal).
- `Pool.lean` / `KeyBinding/Pool.lean`: the deployed instantiation matches what the games consume — `valueBound = 2^64`, `uncommitted = 2` with Theorem 5.4.6 discharged concretely (13 a non-residue), and `break_of_akP_ne` discharged from the break projection via the extract ±-property. All three coordination flags from my earlier direction check (Merkle-interface reconciliation, `break_of_akP_ne`, `Primitives` extensions) are resolved in this revision.
- The value bounds on `NoteCommitBreak` are a genuine games-layer soundness fix: the concrete commitment consumes `v` through a 64-bit encoding, so without the bounds the break structure is inhabitable by value aliasing (`v` vs `v + 2^64`). We'll take the systematic version on the games side (a value role type `VAL` with a bounded conversion to ℕ, letting these per-structure bounds drop again) as a follow-up after this merges.

## Findings

1. **The blinding addition is complete in the spec too — the `Pool.lean` docstrings have the divergence backwards.** (Duplicates the posted blocking threads; recorded so this review is self-contained.) The `noteCommit` and `keyBinding` docstrings describe the final blinding addition as deployed-gadget-faithful complete addition *diverging* from a spec-text incomplete `⸭`, with an exceptional-case caveat for a future spec-equivalence lemma. §5.4.8.4 composes the blinding with complete `+`, as does `halo2_gadgets` — so the model agrees with both, and the claimed divergence and its caveat should be deleted. The point matters for soundness: incomplete addition there would let an adversary choose `rcm`/`rivk` to force an exceptional case.
2. **PR description is stale in two places.** "Remaining follow-up, out of scope here: the deterministic conversion from a classified `BreakData` to the games-facing Sinsemilla-DLR break object" — that conversion is now in-branch (`classifyRelation` / `relationOfBreakData`). The remaining deferred piece on the key-binding side is the `Pool.Break` → relation reduction (its docstring says so correctly). The "net −76 lines" figure also predates the later commits. Please refresh the body.
3. **Namespace base-name reuse (question, not a demand).** `KeyBinding.Pool.{Witness, KB, Break, toInterface}` mirror `KeyBinding.{Witness, KB, Break, toInterface}` within the same namespace tree. The repo convention avoids reusing a base name for a different declaration in one namespace tree (resolution is confusing even when the compiler copes). Is the parallelism deliberate here? If so it may be worth a comment saying the Pool names intentionally mirror the Recovery adapter's; if not, distinct base names would avoid the ambiguity.
4. **Composition with the capstones (informational, no ask).** The games' capstone layer (#108) is generic over the key-binding interface, so the pre-quantum instantiation plugs in `Pool.keyBinding`/`Pool.primitives` directly; the key-binding arm's ε then discharges deterministically through the deferred `Pool.Break` → relation reduction into the same DL terminal as the Merkle and note-commitment arms (no ROM needed pre-quantum). We'll take that reduction on the games side after this merges; `ofPoint_hashToPoint`/`breakCoeffs` supply most of the ingredients. Mechanically, whichever of #98/#108 merges second reconciles `Pool.primitives` with #108's `Primitives` extensions (`bindingVerify` and the binding-signature fields).

Zcash/Security/Ledger/Pool.lean
gio54321: My understanding is that these were stale comments, and the lean function correctly uses complete operations.
Fixed both comments

Zcash/Security/Ledger/Pool.lean
gio54321: Added references for the constants in the doc strings

TalDerei:
Re-checked every thread of mine and @daira's against the current head (`1aab8205`). **All code-level findings are addressed.** Per-item status below, then the residuals that aren't Lean changes.

## My threads

- **Stale `ActionData` docs** (`Bundle.lean`) — fixed in 27cb98c8; now "the ten public-input rows (zcash/orchard#504)".
- **`BridgeTests` not in CI** — `Security.lean:11` imports it, so the assertions actually execute.
- **Trust-policy hole** — `TrustBoundary.lean` now carries `assert_axioms … +native` on `soundnessPost`, `specPost_to_ledger`, `circuit_soundness_to_ledger`, and `actionBreak_iff_classify_isSome`.
- **Stale `ActionSatisfied` TODO and the two missing reductions** — TODO replaced. `ActionBreakData` → games-facing DLR is `relationOfBreakData` / `classifyRelation`; conflicting valid paths → collision is `collisionOfWrongLeaf`, rebuilt over `MerklePrimitives` and returning `Σ i, DefinedCollision`, so an escaped compression can never masquerade as a collision. Both are plain compiled `def`s under `assert_computable +choice` — which also answers @gio54321's point about wanting a function rather than a theorem for PPT inspection.
- **The two TODO requests** (`Bridge.lean:392`, `Merkle.lean:208`) — superseded: both became implementations rather than TODOs.
- **`ActionInstance` and the bundle-level flags** — settled by @daira's ruling that they belong in the game instance. The omission is documented at `Statement.lean:61-66`, and the TODO commits to growing the fields when the games land.
- **`NoteCommitBreak` value aliasing** — `v₁_lt` / `v₂_lt` now travel with the structure (3824c422), so the onward reduction can be written against the break object alone rather than recovering the bound at the call site.
- **Missing consumer-side glue** — `actionBreak_iff_classify_isSome` exported at `Bridge.lean:487` and asserted in the trust boundary.
- **`native_decide` nit** in `signedMagnitude_eq_int_sub` — gone; no `native_decide` left in `Bridge.lean`.
- `EnvAssumptions` remains, deferred to #99 as agreed in that thread.

## @daira's threads

- **Both blocking complete-addition items** (`Pool.lean`, `noteCommit` and `keyBinding`) — fixed in 121d15ae + 822e4b9b. The docstrings now state that the blinding addition is complete, matching both §5.4.8.4 (complete since spec version 2021.2.16) and `halo2_gadgets`' `CommitDomain::commit`, and give the soundness reason: an incomplete addition would let a prover choose `rcm` / `rivk` to force an exceptional case. The claimed divergence and its negligible-escape caveat are deleted.
- **Constant provenance** — the three domain points now cite §5.4.1.9 and name their `zcash/orchard` `src/constants/sinsemilla.rs` identifiers; `uncommitted` already cited §4.2.3 and Theorem 5.4.6. One gap left: `valueBound := 2 ^ 64` (`Pool.lean:213`) has no citation.
- **`assert_no_sorry`** — all 29 removed along with the `Mathlib.Util.AssertNoSorry` import; only `assert_axioms` / `assert_computable` remain, both of which imply it.
- **`SinsemillaDLR` docstring** — suggestion applied verbatim.
- **NU6.3 flags in the game instance** — the rationale is now the TODO text itself, including why the adversary must be able to submit transactions on both sides of the activation.

## Residuals

1. ~~**PR description** (@daira's finding 2) still describes the `BreakData` → Sinsemilla-DLR conversion as out-of-scope follow-up — it's in-branch now — and still quotes "net −76 lines". Both predate the later commits.~~
2. **Namespace base-name reuse** (finding 3) — `KeyBinding.Pool.{Witness, KB, Break, toInterface}` are unchanged and the question is unanswered; either a comment recording that the mirroring of the Recovery adapter is deliberate, or distinct base names.
3. @daira's review is still `CHANGES_REQUESTED` — the blocking content landed in 822e4b9b, so it needs a re-review to unblock the merge.
4. The Lean build is `pending` at this head; worth confirming green before merge, since this is the first run over the merged `main`.

---
Review assisted by Claude Opus 5.

daira:
Post-merge re-check, limited to what the thread above does not already cover:

- Both trust guards pass on the merged state: the csimp census (1 `@[csimp]` declaration, covered) and the native opt-in check (1 module in the lane import cone, 0 opted in, no ambient native-executing checks). This includes the merge's `defaultTargets` union (`CircuitVkCheck` + `Arithmetic`), whose `lakefile.toml` conflict was the merge's only one.
- The `assert_no_sorry` removal lost no census coverage: checked name-by-name, every removed name still has an `assert_axioms` / `assert_computable` entry in `BridgeTests.lean` or `TrustBoundary.lean`.
- Finding 4 (composition with the capstones) is closed: #108's rebase reconciled `Pool.primitives` with the `Primitives.bindingVerify` extension — the binding predicate is a second parameter alongside spend-auth verification, so the bridge statements stay scheme-generic.

PR #96, opened 2026-07-23
feat(fv): wire the top-level deployed constraint-soundness capstone

TalDerei:
**Human Summary:**

Closes #51 and #94. Addresses #12, #33, #67, and #117; their remaining original-scope acceptance criteria are explicitly transferred to #115 and are not claimed complete here. 

Wires in the top-level constraint soundness capstone (`orchard_deployed_knowledge_error_captured`) and computes the explicit deployed knowledge-error bound from the adversary’s query bound Q, parameterized by the assumed DLOG advantage.

-------------

**AI Summary:**

## Summary

This PR rebuilds the deployed capstone on the rewind-free AGM route from #106/#112. The obsolete rewinding, continuation-threshold, fourth-root, and offline-interpolation paths are removed.

At the captured Post-NU6.3 key, `orchard_deployed_knowledge_error_captured` proves

```
(Q+k)·3/|𝔽| + (Q+1)·1/|𝔽| + (ε_DL + 1/|𝔽|) + (Q+23)·algebraicRootBudget + (Q+1)·20470/|𝔽|.
```

The PR supplies the captured degree walk and constants, the `x`-squeeze schedule, the interpolation-free direct-coordinate family, and the consensus-max specialization. `snarkConstraintsDeployed_prob_le_at_consensus_max` applies the composite bound at `numProofs = 2¹⁶ − 1`, `k = 11`, and `Q ≤ T`; for `T = 2¹²²`, the multiopen contribution is at most `2⁻⁸⁶`.

## Causality and computability

Root and constraint traces are actual `OracleComp` bad-set computations with agreement and fresh-query proofs. Exact reprogramming invariance is derived from freshness; the captured schedule and capstones no longer accept bare pinning premises.

All relations charged to DLOG are produced by executable, census-pinned finders. The choice-based outcome provider and offline family constructors are gone; root existence is eliminated only inside `Prop`. The direct outcome and family constructors are ordinary `def`s guarded by `assert_computable`. A computable degenerate root-trace witness establishes consistency of the causal interface; concrete captured-shape traces and runtime accounting remain the explicit #115 handoff below.

The theorem here closes the compressed verifier-identity budget. Row-level semantic promotion and polynomial total-cost accounting are not claimed by this PR.

## Remaining #115 handoff

The immutable Rust key fixture provides verifier metadata, not a prover/oracle trace, and is unchanged. Three substantive tasks remain:

[...]

PR #99, opened 2026-07-24
Integrate circuit soundness with the verifier soundness stack

### LINKED ISSUE #123: The `Zcash` package should not depend on native code
Follow-up work *eventually* that's synergetic with https://github.com/zcash/ironwood/issues/110, cc @daira.

Executing a check through a locally compiled `precompileModules` dylib is a
compiler-trust event with **no axiom trace**, and dylib loading is inseparable from
elaborating any module whose import closure reaches the lane. `assert_axioms` cannot
see it, which is why `scripts/check_native_optin.py` has to enforce it textually at
the module level and hand-mirror the CompElliptic pin's lane (`PIN_LANE`).

The durable fix is packaging, not policing: **no package whose theorems we trust
should be able to reach a native lane at all.**

## Plan

- Split CompElliptic's native lane into a separate `CompEllipticNative` package.
- Move ironwood's own lane (`Zcash.FastFieldNative`, `lakefile.toml:91`) and the
  fixture checks into a package downstream of `Zcash`, which may depend on
  `CompEllipticNative`.
- `Zcash`'s generic theorems take fixture facts as **hypotheses**; the downstream
  package discharges them and instantiates the capstone. This promotes the existing
  `Snark/Soundness/` vs `Snark/Soundness/Deployed/` directory boundary to a package
  boundary.
- Retire `PIN_LANE` and, for `Zcash`, most of the native-opt-in guard: the invariant
  becomes structural.

## Notes

- **Not a duplicate of #110.** That issue is `native_decide` → `decide`/structural —
  axiom-traced and censused. This is the dylib axis, which leaves no trace. Both are
  worth doing; this one recovers more assurance per unit of effort.
- Facts like `Fixture.fingerprint_matches` stay `native_decide` — a 2048-scalar Vesta
  MSM comparison has no kernel-checkable alternative. The goal is to keep them
  *outside* the trusted package, not to prove them.
- Cost: dropping `precompileModules` requires reverting the Montgomery representation
  to `Nat` (limb arithmetic is slow interpreted), taking the VK certificate from <1 min
  to ~8–9 min. That lands on CI's default targets only — `lake build Zcash` does not
  build it. It also deletes the `msmM_spec` / `fftS_spec` simulation machinery that
  exists solely to justify the fast path.
- Currently three modules on `main` reach a lane: `Zcash.Arithmetic.ScalarFftEquiv`,
  `Zcash.FastFieldNative`, `Zcash.Vendor.CompPoly.ScalarFftDefs`. #99 adds
  `Snark/Keygen/Certificate.lean`.

mitschabaude:
**Human summary**. The PR that finally gives `hencodes` at the capstone!

image

## AI Summary

This PR completes the bridge from the verifier soundness stack to the proved Orchard Action circuit statement. It builds on the verifier-native full constraint model merged in [zcash/ironwood#91](https://github.com/zcash/ironwood/pull/91): the circuit now owns the keygen data and correctness package needed to turn canonical verifier constraint satisfaction into the circuit specification, without a free `S`, `hencodes`, decoder, or separately supplied verifying key.

## What this PR adds

- A closed `TopLevelCircuit` boundary with unit input, `True` top-level assumptions, hint-backed prover witnesses, and circuit-owned synthesis/keygen coherence.
- Circuit-derived pinned constraint-system data: operation streams, V1 floor-planner placement, selector activations and compression, query layouts, domain parameters, proof shape, and verifying key.
- A shared verifier gate-expression representation and semantics-preserving adapters from verifier polynomial satisfaction to Clean gate, copy/permutation, lookup, fixed-column, and instance semantics.
- Structured Action public inputs and the bridge from verifier-committed instance rows to the circuit's public statement.
- A circuit-generic Vesta capstone for arbitrary `TopLevelCircuit`s and arbitrary proof multiplicity, followed by a thin Action specialization yielding `BundleStatement` or a nontrivial relation in the supplied URS.
- A concrete deployed-Action specialization whose captured verifying key is certified equal to the circuit-derived key.
- Architecture and follow-up documentation for replacing the remaining interim whole-circuit certificates with generic compiler and local lawfulness arguments.

## Architecture

The public soundness seam remains Ironwood-native: verifier soundness establishes satisfaction of the Ironwood constraint model, while the circuit-owned correctness interface turns that fact into a circuit statement. Clean-facing implementation details are concentrated under `Zcash/Circuits/Integration`; the generic terminal soundness theorems live under `Zcash/Snark/Soundness`, and fixture-dependent deployed specializations live under `Zcash/Snark/Soundness/Deployed`.

The main generic theorem is `topLevelBundleStatement_or_relation_of_vestaTerminal`. `actionBundleStatement_or_relation_of_vestaTerminal` specializes it to the Action circuit without depending on captured fixtures or assuming `numProofs = 1`; the deployed capstone separately connects the circuit-derived key to the canonical captured Action key.


TalDerei:
# Audit: spec bugs, implementation correctness, soundness holes

**TL;DR: no critical or major soundness holes found.** Seven parallel deep-audit passes over the ~43k-line diff (native kernels; keygen/VK certificate; canonical constraint model; permutation/lookup arguments; selector/fixed/gate boundary; copy/lookup/instance boundary; Action statement + capstone glue) all came back clean at the critical/major tier. The headline claim — `actionBundleStatement_or_relation_of_vestaTerminalDerived` has no free `S`/`hencodes`/decoder and binds the verifier-committed public inputs to `BundleStatement` — holds up under adversarial review. What remains are minor hardening items, documented-interim trust surfaces, and one process flag.

On the existing review comment (generic-`TopLevelCircuit` factoring): the audit independently confirmed its one technical premise — `acceptedAdviceSelection_feed_eq`/`acceptedInstanceSelection_feed_eq` are genuinely singleton-bundle-only (`∀ other, other = proofIndex` at `Multiopen/CanonicalSelection.lean:210,450`), so the generic capstone will need per-proof accepted-member selection as described.

## Process / trust-surface findings

1. **Zero CI runs on the PR head.** The head sits on a fork and has no check-runs at all — nothing upstream has machine-verified that this branch elaborates. To close that gap locally: a full `lake build --wfail` of head `b7130fb` against the pinned Clean rev `ab068244` **builds clean** — 8902 jobs, no warnings, no errors. That covers all four default targets, including the ~45 s keygen `Certificate`, `SingleAction.VkMatch`, the interim Action certificates (`ActionPermutationDomainCompute` 65 s, `ActionFixedCoherenceCompute` 30 s, `ActionCopyWitness` 28 s), `Deployed.ActionVesta`, and the `TrustBoundary` axiom census — so the axiom-tier assertions, including the new `+pallas` tier, hold as written. Worth wiring up CI on the fork branch regardless, since this is a one-off local check rather than a standing gate.

2. **The Clean dependency moved to a same-day coordinated feature branch.** `lake-manifest.json` pins Clean to `ab068244` ("Guard V1 lookup selector placement", on `halo2-clean-2`, not Clean's main). The ground-truth semantics this PR bridges into — `Halo2.Constraints`, the V1 floor planner including the new lookup-selector placement guard, `closeWithOperations` — changed in tandem with this PR but outside this repo's review flow. Those Clean-side changes deserve the same audit treatment as this diff, and the pin should eventually land on a protected ref.

3. **Clean's Rust-fidelity is the unpinned foundation.** The fixtures pin VK *artifacts* byte-for-byte, but the semantics the proofs bottom out in (selector `selReplacement` vs `compress_selectors.rs`, rotation sign/wrap, `blindingFactors = max(3, maxAdviceQueries)+2`, table default-fill) is an assumption. The audit spot-verified every sensitive point against halo2 0.3.x and found exact matches — but this deserves a standing note in the trust docs.

## Minor findings (actionable)

4. **Off-curve fixture points silently become the identity** — `Zcash/Snark/Fixtures/SingleAction/Fixture.lean:35`. `mkVestaPoint` falls back to `0` for off-curve coordinates and out-of-range reads default to identity. The 44 commitments and the Lagrange prefix self-correct via the certificate, but `capturedURS.u`, `capturedURS.w`, and the 2048 URS generators are only checked *in aggregate* — and the capstone's escape hatch is stated about exactly those elements (`HasNontrivialRelation capturedURS.g/u/w`). Cheap hardening: one bundled `native_decide` asserting every captured coordinate pair is on-curve and the list lengths match. This is the highest-value small fix in the PR.

5. **`VkMatch.lean:87` compares five pinned-CS fields against themselves.** `capturedPinnedView` is a record-update of the *derived* `actionPinnedCs`, so `numFixedColumns`, `numAdviceColumns`, `numSelectors`, `constants`, `minimumDegree` aren't cross-checked against the capture. The docstring's indirect-coverage argument is sound (each leg was verified — wrong values would break the commitment certificate or the Shape equality), but it lives in prose; these fields should be promoted into the comparison once the exporter emits them.

6. **The pre-compression CS cross-check was deleted** (`actionPre.json`, `TestVkMatchAction.lean:29`). Post-compression coverage improved (derived selector map and placements now checked against the Rust dumps), and only the post-compression CS enters the VK — but one independent check of raw configure output is gone. A configure bug canceling against a compression-port bug is no longer independently excluded.

7. **Two fail-closed-but-opaque encodings worth doc comments:** overlapping selector activations in one packed column would surface as an inscrutable realization-certificate failure rather than a structural error (`FixedColumns.lean:198` + Clean's `selectorFixed`); and the malformed-lookup-selector sentinel (out-of-bounds column at `FixedColumns.lean:91-119`) is load-bearing — widening the column-bound check later would silently pass malformed entries as 0.

## Verified-sound observations (the things most likely to be wrong, and aren't)

- **The VK derivation seam is real**: `vk_eq_derived` is a full 14-field record equality via `VerifyingKey.mk.injEq`; the `native_decide` comparison uses structural `DecidableEq` over affine coordinates (unique representation — no Montgomery/projective aliasing, no false positives). Keygen matches halo2's `build_vk`: blind `+1·w` = `Blind::default()`, δ-power σ columns, and a kernel-checked DFT spec (`bestFftG_dft`) that rules out ω-direction/bit-reversal double-swaps.
- **Verifier acceptance semantics untouched**: the +672-line `Assemble.lean` diff is purely additive provenance lemmas (checked hunk-by-hunk); `Expr` moved to `Common/` byte-identically.
- **Protocol transcription matches halo2 0.3.x exactly**: permutation chunk expressions (δ-coset starts, `Rotation::last` chaining, `l₀/l_last/l_blind` split), all five lookup constraints with correct rotations and θ-fold order, and the y-fold ordering the deployed verifier executes. Row accounting (`n − bf − 1` active rows, `n − bf − 2` last index) is consistent between semantics and pricing.
- **Extraction has no vacuous or adversary-dischargeable premises**: every capstone premise was checked — `hprob1` is the standard knowledge-error floor; route/binding premises carry the relation escape; challenge exclusions are finite sets with *proven* cardinality bounds, schedule-correct against the transcript order (γ bad sets may depend on β; β bad sets only on pre-β data). A malicious `hpoly`/`colPoly`/`i`/`m` falsifies a premise rather than the conclusion.
- **Instance binding is closed end-to-end**: `deployedInstanceCommitment` commits exactly the ten zero-padded rows with blind 1 (halo2's verifier-side `commit_lagrange`), binding to the decoded polynomial goes through the DL-relation escape (never assumed), and the environment satisfying `top.Statement` is the *same* one whose instance column is pinned to `inputs`. Row order 0–9 matches Orchard, and `SpecPost` is a faithful transcription (flags correctly non-boolean; anchor correctly unbound when `vOld = 0`).
- **σ-replay is proven, not trusted**: the executable halo2 `Assembly::copy` union-find replay is proven equivalent to the abstract cycle merge by structural simulation; declared Clean copies are proven (by induction) to land in the V1 copy list.
- **Hygiene**: no `sorry`/`admit`, no new `axiom`, no `@[extern]`/`@[implemented_by]`; the single `@[csimp]` is kernel-proved with no `native_decide` in its chain; the `+pallas` tier addition is correctly scoped to the pre-existing, documented point-count axiom.

## Known-debt confirmations (documented in-tree; no new action)

- The Action fixed/selector layer rests on ~5 whole-circuit `native_decide` certificates ("prominently interim", tracked as 26 obligations in `lawfulness-and-certificate-elimination.md`). The Props are stated over the genuine circuit objects — no projection tricks — so this is `+native` compiler trust, as declared.
- `closeWithOperations` repairing configure/synthesis mismatch is a modeling-fidelity/completeness concern, not a soundness hole for *this* capstone: closure only adds constraints, and an active repair would show up in the certified pinned fields — consistent with the in-tree doc's own analysis.
- The multiopen route premises and challenge exclusions are deliberately deferred to the probability-layer composition (#96); the deterministic completion criterion ("no live `hencodes`") is genuinely met.
- Small robustness notes for doc comments: unrouted `CommitmentId`s resolve to the zero polynomial (`ConstraintResolver.lean:71-91`); out-of-range permutation query refs total to column 0 behind the (discharged) routing-coherence premise; non-primary instance columns commit to `0` where halo2 would give `1·W` (dead for the deployed single-column shape); row 9 (`DISABLE_CROSS_ADDRESS`) ordering is externally anchored only by the unreleased orchard 0.15.3 capture pin — worth an explicit test tying `PublicInputs.rows` order to the captured instance dump.
- PR-body staleness: it still describes interim fork pins for fixture regeneration, but the workflow now pins the orchard 0.15.3 release SHA with halo2_proofs 0.3.4 from crates.io ("No fork pins") — the body should be updated. The tag→SHA pin change itself is an improvement.

---
*Method: the diff was audited against merge-base `6bdba1ee` in seven parallel passes, each reading its cluster's definitions and theorem statements in full (proofs skimmed — Lean checked those), cross-checking protocol shapes against halo2_proofs 0.3.x and orchard, and adversarially asking for each certificate/theorem "what wrong world would still pass this?".*

*This audit was generated with the help of Claude Fable 5, via [Claude Code](https://claude.com/claude-code). Treat the findings as leads to verify against the code, not as authority — the file:line references and the build result are checkable, and the "verified clean" claims are the ones most worth a second human look.*

TalDerei:
organizational request: this touches the existing `Snark/` files, and i think changes like that (across `Snark/` and the other folders) should be organized into cleaner, more idiomatic groupings.

mitschabaude:
status update:
- I'm still reviewing/polishing myself, but given the tight timeline I'm happy to fold in your feedback in parallel
- I've been fixing a numProofs = 1 assumption smuggled into the main capstone theorem
- I'll split out the large body of work around fast curve operations (montgomery field, projective coords, fast msm). it was necessary to make checking the vk match in CI realistic, but is cleaner to review in isolation. planning to upstream most of the code to compelliptic / comppoly
- possibly the work towards `TopLevelCircuit` and `.toVerifyingKey` will become its own PR as well
- want to improve file structure

> changes like that (across Snark/ and the other folders) should be organized into cleaner, more idiomatic groupings.

fully agree and that's also a thing I want to polish a lot. (codex tends to produce quite a mess, code-quality wise)

TalDerei:
> status update:

SGTM!

TalDerei:
> * I've been fixing a numProofs = 1 assumption smuggled into the main capstone theorem

we should generalize this ideally, reference our multi-action fixture and `parametric.lean` (which defines `FixtureMax.shape n` for all consensus-valid n).

mitschabaude:
another thing that should be fixed, hinted at in the [AI slop comment above](https://github.com/zcash/ironwood/pull/99#issuecomment-5078776953):
the current main capstone is stated about the action circuit specifically, and the proof is an ugly hodgepodge of facts proved for all circuits, and facts just proved for the action circuit. this should be cleaned up and there should be an interface that gives you basically the same theorem _for any circuit_. I might land some of that cleanup here but don't see it as a hard blocker for the PR to land.

Another, partially related concern, that is fully reserved to a follow-up: Getting rid of all the `native_decide` checks for various CS correctness facts. These should be proven generically from the CS compiler (`.toPinnedCS` / `.toVerifyingKey`) or locally at the Clean circuit interfaces. Reason: `native_decide` increases trust surface significantly, and also this is uglifying any theorem that holds for arbitrary top-level Clean circuits, with extra side conditions (that we happen to have proved for the action circuit, mostly via `native_decide`).

mitschabaude:
quick answers here:

> **Process / trust-surface findings**

> * **Zero CI runs on the PR head.**

that's incorrect, CI has run many times here, because I as a co-maintainer can approve it

> * **The Clean dependency moved to a same-day coordinated feature branch.** `lake-manifest.json` pins Clean to `ab068244` ("Guard V1 lookup selector placement", on `halo2-clean-2`, not Clean's main). The ground-truth semantics this PR bridges into — `Halo2.Constraints`, the V1 floor planner including the new lookup-selector placement guard, `closeWithOperations` — changed in tandem with this PR but outside this repo's review flow. Those Clean-side changes deserve the same audit treatment as this diff, and the pin should eventually land on a protected ref.
> * **Clean's Rust-fidelity is the unpinned foundation.** The fixtures pin VK _artifacts_ byte-for-byte, but the semantics the proofs bottom out in (selector `selReplacement` vs `compress_selectors.rs`, rotation sign/wrap, `blindingFactors = max(3, maxAdviceQueries)+2`, table default-fill) is an assumption. The audit spot-verified every sensitive point against halo2 0.3.x and found exact matches — but this deserves a standing note in the trust docs.

I think we can treat Clean as an _untrusted_ dependency in this repo. The main capstone proved here doesn't mention anything about Clean, it states soundness purely in terms of concepts native to the ironwood code base. Clean was just a helper device getting there. So it's somewhat irrelevant to this code base if, e.g., Clean's CS generation matches what Rust does, or if core Clean statements like `Halo2.Constraints` have been shifting

TalDerei (approved):
based on my previous review, ACK after #98 lands!

Zcash/Circuits/Action/Bundle.lean
```diff
@@ -273,30 +324,27 @@ def SpecBase (G : Generators) (B : Bases) (wit : ActionData) : Prop :=
...
-  -- diversified-address integrity: `ivk ∈ {Commit^ivk, ⊥}` (break exhibited) and
+  -- diversified-address integrity: `ivk ∈ {Commit^ivk, ⊥}` (guarded ⊥-model) and
   -- `pk_d_old = [ivk] g_d_old`
   (∃ ivk : Fp,
-    SpecOrBreak G.S B.ivkQ
-      (fun bp => ivk = (bp + wit.rivk.2 • B.commitIvkR).x)
-      (hashToPointB G.S B.ivkQ (commitIvkChunks wit.akP.x.val wit.nk.val)) ∧
+    HashGuarded G.S B.ivkQ (commitIvkChunks wit.akP.x.val wit.nk.val)
+      (fun bp => ivk = (bp + wit.rivk.2 • B.commitIvkR).x) ∧
     wit.pkdOld = ivk.val • wit.gdOld) ∧
```
mitschabaude, line 332: why does this use an existential? why not substitute `ivk` with `(bp + wit.rivk.2 • B.commitIvkR).x`

Zcash/Circuits/Action/Bundle.lean
```diff
@@ -273,30 +324,27 @@ def SpecBase (G : Generators) (B : Bases) (wit : ActionData) : Prop :=
...
+  -- Merkle path validity, tied through the `q_orchard` anchor check: the exact
+  -- raw-encoding chain of the witnessed path cells (zcash/ironwood#97), each layer's
+  -- hash in the guarded ⊥-model.
   (∃ root : Fp,
-    Sinsemilla.Merkle.MerkleRoot G B.merkleQ 0 wit.cmOld.x 32 root ∧
-    (Sinsemilla.Merkle.MerkleRootStrict G B.merkleQ 0 wit.cmOld.x 32 root ∨
-      Sinsemilla.Merkle.MerkleBreakAt G B.merkleQ 0 32) ∧
+    Sinsemilla.Merkle.ExactMerklePathData G B.merkleQ 0 32 wit.cmOld.x root
+      (merkleLeftEncoding wit) (merkleRightEncoding wit) (merkleSide wit) ∧
     wit.vOld * (root - wit.anchor) = 0) ∧
```
mitschabaude, line 348: why is `root` not part of the extracted `ActionData`?

Zcash/Circuits/Action/Bundle.lean
```diff
@@ -1797,59 +1870,55 @@ private theorem aafi_output (instCol : Column .instance) (r : ℕ)
...
   output _ _ := ()
   regionCount _ := 395
   output_eq := by intro _ _; rfl
-  regionCount_eq := fun W i => (mainPost_regionCount G B cfg W i).symm
+  regionCount_eq := fun input i => (mainPost_regionCount G B cfg input i).symm
+
+/-- Read the Action statement data for the fixed top-level witness program. -/
+def extractPost (cfg : Config) (_ : Var unit Fp) (i : RegionIndex)
+    (env : Placed Environment Fp) : ActionData :=
+  extractBase cfg () i env
```
mitschabaude, line 1953: why are we adding this no-op wrapper? please don't. in fact, both bundles could just use `extract` directly?

Zcash/Circuits/Action/TopLevel.lean
```diff
@@ -0,0 +1,251 @@
...
+  rfl
+
+/--
+The proof-carrying Orchard Action circuit at arbitrary certified constants, closed as
+a deployable `TopLevelCircuit`: unit public synthesis input, `True` verifier
+assumptions, and no unfulfilled environment contract at its boundary.
+-/
+def topLevelCircuit (G : Generators) (B : Bases) :
+    TopLevelCircuit Fp Unit Config unit where
+  formalCircuit := circuit G B
```
mitschabaude: IMO this shouldn't be generic over `G` and `B`, just plug them in, and remove the noisy separate`orchardActionTopLevelCircuit`

TalDerei (approved):
reACK after fixing merge conflicts!

TalDerei:
and after addressing @daira's feedback OOB: "there's a design rule that I think should be enforced about the native compilation. CompElliptic#13 didn't enable that by default and therefore ironwood#111 and ironwood#99 won't use it, but I looked at what you need to check in order to conclude that, and it's pretty horrid. Fixing that now so that use of native compilation is always an explicit (and clearly documented) opt-in in both CompElliptic and ironwood" per https://github.com/zcash/ironwood/pull/120.

TalDerei:
# Re-audit at `744fe235`: spec bugs, impl correctness, soundness holes

**TL;DR: no critical or high soundness holes.** Ten parallel deep-audit passes over the diff (gate adapter; copy/permutation adapter; lookup adapter; fixed/selector coherence; instance & public-input bridge; canonical verifier layer; keygen/VK certificate + build targets; synthesis laws & gadget diffs; new FFT/Lagrange arithmetic; Clean-boundary + docs), cross-checked against each other, with the load-bearing claims re-verified mechanically.

The adapter chain from verifier polynomial satisfaction to `actionCircuit.Statement` is semantically tight everywhere we could attack it. The real findings are trust-surface and wiring issues. The top one is a merge-resurrected dead file that makes 562 lines of the diff look verified when Lean never elaborates them.

Scope note: this is the head **after** the `agent/top-level-public-input` merge, so it is newer than my previous audit. Among other things it **fixes** the `numProofs = 1` issue you flagged: pre-PR, `orchard_verifier_vesta_member_constraints_terminal[_derived]` fed one shared advice/instance member selection to *every* proof index (`fun _ => rotatedFeed …`), sound only at `numProofs = 1`. Selections are now per-proof and derived from the accepted route via `acceptedAdviceSelection_feed_eq` / `acceptedInstanceSelection_feed_eq`, which makes the bundle-wide statement meaningful for arbitrary `shape.numProofs`.

---

## Findings

### 1. `Zcash/Snark/Soundness/Deployed/ActionVk.lean` is a merge zombie: 562 lines in no build target, never elaborated

**medium** · hygiene / false verification coverage · high confidence

`91413448` ("Delete the deployed lane of the Action capstone tower") deliberately deleted this file, folding the deployed capstone into `Deployed/ActionVesta.lean` as a `subst`-based transport. But a same-branch merge (`bb031613`) reconciling that deletion against parallel commits that had *modified* the file resurrected it, and the follow-up repair commit (`09867a83`, "Repair the deployed capstone merge") rebuilt only `ActionVesta` and missed it. The base-side parent `4b76b56e` still carried its own older copy (and an `ActionVesta` that imported it), so nothing conflicted at the final merge `744fe235`, which kept the fork side's copy alongside its import-free `ActionVesta`.

At head:

- `grep -rn "Deployed.ActionVk"` returns **zero importers**.
- No glob covers it: `FixtureCheck` names `Zcash.Snark.Soundness.Deployed.ActionVesta` as a singleton, not `Deployed.+`, and the `Zcash` lib has no `globs` (root-module closure only). Two independent import-closure computations over all modules agree it is the only orphan besides the pre-existing `Zcash.FastFieldNative` root.
- Its two `assert_no_sorry` lines (`:251`, `:560`) therefore never run. It could contain a `sorry` or a type error and CI stays green.
- It also re-declares `Zcash.Snark.shape_k_eq_capturedURS_k` (`:68`) and `Zcash.Snark.vk_blindingFactors_lt` (`:73`) — the same fully-qualified names as `Deployed/ActionVesta.lean` `:40`/`:45` — so widening the glob is not the fix; the two modules can never be co-imported.

No soundness impact: its role is superseded by the compiled, checked transport. But a reviewer reading the diff sees 562 lines of "deployed terminal" Lean has never checked. **Fix: re-apply the deletion.**

### 2. The deployed lane's axiom budget is pinned nowhere, and `TrustBoundary.lean:735` does not do what it appears to

**medium** · trust-boundary coverage gap · high confidence

`assert_axioms action_bundleStatement_or_relation_of_deployedAccepts +native` resolves to the **generic** theorem (`Zcash.Snark.action_bundleStatement_or_relation_of_deployedAccepts`, from `Soundness/ActionVesta.lean`, imported at `TrustBoundary.lean:33`). `TrustBoundary` does not import `Soundness.Deployed.ActionVesta` at all, and the deployed homonym lives at `Zcash.Snark.Deployed.…`.

So the deployed capstone and the whole keygen-certificate chain (`Keygen.certificate`, `vk_eq_derived`, `shape_eq_mergeDerived`) carry at most `assert_no_sorry` — `shape_eq_mergeDerived` carries no assertion at all — which catches `sorryAx` but is blind to a newly introduced `axiom`, or to an unexpected `native_decide` entering via `Keygen/Certificate → Arithmetic/InvDftMont → ScalarFftEquiv → Vendor/CompPoly` or via the Clean pin. Every other empirical lane (`Fixtures/SingleAction`, `Fixtures/MultiAction`) gets `assert_axioms … +native`; the deployed capstone — this PR's headline — is the exception.

Trap when fixing: adding the import will **not** repoint line 735, since `open Zcash.Snark` does not bring `Deployed.foo` into scope as bare `foo`. The assertion needs the qualified name.

### 3. The deployed capstone's instance commitment is never certified against the captured `g_lagrange` — and both halves of the bridge are already proved

**medium** · model-fidelity gap · medium-high confidence (found independently by two passes)

The capstone's acceptance hypothesis uses `ActionInstanceCommitment.commitment`, whose `instanceKey` generators are *defined* as `commit urs (polynomialCoefficients (2^urs.k) (rowPolynomial omega (Pi.single i 1)))` with `generator_eq := rfl`. That object is mathematically right — the definition audits clean against halo2's `commit_lagrange`, including `Blind::default() = 1` and zero-padding of rows `10..n-1`.

But the deployed verifier commits against the captured `g_lagrange` table (`Fixture.capturedUrsGLagrange`), and nothing compiled connects the two:

- `derivedUrsGLagrange_prefix_eq` (`Keygen/Certificate.lean:221`) proves the captured prefix equals the derived table — and has **no consumer anywhere in the repo**.
- `derivedUrsGLagrange_generator_eq` (`Keygen/Lagrange.lean:186`) + `polynomialCoefficients_single_closed` give the other half.
- Composing them is a short lemma that is not written. Consequence: the 10-generator Lagrange-prefix component of the `native_decide` bundle (`Certificate.lean:139`) is currently costed but dead weight.

Failure mode: if the capture's `g_lagrange` were inconsistent with `(g, ω, k)`, the capstone's antecedent would not be the real verifier's acceptance condition, and no build check would notice — `instance_commitments_derived` is about a different basis object. Every *verifying-key* field does go through the proven path (`committer_eq` → `commitInvDftMontWith_eq`, not the raw fast path), so this is specifically the instance-commitment seam.

### 4. Book overclaims `closeWithOperations` inertness

**medium** · doc-overclaim

`book/src/formal-verification/action-circuit-integration.md:736` states the keygen closure is "inert, as checked by the existing VK/layout fixtures". The PR's own `lawfulness-and-certificate-elimination.md:56-61` correctly explains why that check is insufficient: the comparison says the *repaired* derivation matches the deployed data, not that the repair was inactive, and projection-equivalent repairs need not be observable in any pinned downstream field (open obligations 12a/12b/26). No `rawCS.closeWithOperations ops = rawCS` theorem exists for Action; `TopLevelCircuit.keygenCoherent` is true by construction of the repaired object. Deployed-VK equality bounds the blast radius, so this is modeling fidelity rather than a forgery vector — but the book sentence claims a check that does not exist, and the corrective lives in a different document.

### Lower severity

- **`native_decide` concentration** (medium, acknowledged/interim): copy active-row fit (`actionCopyActiveRowFailures_eq_nil`), selector realization and fixed coverage (`ActionFixedCoherenceCompute`), gate data (`ActionGateCoherenceCompute`), permutation domain/chunks (`ActionPermutationDomainCompute`) all rest solely on whole-circuit certificates with no structural cross-check. We verified the *decided propositions* are the semantically correct ones rather than weaker proxies, and the design is fail-closed. Residual risk is compiler trust only; the files self-document as interim. Worth noting these certificates are the only barrier between "Clean `Constraints` proved" and "`Constraints` asserted at rows the SNARK does not constrain".
- **Fragile-but-double-guarded sentinels** (low, *carried over from the previous audit, still open*): the `(numFixedColumns, row, 0)` out-of-bounds sentinel for malformed selector mappings (`FixedColumns.lean:185-213`, rejected only by the strict `entry.1 < numFixedColumns` conjunct at `:326`) and the `permIndex` / `mkActionCell` mod-wrap totalization (`ActionCopyWitness.lean:649-663`, safe only because `actionCopyValue_eq_encodedAddress` is `rfl`). Both fail closed today; both would silently open if a refactor weakened one guard. The planned structural replacements should preserve them.
- **Non-primary instance columns commit to `0`** (low, latent; *carried over, still open*): `ActionInstanceCommitment.commitment` returns the group identity for any column ≠ `primary`, where halo2 would compute MSM + `1·w`. Dead branch today (one instance column, layout reaches only `primary.index`), but nothing pins "layout ⊆ {primary.index}" as an invariant, so a future multi-column variant would silently mis-model the verifier.
- **Clean-side `permIndex 9999` sentinel** (low, in the pinned dependency): a copy on a non-equality-enabled column silently vanishes from σ instead of erroring, where Rust `Assembly::copy` returns `ColumnNotInPermutation`. Cannot make a *proved* capstone unsound (the copy clause becomes unprovable, and the Action address certificate would fail), but the failure mode is silent misderivation rather than rejection.
- **Hygiene**: Clean pinned by 8-hex abbreviated rev in `lakefile.toml:13` while CompElliptic uses a full SHA (mathlib is pinned by release tag). The manifest resolves it, so the build is pinned; the concern is re-resolution on `lake update`. Worth a full SHA regardless of whether Clean counts as trusted: the `projectCS` / `Layout` / `recordedQueries` fixture-projection code that moved into Clean in this PR is load-bearing for the `CircuitCheck` empirical lane, which sits outside the capstone statement and so is unaffected by the argument that Clean is untrusted because the statement does not mention it. `Zcash.FastFieldNative` root module unreached by default targets (pre-existing, not a regression). `book.yml:61` fetches mutable `pull/31/head` (book-only lane). Dropping the pre-compression VK fixtures (`actionPre.json`, `AddPre`, `MulPre`) loses some selector-swap resolution — the post-compression comparison plus `TestSelMapDerivation` appear to cover it, but the pre-compression fixture was what made that unconditional (*carried over from the previous audit*). The two Clean synthesis-law fields (`lookupRelevantSelectorActivationsExact`, `lookupInputsNoSimpleSelectors`) are proved at real cost but have zero consumers in the Integration layer, which uses the interim sentinel route instead — worth confirming this is staging for the documented structural replacement. Several dead exports in `Certificate.lean` (`derivedFixedCommitments_eq`, `derivedPermutationCommonCommitments_eq`, `vk_eq_toVerifierKey` — the last presents itself in its docstring as the headline claim while carrying no weight). `PublicInputLayout` has no injectivity law though `clean-boundary.md:156` calls it injective; Action's layout is injective, so this is framework-level only.
- **Process**: the head merge commit itself has no CI run. Both parents are individually green (`4b76b56e` 15:55Z, `09867a83` 15:12Z) and the only overlapping files were `lakefile.toml`, `lake-manifest.json`, `TestFloorPlanner.lean`. Note `ActionVk.lean` was already present in both parents — `09867a83`'s green run demonstrates CI never elaborates it — so a green build at head would not have caught finding 1 either; it is still worth running, since the exact head has no run at all.

---

## What checked out sound

- **The capstone means what it claims.** `Statement pub = ∃ privateWitness, Spec pub privateWitness`, with `spec_iff` closing by `rfl` — pinning `Spec` definitionally to `SpecPost` of `combine` — `combine` injecting all ten public fields, and `extract_factorization` quantified over *all* environments forcing the public fields to be literal instance-cell reads. `assumptions_eq := rfl` forces `Assumptions = fun _ => True` — the "True top-level assumptions" claim is literal. No degenerate instantiation path.
- **Non-vacuity of the residual hypotheses.** Each of `claimed`, `hbindAll`, `hquot`, `hroute`, `hevals`, `hlen`, `hprob1`, `hξcur`, `hxgood`, `hgoodY` and the five challenge exclusions is individually dischargeable by existing in-repo machinery (`vanishing_slot_routed`, `deployed_member_node_binding[_at_point]`, `AcceptedModelClaimedEvaluations.ofOpenings` / `ofNodeBinding_or_relation`, `uniformChallenge_szBadSet`, `goodY_failure_measure_le`). None secretly encodes the conclusion. Composing the per-surface prices into one knowledge-error bound remains the documented open item — pre-existing, not a regression.
- **All five constraint families are faithful adapters.**
  - *Gates*: claimed only at extracted activation rows, with selector nonzeroness grounded (not assumed) via root well-formedness + `SelectorActivationsRealized`; rotations wrap mod `n` identically on feed and environment sides; the divisibility → all-rows step matches halo2's quotient argument; `Gate.WellFormed` is a genuine one-sided vanishing transport.
  - *Copies*: σ is forced to agree with the VK's permutation commitments (no independently-supplied permutation on the path); copies claimed only at rows `< n − blinding − 1`; δ-coset and ω-row injectivity genuinely proved; the chunked grand product matches halo2 including variable last-chunk width `[7,7,1]`, the `l_0` chain link, and the same-cycle no-op subtlety of `Assembly::copy`.
  - *Lookups*: θ-compression fold order (`acc·θ + v`) matches `compress_expressions`; collision surface priced per activation/table tuple pair over exactly the returnable rows; Clean membership claimed only at activations, so no default-tuple reasoning is needed; the active-row arithmetic (`u = n − bf − 2`, `l_last` at `−(bf+1)`, `l_blind` over `−1..−bf`) matches `plonk/verifier.rs` with no off-by-one; the permuted-column extraction assumes no injectivity it lacks.
  - *Fixed/selectors*: the compression model reproduces `compress_selectors.rs` (degree-0 alone at root 1, greedy combinations, packed columns offset past `numFixedColumns`); binding is whole-polynomial, so phantom activations from non-root nonzero cells are unreachable; conflicting co-activations fail the realization certificate.
  - *Instance*: commitment definition, `toElements` order vs layout rows 0–9 vs `extractPost`'s reads, all ten cells genuinely constrained (6 `constrainInstance` + 4 `assignAdviceFromInstance`), multi-proof indexing consistent across commitment/transcript/environment, and the AGM opening forces `uComp = 0 ∧ wComp = blind` rather than ignoring those components.
- **`vk_eq_derived` compares the full `VerifyingKey`** — all 14 fields in declaration order via `mk.injEq`, including all 29 fixed-column and 15 permutation-common commitments as real Vesta MSMs (list equality, so total function equality including out-of-range indices), with the `Shape` itself certified so the `List.ofFn` comparisons cannot truncate.
- **The build-target swap is a net strengthening.** All 14 modules under `Zcash/Arithmetic/` and `Zcash/Vendor/CompPoly/` remain in a default-target closure (two independent computations agree). At the merge base, 21 modules under `Circuits/Tests` and `Circuits/Fixtures` were in *no* target at all — corroborated by their loader paths pointing at a path that does not exist in this repo — so this PR compiles them for the first time. And the old `TestVkMatch*` was circular (the query seed was built from the dump's own layouts); the new form seeds from `Halo2.recordedQueries cs` plus an `invalidQueriedCells.isEmpty` guard, so query layouts are now genuinely compared.
- **The new FFT / inverse-DFT / Lagrange arithmetic is correct**: Cooley–Tukey identity with correct bit-reversal indexing, `bitrev` proved an involution, the imperative `bitreverse` proved equal to it, natural-order-in/natural-order-out, ω vs ω⁻¹ bridged explicitly, the `n⁻¹` scaling appearing exactly once per path, and a real tie to mathlib's `Lagrange.interpolate` rather than a self-referential spec. No `partial def`, `sorry`, `unsafe`, `implemented_by`, `opaque`, or `native_decide` in the four new files; all `getElem!`/`getD` at proved-in-range indices.
- **The Clean delta `fca63dc7..4e3613ed` contains no suspicious semantic change**: `Gate.wellFormed` became an intrinsic proof-carrying field (an obligation, not an assumption — soundness-positive, with exactly one custom proof, matching the book); the `closesEnvironment` refactor quantifies over all `ProofAssignment`s, which correctly prevents smuggling advice/instance-dependent facts into `EnvAssumptions`; the guarded V1 placement is self-consistent on both branches since keygen and semantics share the same starts. Rust-faithfulness spot-checks pass (blinding factors, `minimumRows`, `csDegree`, lookup required degree, `compress_selectors` greedy + short-circuit, query registration order).
- **The EnvAssumptions restructure shrank the trust base**: the old bundle *assumed* table contents (`GeneratorTableExact`, the `GeneratorTableLoaded` family, `LookupRangeCheck.TableLoaded`); the new one assumes only config equalities plus the usable-rows bound and *derives* the content facts from the constraint hypothesis, which is sound because `Constraints` on `.loadTable` pins the block plus the default fill.

Residual trust surface, stated plainly: the `native_decide` tier (documented, and pinned except for finding 2's gap); fixture provenance (captured from unpublished instrumented checkouts — `PROVENANCE.md` is candid that the SHA pin proves non-drift, not provenance); the unmodeled Fiat–Shamir transcript (pre-existing, `Challenges` is a free record by design); and the open knowledge-error composition.

---

## Suggested order of action

1. Delete `Deployed/ActionVk.lean` (re-apply `91413448`).
2. Add fully-qualified `assert_axioms … +native` for the deployed capstone and the certificate chain.
3. Write and consume the `instanceKey.generators` ↔ `capturedUrsGLagrange` composition lemma.
4. Fix the `closeWithOperations` inertness sentence in the book.
5. Run CI on the exact merge head before landing.

TalDerei (approved):
reACK on green CI!

.github/workflows/lean.yml
```diff
@@ -70,22 +70,19 @@ jobs:
...
-        run: python3 scripts/check_native_optin.py
-          diff Stamp.lean <(./stamp.sh) \
-            || { echo "::error::Stamp.lean is stale; regenerate with ./stamp.sh > Stamp.lean"; exit 1; }
       # Installs the toolchain from lean-toolchain, fetches the Mathlib olean
       # cache, and runs `lake build --wfail`, which rejects all warnings and builds
-      # `defaultTargets = ["Zcash", "FixtureCheck"]`. Building `FixtureCheck`
+      # the lakefile's `defaultTargets`. Building `FixtureCheck` also covers the
+      # deployed Action/Vesta capstone lane, whose `Keygen/Certificate.lean`
+      # natively certifies the deployed verifying key against the derivation
+      # (~45 s), and it
```
daira, line 85: This needs updating.

TalDerei: going to fold into #114.

TalDerei: and any other feedback post hoc

Zcash/Arithmetic/Fft.lean
daira, line 4: Wrong copyright, in three ways:
* The attribution is "Copyright 2026 Zcash Protocol Developers."
* The licence is dual MIT / Apache 2.0.
* There is no need for copyright information to be in individual files at all. That's just more chances to get it wrong. It's perfectly sufficient for it to be just in `README.md`.

```suggestion
```

Zcash/Arithmetic/FftSpec.lean
```diff
@@ -0,0 +1,918 @@
+/-
+Copyright (c) 2026 Ironwood Contributors.
+Released under the Apache License, Version 2.0.
+-/
```
daira, line 4: Wrong copyright; see above.

```suggestion
```

PR #100, opened 2026-07-24
feat(fv): wiring the permutation and lookup arguments

### LINKED ISSUE #36: Fold into permutation + lookup arguments

TalDerei:
Please ACK. Re-lands the work from #91, which was accidentally merged into the `decode-witness` branch instead of `main`.

This PR points the same `permutation-and-lookup-arguments` branch (13 commits, `6b3b07f..6bdba1e`) at `main`. Those commits are based on `b74c022`, which is already in `main`, so the diff is exactly #91's payload (17 files) and it merges cleanly with no conflicts.

GitHub does not allow changing the base of an already-merged PR, so this is opened as a fresh PR against `main`. The commit SHAs are shared with `decode-witness`, so no duplication results when that branch is later reconciled; `decode-witness` can be deleted once this lands.

Supersedes #91.

--------------

**Human Summary:**

Builds on #30, and closes [zcash/ironwood#36](https://github.com/zcash/ironwood/issues/36).

The soundness stack was written about gate checks only. This folds in the permutation and lookup arguments from #21 and carries them through the capstone chain (introducing full-constraint capstones). So the circuit-satisfaction relation (the predicate) now includes gates, permutation, and lookup.

----------

**AI Summary:**

This PR puts the permutation and lookup arguments into the constraint model, then proves both of them from checks the verifier already runs.

## What Is Proven

- **One set of builders:** the constraint builders never divide, so the same code produces the verifier values and the polynomials the proof is about.
- **Fingerprint:** the fold the verifier computes and the fold the proof uses are now one function run twice, so `hfp` is proved rather than assumed.
- **One check into many:** the verifier folds every constraint by `y` and checks a single value; a good `y` splits that back into one identity per constraint.
- **Rows:** read at a row, the permutation step rule *is* the running-product recurrence — both of its folds are running products.
- **Telescoping:** multiplying that recurrence across the rows, and across chunks through the chaining rule, leaves one product over every cell.
- **Multisets:** two root counts, one per challenge, carry that product to the multiset identity consumed by the #21 kernel — pairs for the permutation, and the two columns separately for the lookup.
- **Endpoints:** cells sharing a permutation cycle hold equal values; every lookup input value is a table value.

## Integration Status

The links are now chained through the deployed full constraint list. `eval_combineConstraints_deployed` identifies the verifier fold with the polynomial constraint fold; `hfold_of_constraint_polys`, `SnarkRelationWithMemberConstraints`, and the terminal wrappers carry that full numerator through the capstone chain. Circuit satisfaction is stated by `circuitSatViaConstraints`, covering gates, permutation, and lookup rather than gates alone.

The remaining top-level work belongs to #67 / PR #96: retain the per-proof decoder indexing, tie extraction to the actual batch and canonical member decodes, construct the adaptive probability coupling, consume the priced good-challenge event, and transfer the resulting composite bound to deployed acceptance.

Two hypotheses are conditions on the concrete key, written as ordinary binders rather than axioms: that the `δ`-powers below the column count lie in distinct cosets of `⟨ω⟩` (what key generation picks `δ` for), and that the Lagrange terms are nonzero at the rows used. Both are decidable for a given key. The case where the running product ends at zero is priced, not assumed away: every result ends in *either* the identity *or* an explicit vanishing factor.

## Issue Scope Check

#36 asked for #21 to be folded into a stack framed around gate checks. The three steps listed by #21 as still to come — grand-product telescoping, the Schwartz–Zippel bridge, and final lookup assembly — are proved here, and `perm_copy_constraints` and `run_structure` are reached from the deployed step rule. The expression-model and fingerprint leg is complete here; the final deployed-acceptance probability composition belongs to #67 / PR #96.


daira (approved):
utACK

Issue #72, opened 2026-07-20
Ledger-game theorems: compose binding-signature + key-binding into Balance / Spendability / Spend Authority


TalDerei:
The binding-signature/key-binding **machinery** is landed and (mostly) tracked, but the **ledger-game theorems** that compose it into the actual Balance / Spendability / Spend-Authority statements are absent from `main` and currently untracked. These are the follow-ups scoped in #44's description ("Chain and Effects models … the Balance-subset and Balance-value theorems … the Faerie-Gold core … the Spendability game; Spend Authority as a corollary") that never became an issue.

The intended framing is the strong-adversary model of the Zcon3 talk [*Understanding the Security of Zcash*](https://github.com/daira/zcash-security), composing the security arguments of [ZIP 2005](https://zips.z.cash/zip-2005).

## Landed dependencies (on `main`)

The pieces these theorems compose are already upstream:

- **Binding-signature balance core** — `Zcash/Security/BindingSignature/{Balance,Orchard,Sapling}.lean` (#4, #20): imbalance ⟹ DL reduction (breaks-as-computed-data), the no-overflow lift (field balance → integer balance), and the per-pool bundle capstones. Rests on RedDSA extractability (`hExtract`, tracked by #22) + DLR hardness.
- **Ledger foundations** — `Zcash/Security/Ledger/{Merkle,Statement}.lean` (#44): Merkle position binding (`collisionOfWrongLeaf`); the abstract Action statement + pinning lemmas `ivk_pinned`, `nk_eq_or_break`, `noteCommitBreakOfNe`, `nf_old_eq_or_break`; and the `KeyBindingInterface` slot.
- **Shared RO-collision / birthday vocabulary** — `Zcash/Security/Common/{RandomOracle,Birthday}.lean` (#44, #50).
- **Key-binding theorem (ROM)** — `Zcash/Security/KeyBinding/{Basic,Instance}.lean` (#50): the ZIP 2005 key-binding theorem, the ±-collision (`CollisionUpToSign`), the birthday bound `3·q·(q−1)/(2·|F|)`, `nk`/`ak`/`qk` pinning, and discharging `break_of_nk_ne`.

## Untracked composition targets (this issue)

*The checkbox states below indicate what has been done in #102 and #108.*

- [x] **Chain & Effects model** — witness-annotated transactions, valid chains, and the `Outputs` / `NonZeroSpends` / `PoolValueBalance` / `Issuance` linearizations.
[...]

Issue #22, opened 2026-06-29
RedDSA extractability


TalDerei:
The binding-signature side of the Balance argument consumes RedDSA extractability: from a transaction whose binding signature verifies under `bvk`, extract `bsk` with `bvk = [bsk] R`. This is the sole cryptographic input to the binding-signature balance argument (`Zcash/Security/BindingSignature/Balance.lean` § Assumptions), consumed by the ledger layer as the function-typed named hypothesis applied to `ValidLedger`'s `binding_verified` conjunct (the extraction input of `ValueShape.premissOrBreak` in `Ledger/Value.lean`).

The named form is a placeholder: it takes `bsk` as a theorem parameter and asserts the discrete log directly, so it is classically satisfiable, with computational content only relative to an efficient extractor. The real obligation is knowledge soundness in extractor-plus-knowledge-error form (`Pr[extract succeeds] ≥ ε − κ`) — the shape the forking layer already produces for the SNARK — and discharging it also means restating the consuming capstones in that form so the probabilities compose (the same restructuring #67 does on the SNARK side).

RedDSA itself is not yet formalized in the repo (`git grep RedDSA` over `Zcash/` hits only docstrings in `Security/BindingSignature/`): no scheme definition, no verification equation, no extractor. The forking machinery under `Snark/Soundness/Forking/` is aimed at Halo2/IPA and would need a signature-scheme instantiation.

The ±-randomized unforgeability obligation (Spend Authority's forgery arm) was previously bundled here; it is tracked separately in #121.

References discussion thread in https://github.com/zcash/ironwood/pull/4#discussion_r3415528911.

*[Restated by @daira (assisted by Claude Fable 5): scoped to extractability, with ±-randomized unforgeability split to #121; the previous body was a one-line pointer naming neither obligation.]*

PR #102, opened 2026-07-24
feat(fv): the deterministic layer of the Balance, Spendability, and Spend Authority games

### LINKED ISSUE #101: Deferred renames: `RT` → `MHASH`
Collecting renames agreed during the ledger-games work, to be done as mechanical PRs at a convenient point:

- ~~**`Sinsemilla.Chain` / `ChainTheorems.lean`** (circuits side): rename the chain-of-pieces modules (e.g. to `Pieces`) so that `Chain` is unambiguous for the ledger model's chain type (`Zcash/Security/Ledger/Chain.lean`), which is the more important use of the name.~~ Edit: no longer needed because we decided to use `Ledger` instead of `Chain` in the ledger model naming.
- **`RT` → `MHASH`** (ledger games): the role-type parameter `RT` in `Zcash/Security/Ledger/` reads as "root" but is the type of any Merkle node hash (`merkleCRH : RT × RT → RT`). Rename it across `Merkle.lean` / `Statement.lean` / `Chain.lean` and the canonical type-parameter lists.

Filed using AI (Claude Fable 5).

daira:
The ledger-model security games (the #72 scope): the witness-annotated ledger that the games run over, the effects read off it, and the deterministic layer of all three games — both Balance theorems, the Spendability core, and the Spend Authority reduction.

- A first commit tightens pre-existing statement-layer docs that described the intended instantiations' onward reductions in the present tense, as if formalized. Throughout the branch, docstrings and commit messages state intended-but-unformalized discharges as intended, checked per commit.

- `Statement.lean`: the mechanical `RT` → `MHASH` rename (fixes #101). The leaf is the extracted commitment itself: `CMX` merges into `MHASH`, matching the spec's typing, and the leaf map is `Extract_ℙ` (`P.extract`, not injective due to ±-fibres) rather than a separate coercion. `leafOf` is deleted — after removing its ρ argument (which only ever came from an offhand suggestion in ZIP 2005 about rehashing the tree), it became the identity function, and the type distinction was not worth the complexity.
- `Merkle.lean` adopts the circuit bridge's raw-encoding partial Merkle interface (from #98 — thanks @gio54321 — so the ledger and circuit layers share one model), and rebuilds the games' tree machinery on it. Hash inputs are raw encodings: an Orchard field element can have more than one accepted 255-bit encoding (the Sapling-era trade-off), so collisions are counted over the encoding domain, per height (counted from the leaves; the spec's root-counted *layer* is recoverable as `MerkleDepth − 1` minus the height), with each height's compression personalized. Compression is partial — `none` is a Sinsemilla exceptional branch — and `Path` records per-height success, so a collision is success-only (`DefinedCollision`): a totalized model would let two escapes masquerade as a collision that the concrete reduction could not consume. On-path encodings are pinned only up to the decode fibre, matching the deployed circuit; siblings are unconstrained; honest tree construction (`subRoot`) uses the canonical encoding. Position binding (`collisionOfWrongLeaf`) and completeness (`path_of_root`) are proven against this model, and the witness carries the pair-plus-side path shape the circuit actually witnesses.
- `Primitives` gains the ledger layer's obligations, following the `emb`/`emb_injective` precedent: `uncommitted` (the sentinel value for unused leaves), `uncommitted_ne` (no extracted note commitment collides with it; spec Thm 5.4.6 concretely), and `spendAuthVerify` over new sighash and signature type parameters `MSG` and `SIG`.
- `Model.lean` (namespace `Model`) defines the ledger the games run over: transactions of witness-annotated Actions with their spend-auth signatures, the leaf list and prefix anchors as deterministic partial functions of public data (an anchor equation also asserts that the referenced tree is defined), and `ValidLedger`'s consensus conjuncts. The transparent chain value pool is tracked only as its overall balance (`transparentPoolBalance`), which validity requires to be nonnegative (the consensus non-negative chain value balance rule for the transparent pool, ZIP 209). This requirement is on the cumulative value rather than per-transaction, because transparent funds accumulate and a per-transaction bound would underpower the adversary. Two deterministic facts come with the model: prefix-roots persist under ledger extension (the append-only tree), and output notes carry pairwise-distinct ρ because each is equal to its action's revealed nullifier — the Orchard Faerie-Gold protection, with no cryptographic hypothesis.
- `Effects.lean` reads a ledger's effects off its witnesses: an `Opening` is the `(rcm, note)` pair a commitment pins, outputs are openings at their leaf indices, and nonzero spends form a multiset of openings at their witnessed path positions. It is a multiset so that a double spend is countable, with Balance-subset to prove it is in fact a set.
- TrustBoundary census entries pin the new definitions at the strict flagless tier. `posVal` folds over `List.finRange` rather than taking a `Finset` sum, because `Fintype (Fin d)` carries `Classical.choice` in current Mathlib. A census-completeness pass also adds axiom probes for the statement layer's exported pinning theorems (`ivk_pinned` and `nk_eq_or_break`), which predate this branch but had no probes.

- `Balance.lean` proves **Balance-subset** as a computed reduction (`balanceSubsetOrBreak`). For a valid ledger, the nonzero spends of the first `i + 1` transactions form a sub-multiset of the positioned outputs of the first `i` — the one-transaction lag captures the strictly-earlier anchor rule, so a transaction can never spend an output of its own transaction. Each spend is the opening of the output that created its commitment, at that output's leaf position, and no positioned opening is spent twice. Otherwise the ledger's own data computes a `BalanceBreak`, with three arms: a Merkle `DefinedCollision` (one height's compression, success-only) when a spend's path validates a leaf that is not the committed one, against a defined tree; a `NoteCommitBreak` when a spend's commitment matches an output's with a different opening; and a key-binding break when two spends of one positioned opening would otherwise reveal the same nullifier twice. The anchor prefix is recovered by decidable search (`Nat.find`, itself axiom-free) rather than extracted from the validity hypothesis's existential. Duplicate spends are located by `findPair`, whose result is certified as a two-element *sublist*; the sublist carries "two distinct occurrences", so the nullifier-uniqueness contradiction is `Sublist.map` plus `Nodup`, with no index arithmetic. `Note`, `Opening`, and `PositionedOpening` gain derived `DecidableEq`, which is what makes the reductions' branch decisions computable.
- `Balance.lean` also proves **Balance-value**, in conservation form (`valueConservationOrBreak`): the shielded pool plus the transparent pool is exactly the minted issuance. The premiss is per-transaction: the net value a transaction's witnesses release from the shielded pool equals its declared `vBalance` — or the premiss hands back a break of an abstract type `VB`. Keeping `VB` abstract keeps the binding-signature machinery out of this layer. The intended deployed discharge is `NontrivialRelation.ofBundleIntImbalance`; that glue — including a binding-signature-verification conjunct on validity — is not yet formalized, and is planned for the stacked capstone PR. The corollary `balanceValueOrBreak` adds validity's `transparent_nonneg` and concludes that the shielded pool holds at most what issuance has minted.
- `Spendability.lean` starts the Spendability game's deterministic layer. The Faerie-Gold core (`faerieGoldCore`) says that nullifiers pin note tuples: two satisfied spends with distinct openings and equal revealed nullifiers compute a break. If the derive-inputs coincide, one commitment has two openings — a `NoteCommitBreak`. If they differ, they are themselves a `NullifierCollision`: two distinct note openings whose nullifiers agree. The collision carries the openings, so `cm` is pinned to a real commitment rather than a free group element — without them a collision would be fabricable (solve the point equation, or use the affine `extract`), certifying nothing. Reducing the collision to a nontrivial discrete-log relation is nonobvious and not yet formalized; it is planned for the stacked capstone PR. `respendOrBreak` inverts the nullifier roadblock: an action revealing the same nullifier as a satisfied spend either re-spends the same opening (the case Spend Authority handles) or computes a break. `validLedger_append` is persistence: a transaction valid just after `ledger` stays valid when appended to any valid extension of `ledger`, given that its nullifiers are still fresh and the boundary conjuncts hold. Anchors persist because prefix roots do. Both reductions sit at the census's strict flagless tier.

- The key-binding interface grows one guarantee for Spend Authority. `ak` is consumed as a single extracted coordinate, so the statement pins `ak^ℙ` only up to its y-sign. `KeyBindingInterface` gains `break_of_akP_ne`: two `KB`-witnesses sharing an `ivk` whose `akP`s are not equal up to sign are a break. `Extractor` gains the matching ±-property `toAK_pm` for the `ak` extraction (both extract maps are `Extract_ℙ` concretely), and `toInterface` discharges the new guarantee from the break projection, exactly as `break_of_nk_ne` is discharged from the `nk` component.
- `SpendAuthority.lean` is the deterministic core of the Spend Authority game. Suppose an action spends a note addressed to a key witness `wV`, in a transaction whose sighash the holder of `wV` never signed. If the action's `akP` matches `akP wV` up to sign, its signature is a `SpendAuthForgery`: a verifying signature under a randomization of `± akP wV`, over an unsigned message, with a `negated` flag recording the sign. The named unforgeability hypothesis for the ±-randomized signature scheme consumes that object; covering both signs is a factor-of-2 loss, fine with ~2^253 ±-fibres remaining. If the keys do not match even up to sign, receivability pins a shared `ivk` (`ivk_eq_of_receivable`, the one-witness form of `ivk_pinned`) and `break_of_akP_ne` yields a key-binding break. The capstone `spendAuthorityOrBreak` reads statement satisfaction and signature verification off `ValidLedger`.

- The book's security-definitions page gets the updated "one connected picture": the capstones grounded in their Lean names, the new nullifier-binding and Spend Authority chains, and nodes colour-coded by formalization status with a key. The Merkle `DefinedCollision` routes to the Sinsemilla discrete-log relation — its prequantum reduction, the same terminal as note-commitment binding — rather than to a hash-collision-resistance node, since BLAKE2b collision resistance does not enter the pre-quantum Balance argument. RedDSA unforgeability with ±-randomized keys is drawn as a named hypothesis rather than a terminal assumption; its intended discharge (the re-randomization reduction of [FKMSSS2016] section 3, adapted to the ± variant, plus forking extraction) is on the diagram, and #22 tracks proving it.

- The map is clickable: every node and each reduction-naming edge label links to its source (links into files added by this branch resolve once it merges; the RedDSA node links to #22), and the diagram's typography matches the book's — Raleway for the grouping titles, Open Sans for labels. The links needed mermaid's `securityLevel` set to `'loose'` in `mermaid-init.js`, which affects every diagram in the book. That widens nothing: mdbook already renders raw HTML from the markdown unsanitized, so diagram text is exactly as trusted as the rest of the book source.

The modelling decisions are recorded in the design doc attached to #72. What stays outside this PR: the named cryptographic hypotheses the reductions' outputs feed (unforgeability of the ±-randomized signature scheme, the reduction of `NullifierCollision` to a discrete-log relation and the hardness it rests on, the key-binding probability bound's composition into game-level statements), and the concrete Orchard instantiation.

Full build green.

Filed using AI (Claude Fable 5).

### DIFF
--- a/Zcash/Security/Ledger/Balance.lean
+++ b/Zcash/Security/Ledger/Balance.lean
@@ -19,10 +19,10 @@ opening is spent twice — or the ledger's own data computes a `BalanceBreak`:
 
 The shape of the argument (design doc, "The theorems"): the anchor pins each spend's
 path to a prefix tree; position binding pins the leaf; `uncommitted_ne` rules out the
-padding; the opening comparison pins the output or computes the note-commitment break;
-and duplicate spends of one positioned opening are found by `findPair`, whose result is
-certified as a two-element *sublist* — carrying "two distinct occurrences" with no index
-arithmetic — so `nfOldEqOrBreak` and nullifier uniqueness finish.
+padding; and the opening comparison pins the output or computes the note-commitment
+break. Duplicate spends of one positioned opening are found by `findPair`. Its result
+is certified as a two-element *sublist*, which carries "two distinct occurrences" with
+no index arithmetic, so `nfOldEqOrBreak` and nullifier uniqueness finish.
 
 Everything is a plain computable `def` per breaks-as-computed-data; the anchor index and
 the duplicate pair are found by decidable search, so no data is conjured from the
@@ -48,6 +48,8 @@ def findPair {α β : Type*} [DecidableEq β] (f : α → β) : List α → Opti
     | some b => some (a, b)
     | none => findPair f t
 
+/-- A successful scan certifies its pair: equal images under `f`, and the two entries
+form a two-element sublist of the input — distinct occurrences, in list order. -/
 theorem findPair_spec {α β : Type*} [DecidableEq β] (f : α → β) :
     ∀ {l : List α} {a₁ a₂ : α}, findPair f l = some (a₁, a₂) →
       f a₁ = f a₂ ∧ List.Sublist [a₁, a₂] l
@@ -67,6 +69,7 @@ theorem findPair_spec {α β : Type*} [DecidableEq β] (f : α → β) :
         obtain ⟨hfa, hsub⟩ := findPair_spec f h
         exact ⟨hfa, hsub.trans (List.sublist_cons_self a t)⟩
 
+/-- A failed scan certifies that all images under `f` are pairwise distinct. -/
 theorem findPair_none {α β : Type*} [DecidableEq β] (f : α → β) :
     ∀ {l : List α}, findPair f l = none → (l.map f).Nodup
   | [], _ => by simp
@@ -96,11 +99,13 @@ theorem flatMap_sublist {α β : Type*} {l : List α} {g g' : α → List β}
       simp only [List.flatMap_cons]
       exact (h a (by simp)).append (ih fun x hx => h x (by simp [hx]))
 
+/-- A ledger prefix contributes a prefix of the flattened action list. -/
 theorem outputActions_prefix {l₁ l₂ : Ledger KW F G RHO PSI MHASH MENC MSG SIG d}
     (h : l₁ <+: l₂) : outputActions l₁ <+: outputActions l₂ := by
   obtain ⟨r, rfl⟩ := h
   exact ⟨outputActions r, by simp [outputActions, List.flatMap_append]⟩
 
+/-- A ledger prefix contributes a prefix of the revealed-nullifier list. -/
 theorem nullifiers_prefix {l₁ l₂ : Ledger KW F G RHO PSI MHASH MENC MSG SIG d}
     (h : l₁ <+: l₂) : nullifiers l₁ <+: nullifiers l₂ := by
   obtain ⟨r, rfl⟩ := h
@@ -346,18 +351,21 @@ def issuanceTotal (issuance : ℕ → ℕ)
 
 section SumToolbox
 
+/-- Summing over a flat map is summing the per-chunk sums. -/
 private theorem sum_flatMap {α β : Type*} (g : α → List β) (f : β → ℤ) (l : List α) :
     ((l.flatMap g).map f).sum = (l.map fun x => ((g x).map f).sum).sum := by
   induction l with
   | nil => simp
   | cons a t ih => simp [List.flatMap_cons, ih]
 
+/-- The sum of pointwise differences is the difference of sums. -/
 private theorem sum_map_sub {α : Type*} (f g : α → ℤ) (l : List α) :
     (l.map fun x => f x - g x).sum = (l.map f).sum - (l.map g).sum := by
   induction l with
   | nil => simp
   | cons a t ih => simp only [List.map_cons, List.sum_cons, ih]; ring
 
+/-- Mapping a function of the entry over an index-zipped list ignores the indices. -/
 private theorem map_zipIdx_fst {α β : Type*} {f : α → β} :
     ∀ (l : List α) (n : ℕ), ((l.zipIdx n).map fun p => f p.1) = l.map f
   | [], _ => rfl
--- a/Zcash/Security/Ledger/Balance.lean
+++ b/Zcash/Security/Ledger/Balance.lean
@@ -5,9 +5,12 @@ import Zcash.Security.Ledger.Effects
 # Balance-subset: every nonzero spend is a committed output
 
 The first Balance theorem, as a computed reduction. For a valid ledger, the nonzero
-spends form a sub-multiset of the positioned outputs — each spend is the opening of the
-output that created its commitment, at that output's leaf position, and no positioned
-opening is spent twice — or the ledger's own data computes a `BalanceBreak`:
+spends of the first `i + 1` transactions form a sub-multiset of the positioned outputs
+of the first `i` — outputs of *strictly earlier* transactions, since an anchor
+references the tree at a transaction boundary, so a spend can never match an output of
+its own transaction. Each spend is the opening of the output that created its
+commitment, at that output's leaf position, and no positioned opening is spent twice —
+or the ledger's own data computes a `BalanceBreak`:
 
 * a Merkle `Collision` — one height's compression collided, with both evaluations
   successful — when a spend's authentication path validates a leaf that is not the
@@ -196,7 +199,7 @@ theorem satisfied_of_spendMem
 theorem anchor_of_spendMem
     (hval : ValidLedger P kv issuance maxActions ledger) {i : ℕ}
     {a : Action KW F G RHO PSI MHASH MENC SIG P.depth} (ha : a ∈ spendActions ledger i) :
-    ∃ j, j ≤ i ∧ rootAfter P ledger j = some a.inst.rt := by
+    ∃ j, j < i ∧ rootAfter P ledger j = some a.inst.rt := by
   obtain ⟨tx, htx, hafil⟩ := List.mem_flatMap.mp ha
   obtain ⟨k, hk, hgetk⟩ := List.mem_iff_getElem.mp htx
   have hki : k < i := lt_of_lt_of_le hk (by simp [List.length_take])
@@ -206,7 +209,7 @@ theorem anchor_of_spendMem
     exact hgetk
   obtain ⟨j, hjk, hrt⟩ := hval.anchor_valid ⟨k, hkl⟩ a
     (by rw [hget]; exact List.mem_of_mem_filter hafil)
-  exact ⟨j, le_of_lt (lt_of_le_of_lt hjk hki), hrt⟩
+  exact ⟨j, lt_of_le_of_lt hjk hki, hrt⟩
 
 /-- **Per-spend pinning.** A nonzero spend of a valid ledger is the positioned opening
 of the output that created its commitment — or the ledger data computes a break. The
@@ -215,11 +218,13 @@ computable. -/
 def spendPinnedOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
     [DecidableEq PSI] [DecidableEq MHASH] [DecidableEq MENC]
     (hval : ValidLedger P kv issuance maxActions ledger) {i : ℕ}
-    {a : Action KW F G RHO PSI MHASH MENC SIG P.depth} (ha : a ∈ spendActions ledger i) :
+    {a : Action KW F G RHO PSI MHASH MENC SIG P.depth}
+    (ha : a ∈ spendActions ledger (i + 1)) :
     (spendRecord a ∈ positionedOutputs ledger i) ⊕' BalanceBreak P kv :=
-  have hex : ∃ j, j ≤ i ∧ rootAfter P ledger j = some a.inst.rt := anchor_of_spendMem hval ha
+  have hex : ∃ j, j < i + 1 ∧ rootAfter P ledger j = some a.inst.rt :=
+    anchor_of_spendMem hval ha
   let j := Nat.find hex
-  have hji : j ≤ i := (Nat.find_spec hex).1
+  have hji : j ≤ i := Nat.lt_succ_iff.mp (Nat.find_spec hex).1
   have hpath : Merkle.Path P.merkle (P.extract a.w.cm_old) a.inst.rt a.w.path a.w.side :=
     (satisfied_of_spendMem hval ha).1.merkle_path (satisfied_of_spendMem hval ha).2
   have hroot : Merkle.root P.merkle (leafFun P (leafList (ledger.take j)))
@@ -264,7 +269,7 @@ def allPinnedOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
     [DecidableEq PSI] [DecidableEq MHASH] [DecidableEq MENC]
     (hval : ValidLedger P kv issuance maxActions ledger) {i : ℕ} :
     (L : List (Action KW F G RHO PSI MHASH MENC SIG P.depth)) →
-    (∀ a ∈ L, a ∈ spendActions ledger i) →
+    (∀ a ∈ L, a ∈ spendActions ledger (i + 1)) →
     ((∀ a ∈ L, spendRecord a ∈ positionedOutputs ledger i) ⊕' BalanceBreak P kv)
   | [], _ => .inl (by simp)
   | a :: t, hL =>
@@ -279,20 +284,23 @@ def allPinnedOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
           exacts [hmem, hall x hx'])
 
 /-- **Balance-subset, as a computed reduction.** For a valid ledger, the nonzero spends
-are a sub-multiset of the positioned outputs — or the ledger data computes a
-`BalanceBreak`. Duplicate spends are located by `findPair`; sharing a positioned opening
+of the first `i + 1` transactions are a sub-multiset of the positioned outputs of the
+first `i` — or the ledger data computes a `BalanceBreak`. The one-transaction lag is the
+strictly-earlier constraint: an anchor references a transaction boundary, so a spend
+never matches an output of its own transaction. Duplicate spends are located by
+`findPair`; sharing a positioned opening
 forces sharing a revealed nullifier (`nfOldEqOrBreak`), which nullifier uniqueness
 forbids, so the surviving branch is a key-binding break. -/
 def balanceSubsetOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
     [DecidableEq PSI] [DecidableEq MHASH] [DecidableEq MENC] [DecidableEq NK]
     [NoZeroSMulDivisors F G]
     (hval : ValidLedger P kv issuance maxActions ledger) (i : ℕ) :
-    (nonZeroSpends ledger i ≤ ↑(positionedOutputs ledger i)) ⊕' BalanceBreak P kv :=
-  match hfp : findPair spendRecord (spendActions ledger i) with
+    (nonZeroSpends ledger (i + 1) ≤ ↑(positionedOutputs ledger i)) ⊕' BalanceBreak P kv :=
+  match hfp : findPair spendRecord (spendActions ledger (i + 1)) with
   | some (a₁, a₂) =>
       have hspec := findPair_spec spendRecord hfp
-      have h₁ : a₁ ∈ spendActions ledger i := hspec.2.subset (by simp)
-      have h₂ : a₂ ∈ spendActions ledger i := hspec.2.subset (by simp)
+      have h₁ : a₁ ∈ spendActions ledger (i + 1) := hspec.2.subset (by simp)
+      have h₂ : a₂ ∈ spendActions ledger (i + 1) := hspec.2.subset (by simp)
       have hrcm : a₁.w.rcm_old = a₂.w.rcm_old := by
         have := congrArg (fun r => r.opening.rcm) hspec.1
         simpa [spendRecord] using this
@@ -306,15 +314,15 @@ def balanceSubsetOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
           -- one nullifier at two distinct occurrences — impossible in a valid ledger
           have hsubnf : List.Sublist [a₁.inst.nf_old, a₂.inst.nf_old] (nullifiers ledger) := by
             have h := (hspec.2.map fun a => a.inst.nf_old).trans
-              (spendActions_map_nf_sublist ledger i)
+              (spendActions_map_nf_sublist ledger (i + 1))
             simpa using h
           have hnd : [a₁.inst.nf_old, a₂.inst.nf_old].Nodup :=
             hval.nf_nodup.sublist hsubnf
           absurd hnf (by simpa using hnd)
   | none =>
-      have hnodup : ((spendActions ledger i).map spendRecord).Nodup :=
+      have hnodup : ((spendActions ledger (i + 1)).map spendRecord).Nodup :=
         findPair_none spendRecord hfp
-      match allPinnedOrBreak hval (spendActions ledger i) (fun _ h => h) with
+      match allPinnedOrBreak hval (spendActions ledger (i + 1)) (fun _ h => h) with
       | .inr b => .inr b
       | .inl hall =>
           .inl (by
--- a/Zcash/Security/Ledger/Balance.lean
+++ b/Zcash/Security/Ledger/Balance.lean
@@ -321,4 +321,185 @@ def balanceSubsetOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
 
 end Validity
 
+/-! ## Balance-value, in conservation form
+
+The value ledger balances: what the shielded pool holds plus what the transparent pool
+holds is exactly what issuance has minted — up to a per-transaction value premiss. The
+premiss says each transaction's witnessed net value matches its declared `vBalance`, or
+a value break of the caller's type is exhibited. The intended deployed discharge is the
+binding-signature machinery (`NontrivialRelation.ofBundleIntImbalance` with the
+statement's value ranges and the action-count bound), with a nontrivial discrete-log
+relation as the break; that glue is not yet formalized, and validity does not yet carry
+binding-signature verification. Balance is then the corollary: given the transparent
+pool never goes negative (`ValidLedger.transparent_nonneg`), the shielded pool holds at
+most what was minted. -/
+
+/-- A transaction's net value, read off its witnesses: spent minus created, over all
+actions (zero-valued dummies contribute zero). -/
+def txNetValue (tx : Tx KW F G RHO PSI MHASH MENC MSG SIG d) : ℤ :=
+  (tx.actions.map fun a => (a.w.note_old.v : ℤ) - a.w.note_new.v).sum
+
+/-- The issuance minted alongside the first `i` transactions. -/
+def issuanceTotal (issuance : ℕ → ℕ)
+    (ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG d) (i : ℕ) : ℤ :=
+  ((ledger.take i).zipIdx.map fun p => (issuance p.2 : ℤ)).sum
+
+section SumToolbox
+
+private theorem sum_flatMap {α β : Type*} (g : α → List β) (f : β → ℤ) (l : List α) :
+    ((l.flatMap g).map f).sum = (l.map fun x => ((g x).map f).sum).sum := by
+  induction l with
+  | nil => simp
+  | cons a t ih => simp [List.flatMap_cons, ih]
+
+private theorem sum_map_sub {α : Type*} (f g : α → ℤ) (l : List α) :
+    (l.map fun x => f x - g x).sum = (l.map f).sum - (l.map g).sum := by
+  induction l with
+  | nil => simp
+  | cons a t ih => simp only [List.map_cons, List.sum_cons, ih]; ring
+
+private theorem map_zipIdx_fst {α β : Type*} {f : α → β} :
+    ∀ (l : List α) (n : ℕ), ((l.zipIdx n).map fun p => f p.1) = l.map f
+  | [], _ => rfl
+  | a :: t, n => by
+    simp only [List.zipIdx_cons, List.map_cons]
+    rw [map_zipIdx_fst t (n + 1)]
+
+end SumToolbox
+
+/-- The transparent pool balance splits into minted issuance plus declared balances. -/
+theorem transparentPoolBalance_eq (issuance : ℕ → ℕ)
+    (ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG d) (i : ℕ) :
+    transparentPoolBalance issuance ledger i
+      = issuanceTotal issuance ledger i
+        + ((ledger.take i).map fun tx => tx.vBalance).sum := by
+  rw [transparentPoolBalance, issuanceTotal, List.sum_map_add, map_zipIdx_fst]
+
+/-- The outputs' value total, transaction by transaction. -/
+theorem positionedOutputs_value_sum
+    (ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG d) (i : ℕ) :
+    ((positionedOutputs ledger i).map fun p => (p.opening.note.v : ℤ)).sum
+      = ((ledger.take i).map fun tx =>
+          (tx.actions.map fun a => (a.w.note_new.v : ℤ)).sum).sum := by
+  rw [positionedOutputs, List.map_map]
+  rw [show ((fun p : PositionedOpening F G RHO PSI => (p.opening.note.v : ℤ))
+        ∘ fun p : Opening F G RHO PSI × ℕ => (⟨p.2, p.1⟩ : PositionedOpening F G RHO PSI))
+      = fun p : Opening F G RHO PSI × ℕ => (p.1.note.v : ℤ) from rfl]
+  rw [show ((outputOpenings (ledger.take i)).zipIdx.map fun p => (p.1.note.v : ℤ))
+      = (outputOpenings (ledger.take i)).map fun o => (o.note.v : ℤ) from
+    map_zipIdx_fst (f := fun o : Opening F G RHO PSI => (o.note.v : ℤ)) _ 0]
+  rw [outputOpenings, List.map_map]
+  rw [show ((fun o : Opening F G RHO PSI => (o.note.v : ℤ)) ∘ outputOpening)
+      = fun a : Action KW F G RHO PSI MHASH MENC SIG d => (a.w.note_new.v : ℤ) from rfl]
+  rw [outputActions, sum_flatMap]
+
+/-- The nonzero spends' value total is every spend's value total: the filtered-out
+spends have value zero. -/
+theorem nonZeroSpends_value_sum
+    (ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG d) (i : ℕ) :
+    ((nonZeroSpends ledger i).map fun p => (p.opening.note.v : ℤ)).sum
+      = ((ledger.take i).map fun tx =>
+          (tx.actions.map fun a => (a.w.note_old.v : ℤ)).sum).sum := by
+  have hfil : ∀ acts : List (Action KW F G RHO PSI MHASH MENC SIG d),
+      (((acts.filter fun a => a.w.note_old.v ≠ 0).map fun a =>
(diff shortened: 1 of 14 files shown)


daira (commented):
Initial self-ACK (will go through it in more detail tomorrow).

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,506 @@
...
+  differ;
+* a key-binding break, when two spends of one note tuple would otherwise have to reveal
+  the same nullifier twice.
+
+The shape of the argument (design doc, "The theorems"): the anchor pins each spend's
+path to a prefix tree; position binding pins the leaf; `uncommitted_ne` rules out the
+padding; the opening comparison pins the output or computes the note-commitment break;
+and duplicate spends of one positioned opening are found by `findPair`, whose result is
+certified as a two-element *sublist* — carrying "two distinct occurrences" with no index
+arithmetic — so `nfOldEqOrBreak` and nullifier uniqueness finish.
```
daira: I will do another pass to clean up long sentences like this; don't let it distract from reviewing the formalization content.

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,506 @@
...
+theorem flatMap_sublist {α β : Type*} {l : List α} {g g' : α → List β}
+    (h : ∀ x ∈ l, List.Sublist (g x) (g' x)) :
+    List.Sublist (l.flatMap g) (l.flatMap g') := by
+  induction l with
+  | nil => simp
+  | cons a t ih =>
+      simp only [List.flatMap_cons]
+      exact (h a (by simp)).append (ih fun x hx => h x (by simp [hx]))
+
+theorem outputActions_prefix {l₁ l₂ : Ledger KW F G RHO PSI MHASH MENC MSG SIG d}
```
daira, line 106: Note-to-self: add doc strings summarizing all theorems, even straightforward ones (to help people less familiar with Lean).

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,506 @@
...
+
+/-- **Per-spend pinning.** A nonzero spend of a valid ledger is the positioned opening
+of the output that created its commitment — or the ledger data computes a break. The
+anchor prefix is recovered by decidable search (`Nat.find`), so the branches stay
+computable. -/
+def spendPinnedOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
+    [DecidableEq PSI] [DecidableEq MHASH]
+    (hval : ValidLedger P kv issuance maxActions ledger) {i : ℕ}
+    {a : Action KW F G RHO PSI MHASH MENC SIG P.depth} (ha : a ∈ spendActions ledger i) :
+    (spendRecord a ∈ positionedOutputs ledger i) ⊕' BalanceBreak P kv :=
```
daira, line 223: This proof is a little complicated, maybe it can be decomposed.

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,506 @@
...
+            simpa using hall a haL)
+
+end Validity
+
+/-! ## Balance-value, in conservation form
+
+The value ledger balances: what the shielded pool holds plus what the transparent pool
+holds is exactly what issuance has minted — up to a per-transaction value premiss. The
+premiss says each transaction's witnessed net value matches its declared `vBalance`, or
+a value break of the caller's type is exhibited; at the deployed instantiation it is
```
daira:
```suggestion
a value break of the caller's type is exhibited. At the deployed instantiation it is
```

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,506 @@
...
+  (tx.actions.map fun a => (a.w.note_old.v : ℤ) - a.w.note_new.v).sum
+
+/-- The issuance minted alongside the first `i` transactions. -/
+def issuanceTotal (issuance : ℕ → ℕ)
+    (ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG d) (i : ℕ) : ℤ :=
+  ((ledger.take i).zipIdx.map fun p => (issuance p.2 : ℤ)).sum
+
+section SumToolbox
+
+private theorem sum_flatMap {α β : Type*} (g : α → List β) (f : β → ℤ) (l : List α) :
```
daira, line 363: Again, summarize every theorem in a doc comment, even trivial lemmas.

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,506 @@
...
+
+section Conservation
+
+variable [Field F] [AddCommGroup G] [Module F G]
+variable {P : Primitives F G IVK NK RHO PSI MHASH MENC MSG SIG}
+variable {kv : KeyBindingInterface KW G IVK NK}
+variable {ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG P.depth}
+variable {issuance : ℕ → ℕ} {maxActions : ℕ}
+
+/-- Run the per-transaction value premiss over a list, stopping at the first break. -/
```
daira, line 471: This has the same structure as `allPinnedOrBreak`, maybe we can factor out a helper.

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,506 @@
...
+  | .inr b => .inr b
+  | .inl hall => .inl (by
+      have hsum : ((ledger.take i).map txNetValue).sum
+          = ((ledger.take i).map fun tx => tx.vBalance).sum :=
+        congrArg List.sum (List.map_congr_left fun tx htx => hall tx htx)
+      rw [poolValueBalance_eq_neg, transparentPoolBalance_eq, hsum]
+      ring)
+
+/-- **Balance-value.** Given the transparent pool never goes negative, the shielded pool
+holds at most what issuance has minted — up to the value premiss's break. -/
```
daira, line 507: Again, we have a missing capstone that actually does the instantiation using the discharged value premiss.

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,506 @@
...
+    | .inl heq =>
+      match allValueOrBreak perTx t (fun x hx => hL x (by simp [hx])) with
+      | .inr b => .inr b
+      | .inl hall => .inl (by
+          intro x hx
+          rcases List.mem_cons.mp hx with rfl | hx'
+          exacts [heq, hall x hx'])
+
+/-- **Value conservation.** Up to the value premiss's break, the shielded pool plus the
+transparent pool is exactly the minted issuance. -/
```
daira, line 490: > The [value] premiss says each transaction's witnessed net value matches its declared `vBalance`, or a value break of the caller's type is exhibited; at the deployed instantiation it is discharged by the binding-signature machinery (`NontrivialRelation.ofBundleIntImbalance` with the statement's value ranges and the action-count bound), and the break type is a nontrivial discrete-log relation.

We have a missing capstone that actually does this instantiation.

daira: [Fable] Correct — two pieces are missing, and both are now planned for the capstone PR stacked on this one.

1. A binding-signature-verification conjunct on `ValidLedger`. `sig_verifies` covers spend authorization only; nothing yet says a transaction's binding signature verifies over its actions' `cv_net`s and its `vBalance`, so there is nothing for the discharge to consume.
2. The glue itself: a def turning that conjunct, plus `ActionSatisfied`'s `cv_net_eq`, the value ranges, and `action_bound`, into this `perTx` premiss with `VB := NontrivialRelation` via `ofBundleIntImbalance` — whose only consumers today are the binding-signature layer's own per-pool wrappers.

The shapes line up on both sides (`cv_net_eq` supplies exactly what `ofOrchardImbalance` consumes, given the new conjunct), so the work is bounded. Recorded in the design doc's decisions log as candidate scope for the stacked PR, alongside honest-sender completeness and the probabilistic capstones.

daira: The doc said the premiss was "discharged" but that was an overclaim. It's literally just some plumbing that is not done yet (and will be in my next PR), but I've put this PR back in draft until the overclaims are fixed. It's still coherent on its own, it just needs doc fixes.

Zcash/Security/Ledger/Effects.lean
```diff
@@ -0,0 +1,94 @@
+import Mathlib
+import Zcash.Security.Ledger.Model
+
+/-!
+# Effects: outputs, spends, and the shielded pool balance
+
+The effects of a valid ledger are read off its witnesses (Zcon3 slides 20-21). An
```
daira:
```suggestion
The effects of a valid ledger are read off its witnesses (Zcon3 "Understanding the Security of Zcash" slides 20-21). An
```
and rewrap the paragraph.

Zcash/Security/Ledger/Model.lean
```diff
@@ -0,0 +1,179 @@
...
+    (kv : KeyBindingInterface KW G IVK NK) (issuance : ℕ → ℕ) (maxActions : ℕ)
+    (ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG P.depth) : Prop where
+  /-- Every action's witness satisfies the Action statement. -/
+  satisfied : ∀ tx ∈ ledger, ∀ a ∈ tx.actions, ActionSatisfied P kv a.inst a.w
+  /-- No nullifier is revealed twice (the consensus nullifier-set rule; this forbids
+  duplicates even within one transaction). -/
+  nf_nodup : (nullifiers ledger).Nodup
+  /-- Each anchor is the root after some earlier transaction boundary: an action of
+  transaction `i` may reference the tree after any `j ≤ i` transactions, so only
+  outputs of strictly earlier transactions. -/
```
daira:
```suggestion
  transaction `i` (zero-based) may reference the tree after any `j ≤ i` transactions,
  so only outputs of strictly earlier transactions. -/
```

Zcash/Security/Ledger/Spendability.lean
```diff
@@ -0,0 +1,171 @@
...
+inputs on which `deriveNullifier` agrees. The collision is terminal vocabulary here:
+nothing in this development reduces it further. The intended deployed discharge is a
+named DL/Pedersen-hash hypothesis; for the Recovery Statement, ZIP 2005's Spendability
+theorem reduces it to an `H^rcm` ±-collision.
+
+Persistence says a transaction valid immediately after `ledger` stays valid when
+appended to any valid extension `ledger'`: anchors persist because prefix roots do
+(`rootAfter_prefix`), satisfaction and signatures are per-transaction, and the only live
+conditions are nullifier freshness and the boundary conjuncts (capacity and transparent
+nonnegativity), which appear as hypotheses. The roadblock inversion (`respendOrBreak`)
```
daira:
```suggestion
appended to any valid extension `ledger'`, under the conditions of nullifier freshness
and the boundary conjuncts (capacity and transparent nonnegativity), which appear as
hypotheses. Anchors persist because prefix roots do (`rootAfter_prefix`); satisfaction
and signatures are per-transaction. The roadblock inversion (`respendOrBreak`)
```

Zcash/Security/Ledger/Spendability.lean
```diff
@@ -0,0 +1,171 @@
...
+and the boundary conjuncts hold for the appended ledger. Anchors persist because prefix
+roots do; satisfaction, signatures, and the action bound are per-transaction. -/
+theorem validLedger_append
+    (hval' : ValidLedger P kv issuance maxActions ledger')
+    (hpre : ledger <+: ledger')
+    (hvalT : ValidLedger P kv issuance maxActions (ledger ++ [T]))
+    (hfresh : ∀ a ∈ T.actions, a.inst.nf_old ∉ nullifiers ledger')
+    (hcap : (leafList (ledger' ++ [T])).length ≤ 2 ^ P.depth)
+    (htrans : ∀ i : ℕ, 0 ≤ transparentPoolBalance issuance (ledger' ++ [T]) i) :
+    ValidLedger P kv issuance maxActions (ledger' ++ [T]) := by
```
daira, line 128: Long proof, can probably be usefully split.

Zcash/Security/Ledger/SpendAuthority.lean
```diff
@@ -0,0 +1,122 @@
...
+# Spend Authority: forgery or break
+
+The deterministic core of the Spend Authority game. Suppose an action spends a note
+addressed to a key witness `wV` (its `pk_d` is `emb (ivk wV) • g_d`), in a transaction
+whose sighash the holder of `wV` never signed. Then the ledger's own data computes one
+of two things. If the action's verification key matches `akP wV` up to sign, the
+action's signature is a `SpendAuthForgery`: a verifying signature under a randomization
+of `± akP wV`, over an unsigned message. If it does not match even up to sign, the two
+key witnesses share an `ivk` (receivability pins it) but not their `akP`, and
+`break_of_akP_ne` turns that into a key-binding break.
```
daira:
```suggestion
of two things:
* If the action's verification key matches `akP wV` up to sign, the action's
  signature is a `SpendAuthForgery`: a verifying signature under a randomization
  of `± akP wV`, over an unsigned message.
* If it does not match up to sign, the two key witnesses share an `ivk` (receivability
  pins it) but not their `akP`, and `break_of_akP_ne` turns that into a key-binding
  break.
```

Zcash/Security/Ledger/SpendAuthority.lean
```diff
@@ -0,0 +1,122 @@
...
+
+The sign matters because `ak` is consumed as a single extracted coordinate, so the
+statement pins `ak^ℙ` only up to its y-sign. A forgery therefore carries a `negated`
+flag, and the unforgeability hypothesis for the randomized signature scheme must cover
+both `akV` and `-akV`. That is a factor-of-2 security loss, fine with ~2^253 ±-fibres
+remaining.
```
daira:
```suggestion
remaining, and explicitly accepted in the design (see the note in spec §4.18.4).
```

Zcash/Security/Ledger/Statement.lean
```diff
@@ -48,57 +54,80 @@ structure Note (G RHO PSI : Type*) where
...
-(reducing breaks onward to random-oracle collisions or DLR relations). This interface is
-provisional: its shape is to be revisited against what the Balance and Spendability games
-actually consume once they are formalized — in particular `Break` being an opaque `Prop`
-limits the games to certificate-level break exhibition. -/
-structure KeyBindingInterface (KW G IVK NK : Type*) where
+condition `KB` enforced by the statement, and a `Break` predicate. `break_of_nk_ne`
+(consumed by Balance) and `break_of_akP_ne` (consumed by Spend Authority) are the
+guarantees the games use; the key-binding layer instantiates `Break` and discharges both,
+reducing breaks onward to random-oracle collisions. `Break` being an
+opaque `Prop` limits the games to certificate-level break exhibition. -/
```
daira: [To Fable] Did we decide whether that was a problem?

daira: [Fable] We decided it in effect while designing the capstones, and it is not a problem; the docstring now records the resolution. Three reasons:

1. The games exhibit the witness *pair* as data everywhere (`BalanceBreak.keyBinding` carries `w₁ w₂`; `KeyBindingBreakData` likewise); only the certificate `kv.Break w₁ w₂` is a `Prop`.
2. The probabilistic discharge needs exactly that shape: `toInterface_break_measure_le` bounds the probability of the *event* that the output pair satisfies `Break`, and `toInterface` interprets the interface's `Break` definitionally as the concrete one, so the games' break arms inherit the bound with no data required.
3. Data-level onward reductions (`CollisionUpToSign.ofBreak`) live below the interface, where the concrete witness structure is visible — the games never compute onward from a `Break` themselves.

If a future consumer did need break data at the games layer, the interface would want a data-valued field (`Break` as a Σ-type, or a `collisionOf` function); nothing so far points that way.

Zcash/Security/Ledger/Balance.lean
daira: Addressed by the amended wording — the overclaim fix rewrote this sentence, splitting it at the same point and stating the discharge as intended rather than done.

Zcash/Security/Ledger/Balance.lean
daira: Right — the same gap as the thread above: planned for the stacked capstone PR (a binding-signature-verification conjunct on validity, plus the glue to `ofBundleIntImbalance`). The docstring here now states the discharge as intended rather than done.

Zcash/Security/Ledger/Balance.lean
daira: Done.

Zcash/Security/Ledger/Balance.lean
daira: Done in the review-pass commit: every theorem, including the private sum-toolbox lemmas, now carries a doc comment.

Zcash/Security/Ledger/Balance.lean
daira: Done in the review-pass commit: every theorem, including the private sum-toolbox lemmas, now carries a doc comment.

Zcash/Security/Ledger/Balance.lean
daira: Done.

Zcash/Security/Ledger/Balance.lean
daira: Deferred by agreement so the capstone work is not held up; tracked for a cleanup pass.

Zcash/Security/Ledger/Balance.lean
daira: Deferred by agreement so the capstone work is not held up; tracked for a cleanup pass.

Zcash/Security/Ledger/Spendability.lean
daira: Deferred by agreement so the capstone work is not held up; tracked for a cleanup pass.

daira (commented):
Self-ACK. Ready for others' review.

Zcash/Security/Ledger/Merkle.lean
```diff
@@ -2,133 +2,397 @@ import Mathlib
...
+  Collisions are therefore counted over the encoding domain. `canon` is the canonical
+  encoding — a section of `decode` — used by honest tree construction.
+* **Compression is partial.** `none` is an escaped compression (for Orchard, a Sinsemilla
+  exceptional branch). Escapes are not totalized away: `Path` records per-layer success,
+  and a `Collision` demands *successful* evaluations on both queries
+  (`RandomOracle.DefinedCollision`), so an escaped query can never constitute a
+  collision. A totalized model would let two escapes masquerade as a collision that the
+  concrete reduction to a hardness assumption could not consume.
+
+Layers count from the leaves: `compress i` is the compression whose children are at
```
daira: In the protocol spec, layers count from the root:

> The layer numbered $h$, counting from layer $0$ at the root, has $2^h$ nodes with indices $0$ to $2^h - 1$ inclusive.

The encoding that is passed to Sinsemilla does count from the leaves, but that is handled by subtracting the layer number from $\mathsf{MerkleDepth^{Orchard}} - 1$ in section 5.4.1.3 'Merkle Tree Hash Function'. So the formalization is faithful to the circuit I think, but not to the terminology used by the spec.

daira: Fixed.

TalDerei (commented):
Audited the branch for spec fidelity, statement-level correctness, and soundness holes (the build and census being green, the audit targets what Lean can't check: whether the statements say the right thing and whether the model is faithful to the spec).

**No correctness bugs found.** The load-bearing arguments check out end to end: the position-binding descent emits a `DefinedCollision` at a single height with both evaluations against the same output, and correctly counts encoding-level disagreements (a non-canonical fibre representative compressing equal to the canonical pair *is* a genuine Sinsemilla collision — the encoding-domain counting is what the concrete bound needs); escapes can neither validate a path nor masquerade as a collision; `spendPinnedOrBreak`'s case split is exhaustive with `uncommitted_ne` consumed exactly at the padding arm; the `posVal`/`subRoot`/`leafFun`/`side` bit-order conventions agree end to end (LSB = height 0, `Fin.snoc` peels MSB at the root); the Balance-value signs match §4.14; the ± handling is coherent from `toAK_pm` through `break_of_akP_ne` to the forgery's `negated` flag. The declared seams (abstract `VB`, the named hypotheses the breaks feed) are exactly where #107 and #22 say they are.

Six findings, none tracked elsewhere; only the first is more than a docstring or nit.

### F1 (medium, assurance): nothing demonstrates a nontrivial `ValidLedger` is satisfiable

The empty ledger satisfies `ValidLedger` trivially, but no instantiation shows the conjuncts are *jointly* satisfiable for a ledger containing a real spend — `ActionSatisfied` (12 conjuncts) × `anchor_valid` × `nf_nodup` × the partial Merkle path. Reduction-style developments are most vulnerable to exactly this failure mode: an over-constrained premiss makes every `OrBreak` theorem vacuously strong, and nothing in the census can catch it. The concrete Orchard instantiation will discharge this eventually, but that is far off.

Suggestion: a toy instantiation (small `ZMod p`, mock compression, depth 1–2) with one output and one spend, proving `ValidLedger` holds — a permanent vacuity guard that would also smoke-test the computable reductions by `#eval`. Happy for this to be a follow-up issue rather than a pre-merge ask.

### F2 (docstring): the efficiency-by-inspection claim strains at `rootAfter`

`spendPinnedOrBreak` evaluates `rootAfter` inside the `Nat.find` loop, and `rootAfter` recomputes the full `2^depth` tree — at deployed depth 32, ~2^32 compressions per anchor candidate. The docstring flags the cost only for *testing*, but breaks-as-computed-data rests on reductions being efficient by inspection, and the inspection step here goes via extensional equality with the standard cached-frontier computation. That equivalence is exactly the kind of "established by inspection" remark the doctrine says to write down — suggest extending the `rootAfter` docstring to cover the reduction-efficiency claim, not just testing.

### F3 (docstring): `transparent_nonneg`'s consensus justification is unrecorded

The conjunct is asserted as a consensus rule, but deployed consensus enforces it only indirectly (UTXO accounting; there is no aggregate-pool rule for the transparent pool, unlike ZIP 209's for shielded pools). The implication — transparent value lives in nonnegative UTXOs, fees are internal transfers, so the cumulative balance is nonnegative — is true but nowhere stated, and the concrete instantiation will need it. One sentence in the docstring would close this.

### F4 (nit): `KeyBindingBreakData` duplicates `BalanceBreak.keyBinding`'s shape

Spend Authority packages the key-binding break as a structure while Balance uses anonymous constructor arguments on an inductive arm. Harmless today, but the probabilistic layer will have to consume both shapes; unifying now is cheaper than later.

### F5 (nit, census): `spendActions`, `spendRecord`, `outputOpening` have no direct entries

Covered transitively (a `noncomputable` dependency would fail the consumers' asserts), so this is convention-consistency only.

### F6 (nit, book): the `MC ----> CR` edge is coarser than the module doc, and the caveat doesn't cover it

`Merkle.lean` (and the neighbouring `SDLR → DL` edge) say Sinsemilla collision resistance reduces to DLR pre-quantum, but the diagram keeps hash-collision-resistance terminal for the Merkle collision, and the closing caveat mentions only the `NullifierCollision` edge's coarseness. Either a caveat sentence or an `MC → SDLR`-style edge would reconcile them. (All 21 click targets resolve on `main` — verified.)

---

Model-faithfulness notes, for the record (all safe-direction, none requiring action): transaction-granularity anchors are strictly more permissive than deployed consensus, so the theorems are strictly stronger; the omissions (blocks, fees, enable-flags, transparent structure) each enlarge the valid-ledger set or are value-neutral; `nf_nodup` correctly covers intra-transaction duplicates and dummy spends' nullifiers.

Reviewed with AI assistance (Claude Opus 5).

TalDerei (approved):

TalDerei:
sorry hold off on merging, this may have been a premature ACK. following on another thread which caught my eye.

Zcash/Security/Ledger/Spendability.lean
```diff
@@ -0,0 +1,171 @@
...
+  nk₁ : NK
+  ρ₁ : RHO
+  ψ₁ : PSI
+  cm₁ : G
+  nk₂ : NK
+  ρ₂ : RHO
+  ψ₂ : PSI
+  cm₂ : G
+  ne : (nk₁, ρ₁, ψ₁, cm₁) ≠ (nk₂, ρ₂, ψ₂, cm₂)
+  eq : P.deriveNullifier nk₁ ρ₁ ψ₁ cm₁ = P.deriveNullifier nk₂ ρ₂ ψ₂ cm₂
```
TalDerei: this seems to suffer from the same issue as https://github.com/zcash/ironwood/pull/50#discussion_r3570934165. affine coordinates can technically cancel the difference, so `NullifierCollision `is trivially constructible. it needs some other things i'd imagine in the break data to show  computational uniqueness of the valid note opening behind a nullifier.

Zcash/Security/Ledger/Model.lean
```diff
@@ -0,0 +1,182 @@
...
+  fun pos => l.getD (posVal pos) P.uncommitted
+
+/-- The anchor after the first `j` transactions: the Merkle root of their leaf list, or
+`none` if a compression along the honest tree escapes. This recomputes the full tree,
+which is exponential in `depth` — fine for proofs, but evaluating at realistic depths
+(e.g. against full-depth test vectors) would need the cached empty-subtree roots that
+real implementations use. -/
+def rootAfter (P : Primitives F G IVK NK RHO PSI MHASH MENC MSG SIG)
+    (ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG P.depth) (j : ℕ) : Option MHASH :=
+  Merkle.root P.merkle (leafFun P (leafList (ledger.take j)))
```
TalDerei, line 94: the “computed reduction” seems exponentially expensive, where's a PPT/security reduction?

Zcash/Security/Ledger/Balance.lean
```diff
@@ -0,0 +1,513 @@
...
+          intro x hx
+          rcases List.mem_cons.mp hx with rfl | hx'
+          exacts [hmem, hall x hx'])
+
+/-- **Balance-subset, as a computed reduction.** For a valid ledger, the nonzero spends
+are a sub-multiset of the positioned outputs — or the ledger data computes a
+`BalanceBreak`. Duplicate spends are located by `findPair`; sharing a positioned opening
+forces sharing a revealed nullifier (`nfOldEqOrBreak`), which nullifier uniqueness
+forbids, so the surviving branch is a key-binding break. -/
+def balanceSubsetOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
```
TalDerei, line 294: it was flagged that balance-subset proves a weaker invariant than #72 specifies (as it related to `NonZeroSpends `and `Outputs`)?

TalDerei (commented):
some more observations after another pass.

daira:
F1. That's the next PR. Completeness is not claimed to be verified in this PR.
F2. It's obvious that the optimization is possible (the real code does it) so I wouldn't consider this to invalidate the efficiency of the reduction.
F3. There actually [is a non-negative chain balance rule for the transparent pool in ZIP 209](https://zips.z.cash/zip-0209#specification). I'll reference it so that it's obvious where that comes from.
F4. Next PR.
F5. They're helpers that are covered transitively.
F6. "`Merkle.lean` (and the neighbouring SDLR → DL edge) say Sinsemilla collision resistance reduces to DLR pre-quantum, ..." Indeed it does. "... but the diagram keeps hash-collision-resistance terminal for the Merkle collision ..." Fair. `DefinedCollision` could instead reduce to SDLR. I believe Orchard does still rely on BLAKE2b collision resistance, but that correctly doesn't show up in the pre-quantum *Balance* proof (except that Halo 2 uses BLAKE2b as the Fiat–Shamir hash).

Zcash/Security/Ledger/Model.lean
daira: It's not exponentially expensive because the Merkle depth is fixed. With my theoretical hat on, $\mathsf{MerkleDepth}$ isn't $O(\lambda)$, it's $O(1)$ (it's decided by how long you need the protocol to run at a given transaction rate, not by the security parameter). Also a security loss of $2^{32}$ is not terrible for a 253-bit hash. And as the comment points out, the reduction is actually implementable in $O(\textit{number of commitments used} \times \mathsf{MerkleDepth})$, which is the same as an honest party's effort to sync the chain.

Zcash/Security/Ledger/Balance.lean
daira: Yes, `(nonZeroSpends ledger i ≤ ↑(positionedOutputs ledger i)) ⊕' BalanceBreak P kv` does not match $\mathsf{NonZeroSpends}(i+1) \subseteq \mathsf{Outputs}(i)$. Well spotted.

Zcash/Security/Ledger/Spendability.lean
daira: You're right — thanks. `NullifierCollision` kept only `(nk, ρ, ψ, cm)` with `cm` a free field, so although `faerieGoldCore` builds it from satisfied spends (where `cm` genuinely opens the note), the standalone certificate carries no evidence of that. With affine `Extract`, `cm₂` is then freely solvable to match x-coordinates, so holding one certifies nothing — it fails the terminal-event test.

Fix: carry the note openings on each side (`rcm`, `note`, `open : noteCommit rcm note = some cm`), mirroring `NoteCommitBreak`, so `cm` is pinned to a real commitment and the collision expands to a nontrivial DL relation at instantiation. The witnesses are already available at the one construction site.

Zcash/Security/Ledger/Balance.lean
daira: Fixed: `balanceSubsetOrBreak` now proves `nonZeroSpends ledger (i + 1) ≤ positionedOutputs ledger i` — the one-transaction lag that encodes the strictly-earlier anchor rule, so a transaction can never spend an output of its own transaction. The tightness was already there and discarded: `anchor_of_spendMem` computed `j < i` (anchor `j ≤ k`, `k < i`) and then weakened it with `le_of_lt`; keeping `j < i` was enough, no new lemmas.

Zcash/Security/Ledger/Spendability.lean
daira: Done: `NullifierCollision` now carries the note openings on each side (`rcm`, `note`, `open : noteCommit rcm note = some cm`), mirroring `NoteCommitBreak`, so `cm` is pinned to a real commitment rather than a free group element.

On the onward step: rather than assume a bespoke "nullifier-collision is hard" hypothesis, we're formalizing the reduction of the collision to a nontrivial discrete-log relation (`NontrivialRelation`) in the stacked capstone PR, before the deadline. The docstring records it as not-yet-formalized in the meantime.

TalDerei (approved):
utACK

PR #89, opened 2026-07-22
Add Ironwood circuit formalization

mitschabaude:
## Summary

- add the Ironwood circuit formalization under `Zcash.Circuits`
- include the circuit aggregator in the `Zcash` library target
- add Clean as a Lake dependency pinned to `fca63dc777ca9de7f533a50196cdd1357d569f4f`

## Motivation

The Ironwood circuits belong with the protocol formalization. Clean retains a temporary duplicate while its ongoing framework refactors settle, allowing later changes to be ported deliberately.

## Validation

- `lake build Zcash`
- `lake build`

PR #103, opened 2026-07-24
chore(fv): circuit/ suggestions

TalDerei:
Suggestions for https://github.com/zcash/ironwood/pull/98.


Zcash/Circuits/Fixtures/PROVENANCE.md
TalDerei, line 1: probing, not sure something like this is actually necessary, but maybe some kind of fixture provenance file might be useful? symmetrically on the `Zcash/Circuits/Fixtures/` side.

Zcash/Circuits/Tests/TestSelectorIsolation.lean
mitschabaude, line 1: This is not necessary, to make 99 work I already had to prove that property (for every gate)

Zcash/Circuits/Fixtures/PROVENANCE.md
mitschabaude: Makes sense, either this or remove those fixtures as they're somewhat redundant. But well they're fast and more detailed than the VK check

Zcash/Meta/AxiomCheck.lean
```diff
@@ -29,19 +29,29 @@ the check stable across the toolchain-dependent axiom naming. -/
 def isNativeDecideAxiom (n : Name) : Bool :=
   n.components.any (· == `native_decide)
 
+/-- The declared curve-order trust tier: the Pallas point count
+(`Zcash.Circuits.Point.pallas_natCard`, `Zcash/Circuits/Specs/Pallas.lean`) — the one
+explicit axiom behind the scalar-multiplication circuit proofs. Kept as a raw name literal
+so this meta module does not import the circuits tree. -/
+def curveOrderAxioms : Array Name := #[`Zcash.Circuits.Point.pallas_natCard]
```
mitschabaude: we should get rid of that axiom, we proved it in compelliptic!

mitschabaude (commented):
Several of the fixes are already in 99

Zcash/Circuits/Fixtures/PROVENANCE.md
TalDerei: yeah i'm agnostic on the direction here.

Zcash/Circuits/Fixtures/PROVENANCE.md
```diff
@@ -0,0 +1,70 @@
...
+| Fact | Status |
+|---|---|
+| halo2_proofs version | 0.3.2 (github.com/zcash/halo2) — recorded in `Layout.lean` |
+| Base (pre-NU 6.3) circuit | orchard 0.14.0 — the `actionBase` dump is byte-identical to the 0.14.0 circuit, verified on the ironwood branch (`TestVkLayoutActionBase.lean`) |
+| Post (NU 6.3 / ironwood) circuit | the orchard ironwood branch (`Circuit::synthesize` with `synthesize_cross_address_checks`) |
+| Instrumented halo2 / orchard commit hashes | **UNKNOWN — not recorded at generation time.** The dump forks have not been published; regeneration is currently not reproducible from pinned sources. |
+
+Closing that last row is the outstanding provenance task: publish the instrumented
+halo2/orchard branches (or fold `dump_lean`/`layout_dump` behind a feature flag
+upstream), record the exact commit hashes here, and add a CI job that regenerates the
```
daira: I thought they were already merged to main? Or does that need https://github.com/zcash/halo2/pull/922 ?

Zcash/Circuits/Fixtures/PROVENANCE.md
```diff
@@ -0,0 +1,70 @@
...
+orchard 0.15.1 / halo2 0.3.3, `orchard-fingerprint-instances`).
+
+## Content pinning
+
+Every JSON fixture is pinned by an FNV-1a-64 content hash **in the consuming test
+source** (`Fixtures/Json.lean` `loadJsonChecked`): a byte change fails the build until
+the pinned literal is updated, so the committed Lean source always pins the exact
+fixture content. Current artifacts:
+
+| File | FNV-1a-64 | Bytes | Pinned in |
```
daira: We should pin with SHA-256.

Zcash/Circuits/Tests/TestSelectorIsolation.lean
daira: That sounds blocking; we shouldn't duplicate this.

Zcash/Circuits/Fixtures/PROVENANCE.md
```diff
@@ -0,0 +1,70 @@
+# Fixture provenance
```
daira, line 1: This is great, thanks 🥰

Zcash/Meta/AxiomCheck.lean
daira: Revert this change and I'll add a commit that removes the axiom.

daira (changes requested):
Let's remove the curve order axiom rather than pinning it.

Zcash/Circuits/Fixtures/PROVENANCE.md
TalDerei: these is separate and not PRd anywhere. opting to not cut another release for purely organizational changes in that PR, which isn't this either.

TalDerei: clarified in https://github.com/zcash/ironwood/pull/103/commits/aee1c6fb71f0696620f41692286583a4039d4a0f.

TalDerei: https://github.com/zcash/ironwood/pull/103#discussion_r3648608875

TalDerei:
> Let's remove the curve order axiom rather than pinning it.

this has been fixed.

Zcash/TrustBoundary.lean
```diff
@@ -812,3 +815,62 @@ assert_axioms updEsc_escapesDuringC_measure_le
 assert_axioms adaptEsc
 assert_axioms adapt_decomposition
 assert_axioms adaptEsc_measure_le
+
+/-! ## The Action circuit — the halo2-native soundness trust surface
+
+The circuit-layer soundness theorems live at the `native_decide` tier: they consume
+`native_decide` certificates only — the six fixed-base window tables, small `interval_cases`
+facts, and CompElliptic's Pallas point-count witness (`pallas_natCard`, now a theorem backed
+by `CompElliptic.Curves.Pasta.Pallas.card_eq` rather than a curve-order axiom). These
```
daira:
```suggestion
facts, and CompElliptic's Pallas point-count witness (`pallas_natCard`). These
```
Removing recency bias.

Zcash/TrustBoundary.lean
```diff
@@ -812,3 +815,62 @@ assert_axioms updEsc_escapesDuringC_measure_le
...
+
+assert_axioms Zcash.Circuits.Action.Circuit.soundness +native
+assert_axioms Zcash.Circuits.Action.Circuit.soundnessPost +native
+assert_axioms Zcash.Circuits.Action.orchardActionCircuit +native
+
+/-! ## The circuit → ledger bridge — exported refinement theorems
+
+The end-to-end refinement from a satisfying Action assignment to the games-facing ledger
+statement (`ActionBreak … ∨ ∃ inst w, …`), and the break classifier's two correctness
+directions (a classified escape is a witness-tied break; a `none` verdict leaves every
```
daira: What does "`none` verdict" mean?

Zcash/TrustBoundary.lean
```diff
@@ -812,3 +815,62 @@ assert_axioms updEsc_escapesDuringC_measure_le
...
+assert_axioms Zcash.Security.Ledger.Bridge.specPost_to_ledger +native
+assert_axioms Zcash.Security.Ledger.Bridge.circuit_soundness_to_ledger +native
+assert_axioms Zcash.Security.Ledger.Bridge.actionBreak_of_classify +native
+assert_axioms Zcash.Security.Ledger.Bridge.classify_none_defined +native
+
+/-! ## The Sinsemilla discrete-log-relation reduction
+
+The onward step from a classified Action escape to the games-facing relation object, which
+the census above stops short of: the escaped chain is turned into an explicit generator
+combination (`ofPoint_hashToPoint`), the coefficient vector is computed from the break datum
```
daira: Nit: break data is plural because it has multiple fields (and because common usage rarely uses "datum"); this is consistent with `relationOfBreakData`.

```suggestion
combination (`ofPoint_hashToPoint`), the coefficient vector is computed from the break data
```

Zcash/TrustBoundary.lean
```diff
@@ -812,3 +815,62 @@ assert_axioms updEsc_escapesDuringC_measure_le
...
+(`breakCoeffs`, with its relation and nontriviality facts), and the two headline reductions
+package that as a `NontrivialRelationOne` at the escaped site's domain point.
+
+`relationOfBreakData` and `classifyRelation` are asserted computable rather than merely
+axiom-bounded: they are the reduction's data path, so the plain-`def` half of the check is
+what stops a later `noncomputable` marking from silently voiding the
+breaks-as-computed-data convention. `+native` covers the deployed bases' on-curve
```
daira:
```suggestion
`relationOfBreakData` and `classifyRelation` are asserted computable, per the
breaks-as-computed-data convention. `+native` covers the deployed bases' on-curve
```

book/src/formal-verification.md
```diff
@@ -66,10 +66,13 @@ different trust standards.
...
-trust. Instantiated at the concrete Vesta curve they additionally inherit one compiler-trust
+trust. Instantiated at a concrete Pasta curve they additionally inherit one compiler-trust
 axiom: CompElliptic's curve point-count, a closed computational fact discharged by `native_decide`
-(below). The `+native` flag on the corresponding build-time checks records exactly which endpoints
-carry it.
+(below). This applies to both the SNARK soundness endpoints (Vesta) and the Action circuit
+soundness (Pallas); the Pallas point count was formerly an explicit curve-order axiom and is now
+this same `native_decide`-backed CompElliptic result, so removing that axiom did not shrink the
+trusted base to nothing — it swapped a bespoke axiom for a re-checkable compiler-trust certificate.
+The `+native` flag on the corresponding build-time checks records exactly which endpoints carry it.
```
daira:
```suggestion
(below). This applies to both the SNARK soundness endpoints (Vesta) and the Action circuit
soundness (Pallas). The `+native` flag on the corresponding build-time checks records
exactly which endpoints carry it.
```
Removing recency bias / undue weight.

daira (commented):
Reviewed the full diff. This tightens exactly the right things: SHA-256 content pinning with a CI guard whose exact-listing `diff` also closes the new-fixture-slips-in-unpinned hole; a provenance record that is accurate about what is and is not reproducible; the census brought to uniform tiers (`+native` assertions replacing the `lean_verify` deferrals, and the new TrustBoundary sections for the Action stack, the bridge exports, and the SDLR reduction); and the Pallas curve-order axiom replaced by CompElliptic's theorem — a strict improvement (a bespoke axiom swapped for a re-checkable compiler-trust certificate). The open threads already on the PR cover the rest; one addition below.

## Making Lake track the fixtures (the module-doc caveat is fixable)

`Fixtures/Json.lean` documents that a local incremental build may skip the semantic `#eval` after a fixture regeneration, with CI as the authority. I tested the two Lake-native candidate mechanisms on a minimal v4.30 package (this repo's toolchain); both fail, and one working mechanism remains:

1. **`input_file` targets + `extraDepTargets` — does not propagate.** The machinery is wired (a bogus target name in `extraDepTargets` fails the build; the `input_file` target builds), but changing the input file's content does not re-elaborate the consuming module, at either library or package scope: extra-deps are build-order only on this toolchain. (Watch a false positive when testing this: any lakefile edit does invalidate module caches via the config hash, so the first build after adding the config re-elaborates once regardless.)
2. **`include_str` — does not track, and is worse than the status quo.** After changing the included file, the cached olean is replayed with the stale embedded string. That would downgrade the current failure mode (a skipped check, CI catches it) to a locally *passing* check over stale data.
3. **A generated stamp import — works, by construction.** Have the `TestVk*` modules `import` a tiny generated `Fixtures/Stamp.lean` rendering `SHA256SUMS` as a string literal. Regenerating a fixture already obliges refreshing `SHA256SUMS` (the CI exact-listing guard enforces it), so the stamp refresh folds into that step, with one more CI `diff` (stamp ↔ SHA256SUMS) closing the forgot-to-refresh hole. The source edit Lake does track — the changed import — then happens exactly when a fixture changes. This restores what removing the FNV pin lost (content change → source change → rebuild), keyed to the OS-side SHA-256 already maintained, with no Lean-side hashing.

So 3 is the fix, and it is cheap and mechanical. Not blocking — the CI-as-authority stance is sound meanwhile.

Zcash/TrustBoundary.lean
TalDerei: it just means that nothing was found to be wrong (ie. every hash in that witness evaluates correctly).

TalDerei:
I addressed the feedback here, so we can just move any other follow-ups to #89.

Issue #104, opened 2026-07-25
Tighten adaptive multiopen reduction so DLOG determines concrete security


TalDerei:
## Problem

PR #96 composes the deployed knowledge-error bound as

```text
(Q+k) * 3/|Fp| + (Q+1)/|Fp| + |basis| * epsilon_DL
  + (Q+4) * tau + (Q+1) * epsilon_x.
```

The adaptive multiopen term currently requires `s <= tau^4`. The deployed action budget bounds
`s` by a quantity of order `C(A)/|Fp|`, where `A = shape.numProofs`. For the captured Orchard
shape:

```text
queryBudget(A) = 50*A + 46
C(A) = 3*queryBudget(A) + max(2^11, queryBudget(A)) + 10.
```

At the consensus maximum `A = 2^16 - 1`:

```text
C(A) = 13,107,194
tau >= (C(A)/|Fp|)^(1/4) approximately 2^-57.59.
```

Therefore `(Q+4) * tau` becomes order-one at roughly `Q = 2^57.6`. Under the usual
time-success interpretation `Q <= T`, this makes the current composite proof vacuous long before
the fixed-slot DLOG term `2050 * epsilon_DL(T)` reaches order one (roughly `T = 2^121.5` under
the generic `epsilon_DL(T) approximately T^2/|Fp|` model).

## Goal

Tighten the adaptive multiopen / Fiat-Shamir coupling so that it does not determine the
constant-success concrete work factor. At the deployed parameters, the DLOG branch should remain
the limiting computational term.

Potential directions include exploiting the concrete prefix and stability structure instead of the
generic four-level peeling argument, deriving a direct sequential lazy-sampling bound, or otherwise
avoiding the combination of the fourth-root loss and the linear `(Q+4)` escape union bound.

## Acceptance criteria

- State the resource interpretation connecting the oracle-query budget `Q` to adversary work `T`.
- Replace or strengthen the `s <= tau^4` / `(Q+4) * tau` step used by the deployed composite
  capstone, or prove a tighter deployed-specific theorem.
- Specialize the replacement at `k = 11` and `shape.numProofs <= 2^16 - 1`.
- Add a concrete comparison showing that the multiopen contribution does not become order-one
  before the DLOG contribution under the stated time-success model.
- Wire the improved term into the PR #96 composite capstone and pin the new endpoint in the trust
  boundary.
- Document clearly which part is a proved reduction bound and which part remains a quantitative
  DLOG hardness assumption.

## Related

[...]

PR #106, opened 2026-07-25
chore(fv): tighten multiopen extractor

TalDerei:
**Human Summary:**

Contributes to https://github.com/zcash/ironwood/issues/104 that'll be fully closed when it's wired into capstones in #96.

Working through #96, i noticed i modeled the multiopen decode layer too generically, and it needs tightening. This became clear because the final adversary bound wasn't dominated by DLOG under some deeper formal analysis, but instead by some lossiness in the multiopen extraction. It performed rewinding like the IPA extractor, but converting that to an adversary-advantage bound revealed a security loss in my construction, so after ~2^58 work the formal bound goes vacuous. Instead of challenge rewinding, it needs AGM-based unbatching, which is more accurate and should be an isolated change. then DLOG should dominate in the final bound. So this provides the multiopen machinery and its concrete bound for #96 to consume.

-----------

**AI Summary:**

This PR replaces the rewinding multiopen extractor with rewind-free AGM unbatching. The old decoder rewound the verifier for accepting `x₄`/`x₁` families and paid a fourth-root threshold; here the online algebraic representations already carry the column witnesses, so unbatching is deterministic and the only probabilistic step is value binding, priced as one explicit root event per squeeze. The multiopen contribution becomes additive, and DLOG again dominates the deployed bound.

## What Is Proven

- **Rewind-free unbatching:** at `x₄`, `x₃`, `x₂`, and per-set `x₁`, represented columns either reconstruct the aggregate AGM coordinates or return an explicit augmented-basis relation. No accepting rewinds are used. The mathematical compatibility constructor still performs offline Vandermonde decoding over the AGM representation function; #96 must replace it with a computable direct-coordinate outcome before applying concrete DLOG hardness.
- **Table-aware squeeze pinning:** every bad-root set may read the run output and earlier random-oracle answers, but is invariant under reprogramming the answer at its own squeeze (`PinnedRootFamily`). `tableReadingPinnedRootEvent` demonstrates that this causal interface is instantiable.
- **Additive root budget:** `algebraicRootBudget` unions the six deployed root events (two IPA shift polynomials plus `x₄`, `x₃`, `x₂`, per-set `x₁`) into one shape-only bound. There is no joint accept floor to fourth-root.
- **Constraint supply from online provenance:** the pre-`x` constraint polynomial is built from the representation list retained by `DeployedBatchWitness`, with the quotient reconstructed pre-`x` as `Σᵢ X^(n·i)·hᵢ` so its bad set is pinned before the `x` squeeze. The only disagreement branch returns explicit relation coefficients, charged through the same single-instance textbook-DLOG finder.
- **Composite endpoint:** `snarkConstraintsDeployed_prob_le_of_root_schedule` bounds deployed extraction failure by `(Q+k)·3/|Fp| + (Q+1)/|Fp| + |basis|·ε_DL + (Q+23)·algebraicRootBudget + (Q+1)·ε_x`.
- **Concrete comparison:** at the consensus maximum (`k = 11`, `numProofs ≤ 2¹⁶ − 1`) the budget is `53,686,986,342,456/|Fp|`, so under `Q ≤ T` the multiopen term is still `≤ 2⁻⁸⁶` at `T = 2¹²²`. The `τ`-based term it replaces went vacuous near `Q ≈ 2^57.6`.

## Remaining Floor

Blake2b and generator hash-to-curve are modeled as random oracles, and Vesta discrete log remains the computational assumption. `DeployedRootSqueezeInvariance` and `DeployedConstraintXSqueezeSchedule` are the online-AGM causal premises — retained representations and earlier squeeze answers may determine a root set, but its own squeeze answer may not — and are stated, not discharged for the deployed family. `DeployedConstraintStaticChecks` remain verifying-key facts for #96 to discharge against the captured Orchard key.

## Issue Scope Check

This delivers #104's replacement term, its `k = 11` / consensus-maximum specialization, and the concrete DLOG comparison, for #96 to consume. The rewind-based budget stack (`VestaBudget`, `BudgetedExtraction`, `FloorBudget`, `AdaptiveCoupling`, `Residual`, `Prefixes`, `Completeness`, `Assembly`; 5,428 lines) is deleted, and the conditional `hExtract` route survives only as compatibility surface that the new capstone does not consume. The per-directory trust boundaries are consolidated into `Zcash/TrustBoundary.lean`, where the new endpoints are pinned.

## Integration-time TODOs (#96)

- Instantiate the captured Orchard verifying key and discharge `DeployedConstraintStaticChecks`.
- Construct `DeployedRootSqueezeInvariance` and `DeployedConstraintXSqueezeSchedule` with their concrete bounds.
- Replace the offline Vandermonde compatibility outcome with a computable direct-coordinate outcome and wire it into `constraintsBadAccept`.
- Connect `numProofs ≤ 2¹⁶ − 1` and `Q ≤ T` to the actual composite bound.
- Account for the combined finder's runtime/PPT overhead before applying a quantitative DLOG time-success model.


Zcash/Snark/Soundness/AGM/DeployedPinnedRoots.lean
```diff
@@ -0,0 +1,613 @@
...
+
+namespace ComputedDeployedRootFSFamily
+
+/-- Forget the tighter root data. -/
+abbrev toFamily (family : ComputedDeployedRootFSFamily shape) :
+    ComputedAlgebraicFSFamily shape := family.toComputedAlgebraicFSFamily
+
+/-- The complete relation producer used by the rewind-free deployed layer: the recursive
+extractor's `relationFinder` extended by the `outcome` relation.  Both branches stay in one
+data-returning finder — a bare `HasNontrivialRelation` would be vacuous in a prime-order group
```
daira: Nit: I think `HasNontrivialRelation` no longer exists. Non-blocking (but this PR requires a rebase anyway).

TalDerei: oh yeah you're right, recalling we renamed this to `NontrivialRelation`.

daira: I think `HasNontrivialRelation` is also mentioned in `TrustBoundary.lean`.

daira (approved):
There is a *lot* of code here (and code movement), so I couldn't do more than skim it. But I saw nothing problematic.

Issue #107, opened 2026-07-25
Binding-signature glue and the games' named hypotheses


TalDerei:
This issue tracks follow-ups to #72 that are left open by #102 ("the deterministic layer of the Balance, Spendability, and Spend Authority games") and #108 ("the ledger-game capstones and shape-level discharges").

## #102 composes the key-binding side, but not the binding-signature side

Balance-subset genuinely consumes the landed machinery — Merkle position binding, note-commitment binding, and the `KeyBindingInterface` (which this PR extends with `break_of_akP_ne` for Spend Authority). But Balance-value keeps the per-transaction value premiss at an abstract type `VB`; the glue to `NontrivialRelation.ofBundleIntImbalance` — the actual binding-signature discharge, plus a binding-signature-verification conjunct on `ValidLedger` — was deferred to the stacked capstone PR, and is now formalized there (#108: `ValueShape.premissOrBreak`), so the landed BindingSignature files are wired in. The shape of the discharge, as landed:

1. `{VB : Type*}` → `NontrivialRelation V R`
2. `perTx` as a parameter → a body that calls `ofBundleIntImbalance`
3. that call needs `hExtract` — supplied through the new `binding_verified` conjunct on `ValidLedger` and a function-typed named extraction hypothesis; underneath it RedDSA knowledge soundness (#22) remains

## #102 includes the deterministic layer only

[...]

PR #108, opened 2026-07-26
feat(fv): the ledger-game capstones and shape-level discharges

### LINKED ISSUE #72: Ledger-game theorems: compose binding-signature + key-binding into Balance / Spendability / Spend Authority

daira:
This PR carries the completeness direction, the probabilistic layer, and the shape-level discharges of the ledger games — the #72 scope, continued on top of the now-merged #102. fixes #72

- `Model.lean` gains append and saturation lemmas: the leaf list distributes over concatenation, and the transparent pool balance ignores extensions within a prefix, saturates at the ledger's length, and steps by issuance plus `vBalance` across an appended transaction.

- `Completeness.lean` checks that the protocol's intended use actually works. `HonestAction` packages what an honest wallet knows for one Action; from that data alone the module constructs the Action and proves the statement (`satisfied` — the Merkle conjunct is `path_of_root` on the honest authentication data, which exists because the referenced tree is defined). `honestTx` bundles any number of Actions up to the action bound and declares its true summed net value, so honest transactions discharge the Balance game's value premiss; `honestTx_valid` shows appending one preserves validity. Inclusion is required only of nonzero-valued spends, and `HonestAction.withDummySpend` constructs the dummy case — an honest wallet does not need a received note to create an Action. Per-action anchors within one transaction are a documented simplification: deployed Orchard encodes one anchor per transaction (Sapling v4 allowed per-spend anchors), and per-action anchors only widen what the adversary may do.

- `Capstone.lean` is the probabilistic layer. An adversary is an arbitrary `PMF` over valid annotated ledgers (`ValidAnnotated`). Restricting to valid ledgers loses no generality: every bad event requires validity, so conditioning on validity can only increase success probability, and a bound over valid samples implies the bound over bare ledgers. Events are reduction-branch preimages —"the reduction lands in this arm on this sample"— so no choice is needed to extract break data. The composition is pure event algebra: each violation event is contained in the union of its arm events, so its probability is at most the sum of the per-arm probabilities (`measure_mono` plus `measure_union_le`), each bounded by a named ε hypothesis. Instances: Balance-subset (Merkle, note-commitment, and key-binding arms), Balance-value (the value-premiss arm, in conservation and corollary forms), and Spend Authority (forgery and key-binding arms, for a victim key witness and its signing history).

- `Nullifier.lean` is the nullifier-binding reduction deferred from #102. `NullifierShape` is the additive derivation shape of the deployed construction (the [Orchard nullifier design](https://zcash.github.io/orchard/design/nullifiers.html)): derivation is a scalar multiple of the nullifier base `K` plus the commitment point, extracted with the ±-property. Commitments open as coefficient combinations of the fixed commitment bases (for Sinsemilla: `𝒬, 𝒮(0), 𝒮(1), …`) plus a multiple of the randomness base `R`. `NontrivialRelation.ofNullifierCollision` computes a nullifier collision directly into the balance argument's terminal — a nontrivial relation among the commitment bases, `K`, and `R`, under the same independent-hash-to-curve-bases discharge. Nontriviality comes from the shape's two coefficient conditions (distinct notes have distinct coefficient vectors; coefficient vectors never cancel additively), both concrete consequences of the initial `2^n • 𝒬` term. The reduction needs no distinct-notes premiss: the collision's own opening distinctness covers every case (distinct notes use the coefficient difference; equal notes force distinct commitment randomness). It therefore composes directly with the collisions `faerieGoldCore` produces, which hold exactly that inequality at construction.

- `spendabilityOrBreak` (in `Completeness.lean`) is the Spendability capstone at the deterministic layer: the owner of a received note can spend it. Either appending the honest transaction keeps the ledger valid, or the ledger's own data computes the obstruction — an action already spending the owner's exact opening (`Respend`, the case Spend Authority covers), a note-commitment break, or a nullifier collision. The roadblock branch decides nullifier membership, searches out the revealing action, and runs `respendOrBreak` against the owner's satisfied action.

- The security map routes `NullifierCollision` to the Sinsemilla discrete-log relation via the formalized reduction, and drops the caveat about an undecomposed `PRF^nf` component. The reduction needs no PRF assumption: the case that would have needed one —equal notes and randomness under distinct `nk`— cannot arise, because equal openings pin the commitment and the two spends coincide. This matches the Orchard design page's conclusion that Faerie Resistance rests on discrete log alone.

- The model gains the binding-signature side: `Primitives.bindingVerify` (the deployed pool's concrete primitives take the predicate as a parameter, like spend-auth verification), transactions carry their binding signature, `Tx.bvk` is the binding verification key (`∑ cv_net − ValueCommit_0(vBalance)`), and `ValidLedger` gains the consensus binding-signature rule and the `valueBalance` range rule (reusing `valueBound`, so no new game parameter). The validity constructors establish both — persistence per-transaction, the honest constructors from two wallet-checkable boundary hypotheses. `Value.lean` then discharges Balance-value's per-transaction premiss at the Pedersen shape. `ValueShape` is the shape `valueCommit v rcv = v • V + rcv • R` over abstract fixed bases — 𝒱 and ℛ at the intended instantiation; the concrete Pallas instantiation, with Sinsemilla-derived bases and RedDSA on Pallas, is deferred per the abstract-route plan. `ValueShape.premissOrBreak` decides the net-value equation; on failure it computes the binding-signature layer's nontrivial `(V, R)` relation via `ofBundleIntImbalance`. The no-overflow bound comes from the statement's value ranges, validity's action-count and `vBalance` range rules, and one numeric hypothesis `(maxActions + 1) * valueBound ≤ r`. The extraction input is the function-typed named form of RedDSA extractability applied to the new conjunct — a placeholder: the total form is classically satisfiable, so it has computational content only relative to an efficient extractor (#22 tracks the forking extractor-with-knowledge-error form). `ValueShape.balanceOrBreak` and `ValueShape.conservationOrBreak` compose the premiss into the Balance-value capstones, with the abstract `VB` instantiated at the shape level. Shape members drop from their base names what the namespace already says, so no base name collides with the abstract capstones they instantiate.

- `KeyBindingArm.lean` bounds the key-binding arms in the key-binding oracle model at the reductions' own events: `(n + 4)(n + 3)/|RIVK|` for any `n`-query-bounded ledger adversary, via `toInterface_break_measure_le` at an unchanged query count. The sampling is the named `kbExperiment` (do-notation, shared with the key-binding layer's bound; the oracle-model types share a universe because `PMF`'s monad instance is single-universe). The adversary outputs a ledger; the arm's witness pair is an oracle-free function of it (`kbPairOf`, the `findPair` duplicate's key witnesses, identified with the reduction's pair by `balanceSubsetOrBreak_kbPair`), so the Balance event is simply "the reduction lands in the key-binding arm". Spend Authority's selection criterion includes the undecidable `Signed`, so its adversary announces a transaction and action index — pointing at its own break — and the machine derives the pair from the indices (`kwAt`, `spendAuthorityOrBreak_pair`); its event is the bundled `SpendAuthorityKBArm`. The pair-recovery functions are censused as computable at the strict flagless tier. What remains for the capstones' ε_kb: the adversary here is an oracle machine rather than the capstones' `PMF` over valid annotated ledgers; that bridge is tracked in #107.

Full build green.

Filed using AI (Claude Fable 5).

### DIFF
--- a/Zcash/Security.lean
+++ b/Zcash/Security.lean
@@ -18,6 +18,7 @@ import Zcash.Security.Ledger.SpendAuthority
 import Zcash.Security.Ledger.Completeness
 import Zcash.Security.Ledger.Capstone
 import Zcash.Security.Ledger.Nullifier
+import Zcash.Security.Ledger.Value
 import Zcash.Security.Common.RandomOracle
 import Zcash.Security.Common.Birthday
 import Zcash.Security.KeyBinding.Basic
--- a/Zcash/Security.lean
+++ b/Zcash/Security.lean
@@ -16,6 +16,7 @@ import Zcash.Security.Ledger.Balance
 import Zcash.Security.Ledger.Spendability
 import Zcash.Security.Ledger.SpendAuthority
 import Zcash.Security.Ledger.Completeness
+import Zcash.Security.Ledger.Capstone
 import Zcash.Security.Common.RandomOracle
 import Zcash.Security.Common.Birthday
 import Zcash.Security.KeyBinding.Basic
--- a/Zcash/Security.lean
+++ b/Zcash/Security.lean
@@ -19,6 +19,7 @@ import Zcash.Security.Ledger.Completeness
 import Zcash.Security.Ledger.Capstone
 import Zcash.Security.Ledger.Nullifier
 import Zcash.Security.Ledger.Value
+import Zcash.Security.Ledger.KeyBindingArm
 import Zcash.Security.Common.RandomOracle
 import Zcash.Security.Common.Birthday
 import Zcash.Security.KeyBinding.Basic
--- a/Zcash/Security.lean
+++ b/Zcash/Security.lean
@@ -17,6 +17,7 @@ import Zcash.Security.Ledger.Spendability
 import Zcash.Security.Ledger.SpendAuthority
 import Zcash.Security.Ledger.Completeness
 import Zcash.Security.Ledger.Capstone
+import Zcash.Security.Ledger.Nullifier
 import Zcash.Security.Common.RandomOracle
 import Zcash.Security.Common.Birthday
 import Zcash.Security.KeyBinding.Basic
--- a/Zcash/Security.lean
+++ b/Zcash/Security.lean
@@ -15,6 +15,7 @@ import Zcash.Security.Ledger.Effects
 import Zcash.Security.Ledger.Balance
 import Zcash.Security.Ledger.Spendability
 import Zcash.Security.Ledger.SpendAuthority
+import Zcash.Security.Ledger.Completeness
 import Zcash.Security.Common.RandomOracle
 import Zcash.Security.Common.Birthday
 import Zcash.Security.KeyBinding.Basic
--- a/Zcash/Security/KeyBinding/Instance.lean
+++ b/Zcash/Security/KeyBinding/Instance.lean
@@ -17,7 +17,11 @@ namespace Zcash.Security
 
 open Zcash.Security.KeyBinding Zcash.Security.Ledger Zcash.Security.RandomOracle Zcash.Snark
 
-variable {G IVK AK NK RIVK ASK QK SK : Type*}
+-- A shared universe for the oracle-model types, so the sampling experiment can be
+-- written in `do`-notation: `PMF`'s monad instance fixes one universe for the whole
+-- block, which independent `Type*` universes could not satisfy.
+universe u
+variable {G IVK AK NK RIVK ASK QK SK : Type u}
 
 /-- The concrete key-binding witness, viewed through the games' `KeyBindingInterface`. -/
 def KeyBinding.toInterface
@@ -36,6 +40,25 @@ def KeyBinding.toInterface
     ⟨h₁, h₂, hivk, fun heq => hne ((Extract.toAK_pm _ _).mp (congrArg BreakProj.ak heq))⟩
 
 
+/-- **The key-binding sampling experiment.** Draw the adversary's index `j` from `p`,
+the two key tables uniformly, and the three components of the combined `rivk` oracle
+uniformly and independently; the sample is `(j, Hask, Hnk, O)` with
+`O = FinalQuery.eval Hleg Hext Hint`. `toInterface_break_measure_le` and the ledger
+arm theorems are stated over this one experiment. -/
+noncomputable def kbExperiment {ι : Type u}
+    [Fintype AK] [Fintype NK] [Fintype RIVK] [Fintype ASK] [Fintype QK] [Fintype SK]
+    [Nonempty NK] [Nonempty ASK] [Nonempty RIVK]
+    [DecidableEq AK] [DecidableEq NK] [DecidableEq RIVK] [DecidableEq QK] [DecidableEq SK]
+    (p : PMF ι) :
+    PMF (ι × (SK → ASK) × (SK → NK) × (FinalQuery AK NK RIVK QK SK → RIVK)) := do
+  let j ← p
+  let Hask ← PMF.uniformOfFintype (SK → ASK)
+  let Hnk ← PMF.uniformOfFintype (SK → NK)
+  let Hleg ← PMF.uniformOfFintype (SK → RIVK)
+  let Hext ← PMF.uniformOfFintype (QK → AK → NK → RIVK)
+  let Hint ← PMF.uniformOfFintype (RIVK → AK → NK → RIVK)
+  pure (j, Hask, Hnk, FinalQuery.eval Hleg Hext Hint)
+
 /-- The probabilistic key-binding bound, delivered at the games' interface: over the
 adversary's private randomness (any distribution `p`) and the five independent oracle
 tables, a bounded-query machine — receiving the sampled `H^ask`/`H^nk` tables and querying
@@ -43,7 +66,7 @@ the combined rivk table — has output witnesses exhibiting the interface's `Bre
 probability at most `(n+4)·(n+3)/|F|`. `toInterface` interprets `Break` as
 `KeyBinding.Break` definitionally, so the games' `∨ kv.Break` branches inherit the bound
 directly. -/
-theorem toInterface_break_measure_le {ι : Type*}
+theorem toInterface_break_measure_le {ι : Type u}
     [AddCommGroup G] [Field IVK] [Field RIVK] [Module RIVK G] [NoZeroSMulDivisors RIVK G]
     [SMul ASK G] [Fintype AK] [Fintype NK] [Fintype RIVK] [Fintype ASK] [Fintype QK]
     [Fintype SK] [Nonempty NK] [Nonempty ASK]
@@ -52,12 +75,7 @@ theorem toInterface_break_measure_le {ι : Type*}
     {A : ι → (SK → ASK) → (SK → NK) → OracleComp (FinalQuery AK NK RIVK QK SK) RIVK
       (KeyBinding.Witness G IVK AK NK RIVK QK SK × KeyBinding.Witness G IVK AK NK RIVK QK SK)}
     {n : ℕ} (hQ : ∀ i Hask Hnk, (A i Hask Hnk).QueryBound n) :
-    ((p.bind fun i => (PMF.uniformOfFintype (SK → ASK)).bind fun Hask =>
-        (PMF.uniformOfFintype (SK → NK)).bind fun Hnk =>
-          (PMF.uniformOfFintype (SK → RIVK)).bind fun Hleg =>
-            (PMF.uniformOfFintype (QK → AK → NK → RIVK)).bind fun Hext =>
-              (PMF.uniformOfFintype (RIVK → AK → NK → RIVK)).map fun Hint =>
-                (i, Hask, Hnk, FinalQuery.eval Hleg Hext Hint))).toOuterMeasure
+    ((kbExperiment p)).toOuterMeasure
         (setOf fun (i, Hask, Hnk, O) =>
           (KeyBinding.toInterface Extract S hfn Ggen
               ⟨Hask, Hnk, fun sk => O (.legacy sk),
--- a/Zcash/Security/Ledger/Balance.lean
+++ b/Zcash/Security/Ledger/Balance.lean
@@ -332,6 +332,79 @@ def balanceSubsetOrBreak [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
             obtain ⟨a, haL, rfl⟩ := List.mem_map.mp (by simpa using hx)
             simpa using hall a haL)
 
+/-- The key-binding arm's witness pair, when the break lies in that arm. -/
+def BalanceBreak.kbPair : BalanceBreak P kv → Option (KW × KW)
+  | .keyBinding w₁ w₂ _ => some (w₁, w₂)
+  | .merkle _ => none
+  | .noteCommit _ => none
+
+/-- The key-binding pair read directly off the ledger: the two key witnesses of the
+duplicate positioned opening that `findPair` locates. Oracle-free — it consumes no
+validity proof — so an oracle machine can output it. `balanceSubsetOrBreak_kbPair` shows
+it is exactly the pair of whatever break the reduction computes. -/
+def kbPairOf [DecidableEq F] [DecidableEq G] [DecidableEq RHO] [DecidableEq PSI]
+    (ledger : Ledger KW F G RHO PSI MHASH MENC MSG SIG P.depth) (i : ℕ) :
+    Option (KW × KW) :=
+  (findPair spendRecord (spendActions ledger (i + 1))).map fun p => (p.1.w.kw, p.2.w.kw)
+
+/-- Per-spend pinning never computes a key-binding break: its breaks are the
+note-commitment collision (`hop` fails) and the Merkle collision (`hleaf` fails). -/
+theorem spendPinnedOrBreak_kbPair [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
+    [DecidableEq PSI] [DecidableEq MHASH] [DecidableEq MENC]
+    (hval : ValidLedger P kv issuance maxActions ledger) {i : ℕ}
+    {a : Action KW F G RHO PSI MHASH MENC SIG P.depth}
+    (ha : a ∈ spendActions ledger (i + 1)) {b : BalanceBreak P kv} :
+    spendPinnedOrBreak hval ha = .inr b → b.kbPair = none := by
+  revert b
+  fun_cases spendPinnedOrBreak hval ha <;>
+    intro b hb <;>
+    (try simp_all only [reduceCtorEq, PSum.inr.injEq, BalanceBreak.kbPair]) <;>
+    (subst hb; rfl)
+
+/-- The pinning scan never computes a key-binding break (`spendPinnedOrBreak_kbPair`,
+propagated through the recursion). -/
+theorem allPinnedOrBreak_kbPair [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
+    [DecidableEq PSI] [DecidableEq MHASH] [DecidableEq MENC]
+    (hval : ValidLedger P kv issuance maxActions ledger) {i : ℕ} :
+    ∀ (L : List (Action KW F G RHO PSI MHASH MENC SIG P.depth))
+      (hL : ∀ a ∈ L, a ∈ spendActions ledger (i + 1)) {b : BalanceBreak P kv},
+      allPinnedOrBreak hval L hL = .inr b → b.kbPair = none := by
+  intro L
+  induction L with
+  | nil => intro hL b hb; simp only [allPinnedOrBreak, reduceCtorEq] at hb
+  | cons a t ih =>
+      intro hL b hb
+      rw [allPinnedOrBreak] at hb
+      split at hb
+      · rename_i b' hsp
+        rw [PSum.inr.injEq] at hb; subst hb
+        exact spendPinnedOrBreak_kbPair hval _ hsp
+      · split at hb
+        · rename_i b' hall
+          rw [PSum.inr.injEq] at hb; subst hb
+          exact ih (fun x hx => hL x (by simp [hx])) hall
+        · simp only [reduceCtorEq] at hb
+
+/-- **Key-binding-arm localization.** Whatever break `balanceSubsetOrBreak` computes, its
+key-binding pair is the oracle-free `kbPairOf`: a key-binding break arises only from the
+`findPair` duplicate, whose pair is `kbPairOf`, and every other break has no pair. This is
+what lets an oracle machine output the arm's pair without running the reduction. -/
+theorem balanceSubsetOrBreak_kbPair [DecidableEq F] [DecidableEq G] [DecidableEq RHO]
+    [DecidableEq PSI] [DecidableEq MHASH] [DecidableEq MENC] [DecidableEq NK]
+    [NoZeroSMulDivisors F G]
+    (hval : ValidLedger P kv issuance maxActions ledger) (i : ℕ) {b : BalanceBreak P kv} :
+    balanceSubsetOrBreak hval i = .inr b → b.kbPair = kbPairOf ledger i := by
+  revert b
+  fun_cases balanceSubsetOrBreak hval i <;> intro b hb <;>
+    (try (rw [PSum.inr.injEq] at hb; subst hb)) <;>
+    (try simp_all only [reduceCtorEq])
+  · -- key-binding break from a `findPair` duplicate: its pair is `kbPairOf`
+    simp_all [BalanceBreak.kbPair, kbPairOf]
+  · -- no duplicate: the pinning scan's break has no pair, and `kbPairOf` is `none`
+    rename_i hfpNone _ _ hallp
+    have hnone := allPinnedOrBreak_kbPair hval _ _ hallp
+    simp [kbPairOf, hfpNone, hnone]
+
 end Validity
 
 /-! ## Balance-value, in conservation form
--- a/Zcash/Security/Ledger/Bridge.lean
+++ b/Zcash/Security/Ledger/Bridge.lean
@@ -162,13 +162,13 @@ private theorem commitIvkHash_of_action_hash {ak nk : Fp} {bp : Point Fp}
 
 /-- The circuit's randomized-key public coordinates are the affine view of the
 concrete ledger randomization. -/
-theorem public_rk_eq_randomizePublic (verify : PallasGroup → MSG → SIG → Prop)
+theorem public_rk_eq_randomizePublic (verify bverify : PallasGroup → MSG → SIG → Prop)
     {wit : ActionData}
     (hak : wit.akP.OnCurve)
     (h : ({ x := wit.rkX, y := wit.rkY } : Point Fp) =
       wit.alpha.2 • orchardBases.spendAuthG + wit.akP) :
     PallasGroup.toPoint
-      ((Pool.primitives verify).randomizePublic wit.alpha.2
+      ((Pool.primitives verify bverify).randomizePublic wit.alpha.2
         (PallasGroup.ofPoint wit.akP (.inl hak))) =
       ({ x := wit.rkX, y := wit.rkY } : Point Fp) := by
   change PallasGroup.toPoint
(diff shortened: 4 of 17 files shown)


daira (commented):
Self-ACK

(To be clear, this means: "ready for wider review". I should just say that.)

TalDerei (commented):
# Audit: spec bugs, implementation correctness, soundness holes

**TL;DR: nothing proven here is wrong** — every new theorem states what it claims, and the algebra of each new reduction re-derives correctly against the deployed Orchard spec. The findings are all at the *seams*. Two of them block composition: the key-binding "ε discharge" bounds a neighbouring event rather than the capstone's `εkb`, with no bridge and no obvious way to build one; and the nullifier-to-SDLR reducer is unreachable from the games that produce collisions, so the security map's new edge asserts a composition that doesn't exist yet. That second one is closable in-branch by a one-field strengthening, which would retire #107's ρ-uniqueness residual outright rather than deferring it. Separately, the PR body names a theorem that isn't in the branch.

## Findings

### 1. High — the key-binding arm's ε bounds the wrong event

`balanceSubset_measure_le` (`Capstone.lean:195`) takes `εkb` as a bound on `balanceSubsetBreakEvent i .keyBinding` (`Capstone.lean:167`) — "the reduction lands in the key-binding arm, with *any* pair" — for an adversary of type `PMF (ValidAnnotated …)`. What `balanceSubset_keyBindingArm_measure_le` (`KeyBindingArm.lean:76`) proves lives in a different probability space (the oracle-table experiment) and bounds a *different* event: the reduction lands in the arm **and its pair equals the adversary's own annotated output pair**. The mismatch cuts both ways, and neither direction is repairable with what's in the branch:

- **False negatives.** A wrong annotation makes the oracle event false even when the sampled ledger genuinely breaks key binding, so the bounded event is strictly smaller than the arm event — no inclusion runs in the direction the capstone needs.
- **Union over annotations is vacuous.** Recovering a ledger-only bound by unioning over constant-pair annotators costs a factor of |KW|², and `KW = KeyBinding.Witness G IVK AK NK RIVK QK SK` carries group elements, so the factor is astronomical and the bound collapses.
- **Conditioning on validity doesn't preserve the query bound.** There's no machinery to turn an oracle machine's output into a `PMF (ValidAnnotated …)` at all, and the obvious route — rejection-sample until the output ledger is valid — re-runs the machine, so the composite no longer satisfies the `QueryBound n` hypothesis `toInterface_break_measure_le` requires. `Capstone.lean:14-18`'s "conditioning on validity can only increase success probability" is prose, not Lean, and this is why it isn't cheap to make formal.

So `balanceSubset_measure_le`'s `hkb` remains undischarged. The module doc discloses the pair annotation and the dependent-type obstruction that forces it, but not that this changes *which event* is bounded. The PR body ("discharges the Balance-subset key-binding arm's ε") and #107's table ("Key-binding probability bound → game-level statements — **done** in #108") both overstate what landed. Honest framing: a key-binding bound for pair-annotated ledger adversaries now exists in the oracle model; bridging it to the capstone's `εkb` is an open and currently untracked obligation. The plausible shape is running the pair-*finding* half of the reduction as an oracle machine with a charged query count — evaluating `kv.ivk` over the ledger's witnesses is what actually costs queries, not the validity check.

### 2. High — the PR body names a theorem that doesn't exist

The body says "The Spend Authority key-binding arm is discharged the same way (`spendAuthority_keyBindingArm_measure_le`)". That declaration appears nowhere on the branch, and `KeyBindingArm.lean`'s own module doc says the opposite: "The same pattern fits Spend Authority's key-binding arm; **only Balance-subset is discharged here**." #107's most recent edit repeats the error ("**both** key-binding arms' ε are formalized there"). Either correct the body and #107, or add the theorem — and note that per finding 1 it would inherit the same bridge gap, so adding it doesn't discharge `spendAuthority_measure_le`'s `εkb` either.

### 3. High — the nullifier-to-SDLR reducer is unreachable, and the security map's edge overstates what composes

`NontrivialRelation.ofNullifierCollision` (`Nullifier.lean:147`) has **zero callers**: the only occurrences in the tree are its own definition, its docstring, and the `TrustBoundary` pin. The blocker is a hypothesis mismatch. `NullifierCollision.ne` (`Spendability.lean:77`) says only that the derivation inputs `(nk, ρ, ψ, cm)` differ, which does not imply distinct notes; the reducer requires `note₁ ≠ note₂`. So `spendabilityOrBreak` (`Completeness.lean:429`) and `faerieGoldCore` can still return a collision that cannot be routed onward. The premiss itself is disclosed — `Nullifier.lean:34-39` calls it a consumer obligation and names ledger-level ρ-uniqueness as the intended discharge, and #107 tracks it — but the book's new `NFC --> SDLR` edge is not hedged to match: it asserts that `NullifierCollision` rests on the Sinsemilla DL relation when no path from a game-produced collision to that reduction exists today. The edge label ("distinct-note openings compute") is literally true *about the reducer*, but an edge in a trust graph claims composition, not the existence of a lemma. Either mark it planned, or close the gap — which is cheap:

**The premiss is avoidable.** The sole producer, `faerieGoldCore` (`Spendability.lean:85`), already holds the fact needed at construction time — `(rcm₁, note₁) ≠ (rcm₂, note₂)` — and simply doesn't record it. The reduction goes through from that opening inequality alone:

- **notes distinct** — exactly the two sign cases already written (`coeff_sub_ne` / `coeff_add_ne`);
- **notes equal, `rcm` distinct** — write `cmᵢ = ∑ cf • bases + rcmᵢ • R` with the *same* `cf`. The `+` case gives `(s₁ − s₂) • K + (rcm₁ − rcm₂) • R = 0`, nontrivial via `β = rcm₁ − rcm₂ ≠ 0`; the `−` case gives `a = cf + cf ≠ 0` by `coeff_add_ne`, needing no distinctness at all.

The one genuinely irreducible case — equal notes *and* equal `rcm`, differing only on `nk`, i.e. the PRF-collision case the Orchard design page discusses — is already excluded by the opening inequality, since equal openings pin `cm` and contradict `faerieGoldCore`'s `hne`. Every case split is decidable under the `DecidableEq` instances `faerieGoldCore` already assumes.

So: add `ne_opening : (rcm₁, note₁) ≠ (rcm₂, note₂)` to `NullifierCollision` (or weaken the reducer's hypothesis to it), wire it at the `faerieGoldCore` call site, and #107's "ρ-uniqueness discharge of the distinct-notes premiss" residual disappears — no collision-to-ledger-outputs composition needed, and the map edge becomes real. `output_rho_nodup` stays valuable on its own terms but comes off the Faerie-Gold critical path.

### 4. High — binding-signature extraction has a non-composable shape, and the deterministic form is vacuous

`hextract : ∀ bvk msg sig, bindingVerify bvk msg sig → bvk = extractBsk bvk msg sig • S.R` (`Value.lean:131`) posits a *total* function recovering `bsk` from every verifying triple with zero failure probability. Two problems, and the second is the sharper one:

- **It can't be proven in that shape.** Real RedDSA forking extraction is adversary- and transcript-dependent, probabilistic, and carries a knowledge error; it does not yield a total function of `(bvk, msg, sig)`. Balance-value needs a probabilistic extractor-failure arm and an error term, not a hypothesis.
- **As stated it is classically satisfiable, hence vacuous.** In any group where `R` generates — i.e. the deployed one — every `bvk` *is* some multiple of `R`, so `Classical.choice` discharges `hextract` outright. `ValueShape.premissOrBreak`'s output is therefore a computational break only relative to an **efficient** `extractBsk`, which nothing in the development supplies or even types (there's no cost model on the function argument). The `assert_computable … +choice` pin does guarantee the reduction conjures nothing itself, but the extractor is an *input*, so that guarantee is only relative to it.

This is #22/#107's documented shape mismatch. The audit-relevant consequence: `balanceValue_measure_le` / `valueConservation_measure_le` cannot yield concrete numbers even in principle until the capstones are restated in extractor-plus-knowledge-error form — the same restructuring #67 is doing on the SNARK side. Correctly disclosed in `Value.lean` and #107; fine as an interface, but consumers shouldn't read the current ε as dischargeable.

### 5. Low — `vbalance_bound` is not the signed-64-bit rule it's labelled as

`vbalance_bound : ∀ tx ∈ ledger, tx.vBalance.natAbs < P.valueBound` (`Model.lean:195`) is documented as "the consensus `valueBalance` range rule", but with the intended `valueBound = 2^64` it admits values such as `2^63` that no int64 can encode. [ZIP 225](https://zips.z.cash/zip-0225) specifies `valueBalanceOrchard` as signed 64-bit, so the real rule is `−2^63 ≤ vBalance ≤ 2^63 − 1` — roughly twice as tight on each side. The looseness is deliberate (the PR body: "reusing `valueBound`, so no new game parameter") and safe, since it only widens the adversary and the conservation inequality doesn't care; `valueBound` is genuinely the right bound for the uint64 note values it also serves. But the conjunct is a *consequence* of the consensus rule, not the rule, and the docstring presents it as the rule. Secondary nit in the same docstring: the parenthetical `|vBalance| ≤ 2^63 − 1` excludes `−2^63`, which int64 does admit.

## Verified sound (the things most likely to be wrong, and aren't)

- **Model changes** — `Tx.bvk = ∑ cv_net − ValueCommit_0(vBalance)` matches spec §4.14 exactly. `binding_verified` is a genuine consensus rule, and both new conjuncts strengthen `ValidLedger` toward deployed consensus while staying honest-satisfiable (`honestTx_valid` carries the two new wallet-checkable boundary hypotheses rather than assuming them away). The append/saturation lemmas are correct.
- **Value discharge arithmetic** — the no-overflow chain `natAbs(net − vBal) ≤ maxActions·B + natAbs(vBal) < (maxActions+1)·B ≤ r` is strict exactly where it must be; `bvk_eq` correctly identifies the model `bvk` with `bindingVK` over the cast bundle with an empty output side (Orchard's single-bundle shape); the deployed instantiation `(2^16)·2^64 ≪ 2^254` is comfortably satisfiable.
- **Nullifier reduction algebra** — both sign cases re-derived by hand. The coefficient conditions are dischargeable at deployed Sinsemilla: fixed message length gives the initial `2^n • 𝒬` term, and per-base coefficients `< 2^n < r` so binary expansions determine the message; `v ≥ 2^64` is handled by `noteCommit` partiality. One instantiation-time note: `extract_pm`'s deployed discharge also needs to cover `Extract(𝒪) = 0` versus a hypothetical `x = 0` point, i.e. the spec's "no Pallas point has x-coordinate 0".
- **Completeness** — `HonestAction.satisfied` routes `path_of_root` through `included` correctly; the dummy-spend construction is faithful (`v = 0`, `pk_d = [ivk] g_d`, inclusion vacuous); `honestTx_valid` covers all nine validity conjuncts with correctly-scoped boundary hypotheses; the anchor indices (`anchor ≤ ledger.length`, new tx at index `length`) respect the strictly-earlier-outputs rule.
- **Spendability capstone** — the `find?`-based roadblock search composes with `respendOrBreak` as claimed (`hs.inst.nf_old = hs.nf` by `rfl`), and the `Respend` arm is exactly the Spend-Authority-covered case.
- **Capstone event algebra** — all three containment lemmas and both union bounds are correct. `ValidLedger` being `Prop`-valued does make the bundled-proof events canonical, as the docs claim, and `ValidAnnotated` is nonempty (empty ledger), so nothing is vacuous.
- **Trust boundary** — all new endpoints are pinned at the right tiers, and the macros do enforce what the file claims (plain-`def` plus axiom bound). Full CI green on head `7035442`; `lake build Zcash.Security Zcash.TrustBoundary` and `git diff --check` both pass.

## Nits

- `Capstone.lean:26-28` ("the key-binding arm against `toInterface_break_measure_le` … not wired here yet") predates the final commit. It remains technically true — per finding 1 nothing is wired — but now reads as stale; worth rewording to point at `KeyBindingArm.lean` and name the remaining bridge explicitly.
- Pre-existing but adjacent: `TrustBoundary.lean`'s "Ledger-layer break reductions" block comment claims the strict flagless tier ("no `Classical.choice` even in erased positions") while the block contains `nfOldEqOrBreak +choice`.

---

*Method: this consolidates two independent audit passes over the diff against merge-base `957dd51`, reading the full diff plus every pre-existing definition it consumes (`Model`, `Statement`, `Effects`, `Balance`, `Spendability`, `SpendAuthority`, `Merkle`, `BindingSignature/Balance`, `KeyBinding/Instance`, `DiscreteLogRelation`, `Meta/AxiomCheck`), re-deriving each new reduction's algebra by hand, checking every probabilistic event's containment lemma, and cross-checking shapes against the deployed Orchard spec (§4.14 `bvk`, ValueCommit/NoteCommit/DeriveNullifier, Sinsemilla coefficient structure) and ZIP 225.*

*This audit was generated with the assistance of Claude Fable 5, via [Claude Code](https://claude.com/claude-code). Treat the findings as leads to verify against the code, not as authority — the file:line references are checkable, and the "verified sound" claims are the ones most worth a second human look.*

TalDerei:
> Self-ACK

the review i've been engaging in suggests this isn't a sufficient and completed security capstone yet.

daira:
Thanks for the review! There are definitely some substantive things that need fixing. 2 was just unpushed; 4 is an unfixed pre-existing modelling bug; 1 is less of a problem than the review says.

> 3. [...] The premiss itself is disclosed — `Nullifier.lean:34-39` calls it a consumer obligation and names ledger-level ρ-uniqueness as the intended discharge, and #107 tracks it — but the book's new NFC --> SDLR edge is not hedged to match: it asserts that NullifierCollision rests on the Sinsemilla DL relation when no path from a game-produced collision to that reduction exists today. The edge label ("distinct-note openings compute") is literally true about the reducer, but an edge in a trust graph claims composition, not the existence of a lemma.

We have different ideas about what the graph means. SDLR is red! That means "not done yet". The edge label is what is done. You can't compose with a thing that isn't formalized yet. And the graph simply can't capture all the detail about what precisely doesn't (yet) compose in that case; all it can do is colour the target red.

> NullifierCollision.ne (`Spendability.lean:77`) says only that the derivation inputs (nk, ρ, ψ, cm) differ, which does not imply distinct notes; [...]

Huh? It does trivially imply distinct notes; that's a one-liner.

daira:
Oops, I had not pushed the commit that fixes 2. Pushed now.

daira:
[Claude, reviewed by me.] Thanks — this is a genuinely useful audit, and the "verified sound" section is the part I'd most want someone else to confirm. Disposition of the findings:

**Finding 3 (unreachable reducer) — fixed** (`be5cb12`). Your case analysis was exactly right: `faerieGoldCore` already holds `(rcm₁, note₁) ≠ (rcm₂, note₂)` at construction and discarded it. I strengthened `NullifierCollision.ne` to record that opening inequality (replacing the derive-input inequality, which the reducer never consumed), wired it at the call site, and dropped `ofNullifierCollision`'s premiss. The `+` case splits on note equality — distinct notes give `coeff_sub_ne`, equal notes force `rcm₁ ≠ rcm₂` (nonzero `β`); the `−` case needs no distinctness. The irreducible equal-note/equal-`rcm`/distinct-`nk` case can't arise, since equal openings pin `cm` and the spends coincide. So the reducer is reachable, the map edge is honest, and this retires #72's ρ-uniqueness residual outright (`output_rho_nodup` stays useful but comes off the Faerie-Gold critical path). #107 updated.

**Finding 2 (missing theorem) — already resolved, you audited one commit too early.** `spendAuthority_keyBindingArm_measure_le`, the `KeyBindingArm.lean` module-doc sentence, and the census entry all landed in `bc2adad`, pushed after your audit's `7035442` head. [Daira-Emma: I pushed before posting this.] Your rider stands: it inherits finding 1's gap.

**Finding 1 (wrong event) — you're right, and I'm correcting the framing; one factual amendment.** Agreed: `balanceSubset_keyBindingArm_measure_le` bounds the pair-annotated event (reduction lands in the arm **and** its pair equals the annotation), not the capstone's any-pair `εkb`, and "discharges the ε" in the body / "done" in #107 overstate that. I'm correcting all three to your honest framing — a key-binding bound for pair-annotated adversaries exists in the oracle model; bridging it to `εkb` is open — and tracking the bridge.

One correction that makes the bridge more tractable than "no obvious way": **the key-binding pair is an oracle-free total function of the ledger.** The interface's `ivk`/`nk`/`akP` are plain projections of `Witness` (`toInterface` sets `nk w := w.nk`), so evaluating them costs no queries, and the arm arises exactly on the decidable branch "`findPair` finds a duplicate positioned opening with distinct `nk`". So a total `kbPairOf : Ledger → Option (KW × KW)` exists, the composite `LA.bind (pure ∘ kbPairOf)` lets the adversary output only a ledger, and the bounded event becomes the honest any-pair arm event. That upgrade is the right next step (it needs a localization lemma through `balanceSubsetOrBreak`'s match spine, plus `allPinnedOrBreak`-never-keyBinding). What stays genuinely open is your other point — turning an oracle machine into a `PMF (ValidAnnotated …)` without breaking the `QueryBound`; that's the named-hypothesis gap, now tracked, not something I'm claiming to close.

**Finding 4 (extraction shape/vacuity) — correct and sharp; documenting, not fixing.** Right on both: the total `hextract` is classically satisfiable (every `bvk` is a multiple of a generating `R`), so `premissOrBreak` has computational content only relative to an *efficient* `extractBsk`, which nothing types or supplies. This is the #22/#107 extractor-plus-knowledge-error shape mismatch (the same restructuring #67 is doing SNARK-side). I'm hardening the `Value.lean` docstring to say exactly that — a placeholder for the forking extractor-with-error form, not a dischargeable hypothesis — and leaving the restructuring tracked.

**Finding 5 + nits — fixing.** On the range rule: the actual Orchard consensus rule is `valueBalance ∈ [−MAX_MONEY, MAX_MONEY]` (MAX_MONEY ≈ 2⁵¹, even tighter than ZIP 225's signed-64-bit), so `vbalance_bound : natAbs < valueBound` is a loose-but-correct *consequence*, and the `−2^63` nit is moot. Rewording the docstring to say so. Also fixing the stale `Capstone.lean` module-doc sentence (pointing it at `KeyBindingArm.lean` and naming the bridge) and the pre-existing `TrustBoundary` "strict flagless" block comment that contradicts its own `nfOldEqOrBreak +choice`.

The finding-4/5/nits and the finding-1 reframe land as follow-up commits on the branch; the `kbPairOf` any-pair upgrade I'll take as its own piece so it gets interactive proof iteration rather than a blind pass.

daira:
Status update on the audit findings:

**Landed** (ebe0db9, pushed): the finding-4 and finding-5 documentation, and both nits. `Value.lean` now says plainly that `hextract` is a placeholder — classically satisfiable in total form, with computational content only relative to an efficient `extractBsk` (#22/#107 track the forking extractor-with-knowledge-error form). `Model.lean`'s `vbalance_bound` cites the actual rule (§7.1.2, [NU5 onward]: valueBalanceOrchard ∈ {-MAX_MONEY .. MAX_MONEY}) and frames the conjunct as its loose consequence. The stale `Capstone.lean` module-doc sentence and the `TrustBoundary` "strict flagless" comment are fixed. (Also: the finding-3 commit is 30f4d71 — be5cb12 in my earlier comment was superseded by a message amend.)

**Reframed**: the PR description and #107 no longer claim the key-binding arms' ε is discharged; they state the oracle-model bound as partial, with the gap to the capstones' ε_kb tracked in #107.

**Plan for finding 1**: next on the branch, as its own piece. First a named sampling experiment (do-notation) so the arm statements stop spelling out the nested `uniformOfFintype` chain; then the upgrade sketched in my earlier comment — the adversary outputs only a ledger, the pair is computed from it, and the event becomes the reduction's own arm event, via a localization lemma through `balanceSubsetOrBreak`. The oracle-machine → `PMF (ValidAnnotated ...)` bridge stays a named gap in #107; this PR does not claim to close it.

TalDerei:
> Status update on the audit findings:

should we just fold this into #107 so follow-ups are self-contained in one place? if so, can you update the existing issue comment?

TalDerei (approved):
ACK

daira:
Rebased on main; ready for re-review. Only the last three commits are substantially new.

daira (commented):
Self-ACK

daira:
Following up on finding 1: the rebuild landed before the merge. Both key-binding arm theorems now bound the reductions' own events — the arm's witness pair is recovered from the adversary's output by an oracle-free function (`kbPairOf` from the ledger for Balance; `kwAt` at announced indices for Spend Authority, whose selection criterion includes the undecidable `Signed`), identified with the reduction's pair by localization lemmas, and the pair recovery is censused as computable at the strict flagless tier. The remaining gap to the capstones' ε_kb — the oracle-machine → `PMF` bridge — is tracked in #107.

TalDerei:
This can fold into PR that closes both #107 / #22. 

### The nullifier reduction omits required value bounds, preventing concrete instantiation

`NullifierCollision` allows arbitrary `Note` values, while `NullifierShape.coeff_sub_ne` requires coefficient injectivity for every pair of distinct notes. However, the concrete `Pool.noteScalars` implementation casts the unbounded `Note.v : Nat` into `Fp`.

Consequently, two notes that are identical except for values differing by the Pallas base-field modulus are distinct as `Note`s but have identical scalar encodings. Thus `coeff_sub_ne` cannot hold globally for the concrete implementation, and the abstract nullifier reduction cannot be instantiated with `Pool.noteScalars` as currently typed.

This is not a protocol-level collision: valid action witnesses already constrain note values. The missing bounds need to cross the reduction boundary.

Suggested fix:

- Add bounds for both note values to `NullifierCollision`.
- Populate them from the existing `ActionSatisfied.v_old_lt` witnesses in `faerieGoldCore`.
- Scope `coeff_sub_ne` to notes satisfying those bounds.

This mirrors the explicit value-bound treatment already used for `NoteCommitBreak`.

daira:
Yes that's known and part of #107.

Issue #86, opened 2026-07-22
Public-instance soundness


TalDerei:
References: https://github.com/zcash/ironwood/issues/65.

Sound handling of public instances = three steps that must land in order.

1. [x] **Derivation — #85 (this PR).** Public inputs into Lean + `commitLagrange` + `instance_commitments_derived` + regenerated fixtures. 
2. [x] **Decode gap — #30 (`decode-witness`, closes #18).** Binds the extracted IPA witness to the verifier's opened columns. Already binds `decodeInstance` *abstractly* to whatever `C_inst` opens to; the concrete work is advice-scoped.
3. [x] **Instance binding — follow-up (after 1 + 2).** Identify that abstract instance column with the public inputs: `decodeInstance`'s recovered column `= commitLagrange(publicInputs)`. Small — reuses #30's `batch_open_soundV` and #85's derivation; the easy, known-polynomial half of the decode gap.

PR #113, opened 2026-07-26
feat(fv): join the captured instance commitments to the deployed capstone

TalDerei:
Merges after #114 (its base).

Contributes the last step of https://github.com/zcash/ironwood/issues/86. The binding argument itself is already on `main` (#99): `instanceOpening_eq_or_relation` through `publicInputEncoding_or_relation` identifies a routed instance column with the verifier's public inputs, or computes a `(g, U, W)` relation.

-------

**AI Summary:**

What was missing was not the argument but a join. The deployed Action capstone reads its instance commitments from `actionCircuit.instanceCommitment`, computing them from the public inputs the way halo2's `verify_proof` computes them from `instances`. The captured artifacts — the Fiat–Shamir fingerprint, `assemble?`, the negative tests — run against `Fixture.derivedInstanceCommitment`, which `instance_commitments_derived` pins to the captured points (#65/#85). Nothing tied the two families together, so the capstone could not be instantiated at the captured proof.

`Keygen/InstanceCapture.lean` ties them, in two steps:

- **The bases agree.** `commitLagrange_eq_commitInstance` identifies halo2's captured Lagrange computation with the abstract `LagrangeCommitmentKey.commitInstance` at `capturedURS`. No new fixture check: `derivedUrsGLagrange_prefix_eq` already certifies the exported prefix against the monomial derivation, and `LagrangeCommitmentKey` is now a `Subsingleton` — `generator_eq` pins every generator — so the key a statement names is the key the certificate discharges.
- **The rows agree.** `capturedActionInputs` reads the capture back as the circuit's own `PublicInputs` record. The Action layout has `usedRows = 10`, so `publicInputRows input ⟨0⟩` is exactly the element vector, with no circuit elaboration; only that ten-element serialization spends `native_decide`.

`instanceCommitment_capturedActionInputs` is the resulting family equality, and `instanceCommitment_eq_capturedInstanceCommitments` composes it with #85 to reach the captured points.

**Threaded into the capstone.** `action_bundleStatement_or_relation_of_deployedAccepts_family` restates the deployed capstone over a supplied commitment family plus an equation identifying it, discharged by `subst` — the same theorem, stated so a caller may use its own vocabulary. `action_bundleStatement_or_relation_at_capturedInstances` instantiates it: acceptance stated at `Fixture.derivedInstanceCommitment`, concluding `BundleStatement capturedActionInputs ⊕' NontrivialRelation`.

## Review notes

- Transporting the capstone by `▸` instead of the `_family` restatement was tried and rejected: building the motive over that ~150-line type exhausts the default heartbeat budget even though the pattern's head is already `local irreducible`. Raising the budget would hide the cost, not remove it.
- The new `native_decide` carries no `assert_axioms +native` entry, matching `Keygen/`'s existing convention (`Keygen.certificate` is `assert_no_sorry` only). Adding parity with the `Fixtures/` lane is a scoping call.
- This PR previously carried a parallel binding layer of its own (`Soundness/InstanceBinding.lean`, `Fixtures/InstanceWitness.lean`). It was dropped: #99's route proves the same identification from a weaker premise and is wired to the capstone, while those declarations had no consumer.

### DIFF
--- a/Zcash/Snark/Soundness/Deployed/ActionVesta.lean
+++ b/Zcash/Snark/Soundness/Deployed/ActionVesta.lean
@@ -1,6 +1,7 @@
 import Zcash.Snark.Soundness.ActionVesta
 import Zcash.Circuits.Integration.ActionTerminal
 import Zcash.Snark.Keygen.Certificate
+import Zcash.Snark.Keygen.InstanceCapture
 import Zcash.Snark.Soundness.Canonical.Vesta
 import Zcash.Snark.Soundness.Multiopen.CanonicalSelection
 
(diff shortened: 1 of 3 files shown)


Zcash/Snark/Soundness/Deployed/ActionVesta.lean
```diff
@@ -398,6 +399,202 @@ noncomputable def action_bundleStatement_or_relation_of_deployedAccepts
...
+the captured family. This is the last step of ironwood#86: sound handling of public instances now
+reaches the deployed artifacts, not merely a derived commitment family the capture was never tied to.
+
+Acceptance is stated at `Fixture.derivedInstanceCommitment`; the conclusion is
+`BundleStatement Keygen.capturedActionInputs ⊕' NontrivialRelation …`.
+-/
+noncomputable def action_bundleStatement_or_relation_at_capturedInstances :=
+  action_bundleStatement_or_relation_of_deployedAccepts_family
+    Keygen.capturedActionInputs Fixture.derivedInstanceCommitment
+    Keygen.instanceCommitment_capturedActionInputs.symm
```
mitschabaude: if this is our ultimate theorem, it should have a type annotation, otherwise it's not readable

TalDerei: good point, addressed.

PR #114, opened 2026-07-26
feat(fv): AGM extractor follow-ups

TalDerei:
**Human Summary:**

Follow-up to [#138](https://github.com/zcash/ironwood/pull/138). Recall that #138 connects the extractor to the real Orchard circuit and composes a final adversary bound (~123-bit protocol work-factor target, conditional on Vesta DLOG hardness, in the AGM/random-oracle model), using straight-line AGM extraction rather than special soundness.

This PR introduces a few improvements:

1. Surgically changes integration-layer definitions that were marked `noncomputable`, and makes the Action bridge free of `Classical.choice`. It emulates the needed polynomial layer lifted and adapted from Mathlib (where `Polynomial` operations are noncomputable); a follow-up can use [CompPoly](https://github.com/Verified-zkEVM/CompPoly).
2. Generalizes the random-oracle adversary model: the capstone now quantifies over an arbitrary adaptive represented online-AGM prover rather than the more restrictive phase-equipped construction.
3. Generalizes the concrete Action capstone from `actionProofParams` (`numProofs = 1`) to every `numProofs ≤ orchardConsensusMaxProofs`, so `BundleStatement` quantifies over every Action in a consensus-valid bundle.

--------

**AI Summary:**

## What this proves

`Fixture.orchard_action_acceptFalseStatement_adaptive_2pow123_workFactor_generatorRO_for`, for every `numProofs ≤ orchardConsensusMaxProofs` and every adaptive represented online-AGM prover making at most `2^123` oracle queries:

> `Pr[deployed Action verifier accepts ∧ BundleStatement false] ≤ Adv_DLOG(2^127, 2^127) + 2^-83`

The event is plain acceptance of a false statement. There is no “or a discrete-log relation exists” escape branch, which at prime order would make the claim vacuous. The theorem also proves the envelope it quotes: at most `2^127` oracle queries, `2^127` group work, and `2^123` direct-coordinate operations. Its final conjunct transports the ideal generator-random-oracle event bound to an actual transcript experiment under an explicit `PMFEventBiasLE` premise, adding that conversion-bias budget. `Adv_DLOG` is a caller-supplied premise, so this is a conditional bound, not a proved security exponent.

The single-Action compatibility theorem remains available with its tighter `2^-84` statistical remainder.

## Adversary model

The primary endpoint quantifies over `ComputedAdaptiveOnlineAGMFSFamily`: adaptive, malicious oracle queries, with the algebraic trace and all five challenge surfaces derived from the run itself. Nothing is supplied from outside—no phases, traces, cuts, or views. That covers every prover in the represented online-AGM model, not every non-algebraic random-oracle adversary.

The sequential route remains as a second interface, and now builds the cuts it consumes instead of assuming them.

## Cost accounting

Each provenance aggregate computes the adversary output and annotation log once, then reuses that retained log for all of its lookups: five Action checks, six pre-IPA checks, and `k = 11` IPA-round checks. Cached lookup performs no oracle calls, and the retained log length is proved pointwise bounded by `Q`. No cache sharing between independently charged finder stages is assumed; every charged traversal pays its own `11+k` transcript reads.

| finder | charged traversals | oracle queries |
|---|---:|---:|
| straight-line constraint | 4 | `4Q + 4(11+k)` |
| sequential Action | 6 | `6Q + 6(11+k)` |
| bare adaptive Action | 8 | `8Q + 8(11+k)` |

Queries, group work, and direct-coordinate work are tracked separately: an `8T` envelope for sequential and a deliberately loose `16T` envelope for adaptive. The recursive extractor survives only as a separately priced alternative, at AFK-style `(8Q+1)·10^k`—roughly a `2^39.5` factor on `Q` at `k = 11`.


daira:
This is failing book checks.

Zcash/Snark/Fixtures/MaxShapeBounds.lean
```diff
@@ -92,30 +135,62 @@ theorem consensusPinnedRootMultiopenModel_at_2pow122 :
...
+field solves.
+
+`T = 2^122` cannot also be used as the combined finder's call budget.  At `k = 11`, the proved
+AFK expectation is between `2^161` and `2^162` calls.  Making the generic Markov tail at most
+`2^-86` would require `L+1` at least that expectation times `2^86`, a number between `2^247` and
+`2^248`.  These facts correct the old `t ≈ T` interpretation; they do not claim that such a large
+budget gives a useful concrete DLOG advantage.
+
+The combined finder, not the raw adversary, is the DLOG solver.  Its unconditional expected
+black-box call bound is `afkRunBound Q 11 + 3`: the AFK recursive bound plus the direct-coordinate
```
daira: Expand "AFK". Non-blocking.

TalDerei:
> This is failing book checks.

removed my review because still iterating, #96 is ready though.

Zcash/Snark/Fixtures/StraightLineMaxShapeBounds.lean
```diff
@@ -0,0 +1,203 @@
...
+      (straightLineConstraintFailure_prob_le_at_consensus_max_generatorRO
+        B hB query hquery family static schedule profile.toStraightLineConstraintDlogProfile
+        profile.queryBound) ?_
+    exact add_le_add
+      (profile.advantage_mono hcost.1 hcost.2)
+      consensusStraightLineStatisticalModel_at_2pow122
+  · exact profile.solverCost_at_2pow122
+
+/-- The final work-factor arithmetic used in prose: an explicit five-bit overhead takes the
+`2^122` adversary target exactly to the `2^127` DLOG-solver scale. -/
```
daira: The expected Pollard rho security of Pallas/Vesta is ~ $2^{126.0}$, as I calculated for my Zcon3 presentation. (This is partly because of the endomorphisms.) So one extra bit should be subtracted.

Zcash/Snark/Soundness/Forking/Oracle.lean
```diff
@@ -26,28 +26,63 @@ from one assumption:
...
+assumed away. The Fiat–Shamir layer also produces `DeployedAlgebraicForkingInstance` values and
+relates its acceptance event to the relation event, which was the gap tracked by issue #15; that
+issue's scope (binding ⟶ plain DL in the AGM) is discharged by `Soundness.AGM.Adapter` and
+`.Probability`.
+
+What is left of the floor is therefore *not* "there is no adversary experiment". It is:
+
+* **Efficiency modeling.** The generic endpoints are parameterized by
+  `ComputedAlgebraicFSFamily.ReductionEfficient R`, and their `_poly` forms discharge it
+  unconditionally via `reductionEfficient_poly` at `(8·Q+1)·10^k`.
```
daira: What does this formula refer to? It's not clear.

book/src/formal-verification/glossary.md
```diff
@@ -59,8 +59,9 @@ development. Anchors point to a module + name under `Zcash/`.
...
+<div class="g"><div class="g-head"><span class="term">expected extraction time</span><span class="anchor">Forking.Adversary.ExpectedRunsPoly.recursiveAlgebraicFork_oracle_tape_sum_runs_le_poly · AGM.ProbabilityCoins.relationWithCoins_prob_le_of_truncated_textbookDL</span></div><div class="def">The unconditional AFK analysis bounds the recursive extractor's expected adversary-run count, averaged over oracle tables and extractor tapes, by <code>(8·Q+1)·10ᵏ</code>. This is field-independent and does not assume <em>fork spread</em>; <code>reductionEfficient_poly</code> wires it into the family runtime interface. The deployed combined finder adds at most three calls, truncates at the fixed budget <code>L</code>, and charges the over-budget tail explicitly as <code>((8·Q+1)·10ᵏ+3)/(L+1)</code>. The earlier <code>(6/δ)ᵏ</code> bound under <code>FamilyForkSpread</code> remains as an optional density-sensitive alternative.</div></div>
```
daira: Define "AFK", and expand the acronym or link the definition in other glossary entries.

book/src/formal-verification/proof-journey-data.json
```diff
@@ -391,13 +406,19 @@
...
         "On a real binding run - acceptance while the accepted value disagrees with the carried aggregate opening - every produced instance yields an explicit relation: the kernel's relation branch, or the commitment collision between its clean opening and the carried opening.",
-        "The full real binding probability is bounded by extraction loss, the z = 0 slice, and the fixed-slot DL term.",
-        "The fixed-slot relation-to-discrete-log transformation is deterministic and its miss refers to the same relation.",
-        "For distinct parameter queries, the Fiat-Shamir binding event in the uniform generator-RO model is bounded by the discrete-log advantage."
+        "The full real binding probability is bounded by extraction loss, the z = 0 slice, and the programmed-basis DL term.",
+        "The programmed-basis relation-to-discrete-log transformation is deterministic and its miss refers to the same relation; the miss costs exactly one 1/|F| hyperplane.",
+        "The recursive extractor has the unconditional field-independent expected-call bound (8·Q+1)·10^k, with no FamilyForkSpread premise.",
```
daira: $10^k$ is a lot, no?

daira (approved):
ACK modulo CI failures, with comments. It may be necessary to explain more about the $(8·Q+1)·10^k$ expected call bound.

clarus:
According to Codex (no idea if this is actually true):

--- 

One concrete issue remains for the advertised finite-security/work-factor profile:

- `AGM/StraightLineFiniteSecurity.lean:20-24` uses `4Q + 3(11+k)`, while `AGM/StraightLinePinnedRoots.lean:208-210` runs the adversary and then performs the initial direct finder's `11+k` designated oracle reads. `Q` bounds the adversary itself, not those post-run reads. On the worst four-invocation path, absent an explicit caching/charging convention, the conservative count is `4Q + 4(11+k)`. Correct the formula or state/prove the convention before relying on the five-bit finite-security profile. This does not affect the AFK/probability mathematics.

---

daira (commented):
Reviewed at head `847df819` (the post-#122 rebase), with close reads of the new probability core and targeted reads of the straight-line modules. Verdict: the proof content I checked is sound and the rebase is clean; the main ask is a description refresh, because the force-push delivered strictly more than the body claims.

## What I verified

- `relProb_le_tightSuccProb_add_inv` (`AGM/Probability.lean`, the piece the description flags for a close read): the argument is a hyperplane-slice count. For a fixed basis sample with a nontrivial returned coefficient vector, the planted-offset vectors annihilating the mask combination form a `1/|F|` slice, made rigorous by injecting `Miss × F` into the relation rectangle via pivot-slot update — injectivity holds exactly because the annihilation equation with a nonzero pivot coefficient recovers the discarded pivot coordinate. The additive form then follows from the coarse `|relSet| ≤ |ι → F|`. Sound, and the additive `1/|F|` in `relation_prob_le_of_tightDL` replaces the old multiplicative `2^k + 2` slot guess as advertised.
- The efficiency floor is now unconditional: `reductionEfficient_poly` discharges `ReductionEfficient` from the adversary's query bound alone via `afkRunBound Q k = (8Q + 1) · 10^k` (field-independent), with `reductionEfficient_of_forkSpread` demoted to an optional density-sensitive alternative and `reductionEfficient_exponential` kept as the record of the old analysis. `knowledgeSoundness_under_DL_computed` bounds the right event — the adversary wins the full FS game and `cleanOpening` returns no witness data — and the `_poly` corollary needs no efficiency premise at all.
- The concrete security accounting does not pay the AFK factor. `afkRunBound Q 11 = (8Q + 1) · 10¹¹` is a ~2³⁹·⁵ multiplier, which would eat ~40 bits of any time-success comparison — but it prices only the retained recursive alternative. `StraightLineFiniteSecurity` composes the primary route from at most four prover invocations (`straightLineDlogGroupWork = 4 · proverGroupWork + reductionGroupWork`, with `4Q + 3(11 + k)` oracle queries), records group work, RO queries, and direct-decode field work as distinct quantities, and deliberately asserts no generic-group DLOG formula — the finite-security Vesta advantage is a caller input. The ~123-bit headline can only stand on this route.
- The rebase is genuinely onto current main (merge-base = `2b2ad3e7`), the census delta is fully qualified and `+native(owner)`-annotated per the new macros (their exactness is machine-checked, so a green Lean CI settles it), and every census entry the PR removes corresponds to a deleted declaration — no coverage loss.
- The `hpinned` surviving in `ScheduleBudget.lean` is main's generic schedule constructor argument, with the trace-deriving stronger constructor beside it — not the bare capstone premise #96's description flagged; the straight-line code no longer references any removed machinery (build-enforced by the rebase).
- `ZeroData.lean` is the zero-data multiopen assembly toolkit backing the witness/satisfiability route; its module doc matches its content.
- The book-CI failure shown on the PR is from the pre-push head: all seven source files the proof-journey validation reported missing exist at `847df819`.

## Findings

1. **The description predates the force-push and now undersells and mis-describes the PR.** It presents `FamilyForkSpread` as the mechanism discharging `ReductionEfficient` (now the optional route; `reductionEfficient_poly` is unconditional), and its review note says "Deployed endpoints still route through the fixed-slot AGM bound; migrating them is a separate change" — but `Composition/StraightLineDeployed.lean` and `Fixtures/MultiAction/StraightLineKnowledgeError.lean` describe themselves as the *primary* deployed route and endpoint, with the recursive AFK capstone retained as the separately priced alternative. Please refresh the body — and in doing so, quantify "separately priced": the recursive alternative's call bound is `(8Q + 1) · 10^k` ≈ 2³⁹·⁵ · Q at the captured `k = 11`, versus the primary route's four prover invocations, so the two routes differ by ~40 bits in any time-success reading and the headline number rests on the straight-line route alone.
2. **"Closes #115" needs a per-task accounting.** #96 transferred three concrete tasks to #115 (captured-shape prover family carrying the fresh-query computations; a field-independent polynomial *total-cost* bound for the direct-coordinate postprocessing, not merely an invocation count; semantic promotion or an explicit compressed-identity scope per endpoint). `straightLineDlogGroupWork` suggests the total-cost side is now genuinely addressed, but the refreshed description should say explicitly which of the three land here and which remain — "Closes" only if all three do.

TalDerei:
follow-ups for integrating #99 (references https://github.com/zcash/ironwood/issues/115#issuecomment-5096888287) will be part of this PR.

book/src/formal-verification/proof-journey-data.json
```diff
@@ -380,24 +396,66 @@
...
-        "The full real binding probability is bounded by extraction loss, the z = 0 slice, and the fixed-slot DL term.",
-        "The fixed-slot relation-to-discrete-log transformation is deterministic and its miss refers to the same relation.",
-        "For distinct parameter queries, the Fiat-Shamir binding event in the uniform generator-RO model is bounded by the discrete-log advantage."
+        "The full real binding probability is bounded by extraction loss, the z = 0 slice, and the programmed-basis DL term.",
+        "The programmed-basis relation-to-discrete-log transformation is deterministic and its miss refers to the same relation; the miss costs exactly one 1/|F| hyperplane.",
+        "The recursive extractor has the unconditional field-independent Attema-Fehr-Klooß-style expected-call bound (8·Q+1)·10^k, with no FamilyForkSpread premise. 10^k is large: at the deployed k = 11 and Q = 2^122 the expectation is about 2^161.5, far beyond Vesta's ~2^126 discrete-log cost, so this is a structural polynomial bound rather than a concrete security margin - the concrete margin is the straight-line endpoint's.",
```
daira: Link the paper and give a date — e.g. `expected-call bound (8·Q+1)·10<sup>k</sup> in the style of [Attema–Fehr–Klooß 2021](https://eprint.iacr.org/2021/1377)`.

TalDerei:
@daira currently running auditing scans, please hold off reviews shortly.

daira:
> So we parameterize the concrete action capstone over some n < consensusmax, so BundleStatement quantifies over Fin n.

There's an off-by-one in the PR description here; you have two <'s (one implicit in `Fin`).

TalDerei:
> The recursive extractor survives only as a separately priced alternative, at AFK-style `(8Q+1)·10^k` — roughly a `2^39.5` factor on `Q` at `k = 11`

now stale by https://github.com/zcash/ironwood/pull/142.

TalDerei:
> There's an off-by-one in the PR description here; you have two <'s (one implicit in `Fin`).

human descriptions are error prone :) code obviously operates over 2^16 - 1.

Zcash.lean
mitschabaude, line 1: can you confirm that we want the trustboundary file to depend on the large fixtures, and consequently not live in the main target?
previously we had fixture-dependent trust assertions, like the old Action capstones, living in `Fixtures/{SingleAction,MultiAction}/TrustBoundary.lean`, so there's an inconsistency now.

mitschabaude:
comments for another PR:
* IMO we haven't really figured out where our main theorems are supposed to live. this for example places `StraightLineActionTerminal.lean` in `Circuit/Integration` and `ActionCapstone` in `Fixtures/`. I would have expected both in `Soundness`. Any import from Soundness into `Circuit/Integration` feels like a smell of inverted dependencies.
* `orchard_action_acceptFalseStatement_prob_le_captured` and its helpers continues our tradition of not following Lean theorem naming convention: there's nothing in the theorem statement called "orchard", nothing literally called "action", no property called "AcceptFalseStatement", and nothing called "captured"
* `orchard_action_acceptFalseStatement_prob_le_captured` also continues our tradition (that I contributed a lot to) of stating our main theorems in barely readable form because they collect dozens of hypotheses instead of grouping these into clear bundled concepts

mitschabaude (approved):

Zcash.lean
TalDerei: the original motivation there was premised on those fixture-specific trust boundary files being removed and everything consolidated and namespaced under one top-level file, but that never happened. thinking about this more, i don't like the approach this PR takes; will:
- move the new action capstone entries into the `Fixtures/MultiAction/TrustBoundary.lean` fixture, 
- restore import `Zcash.TrustBoundary in `Zcash.lean`, 
- and delete the `TrustCheck` target as unnecessary

TalDerei:
> comments for another PR:

this is really great feedback, and i've been kicking the can down the road here. will consolidate the organizational cleanups into a dedicated PR so they don't get buried here.

Zcash/Circuits/Integration/AdaptiveActionEvent.lean
```diff
@@ -0,0 +1,1139 @@
+import Zcash.Circuits.Integration.AdaptiveActionSurfaces
+import Zcash.Snark.Soundness.AGM.AdaptiveComposition
+
+/-!
+# Literal Action soundness for arbitrary adaptive online-AGM adversaries
+
+This file exposes the executable adaptive Action terminal as one combined relation finder.  Its
+public event is the literal deployed-verifier event `accept ∧ ¬ BundleStatement`; no
```
daira: That's correct for a violation of soundness, but don't we need knowledge soundness?

TalDerei: good observation! yeah this is a knowledge soundness gap because `BundleStatement` is a prop (introduced in #99), but i wouldn't characterize it strictly a "breaks as computed data" violation.

Zcash.lean
TalDerei: https://github.com/zcash/ironwood/pull/114/changes/26aee138938b34d1f5caed01ec18b7701a6d7494

Zcash/Circuits/Integration/AdaptiveActionEvent.lean
TalDerei: the only other one i could find was the ledger witness, which is also prop-only: https://github.com/zcash/ironwood/blob/26aee138938b34d1f5caed01ec18b7701a6d7494/Zcash/Security/Ledger/Bridge.lean#L779-L790 (but we're already tracking in #22).

TalDerei: fixed and updated proof map: https://github.com/zcash/ironwood/pull/114/changes/e189de4519645107f84579aa4e8aa0735836540f

book/src/formal-verification/proof-journey-data.json
```diff
@@ -440,22 +465,15 @@
...
+        "The Action knowledge extractors are executable and return the actual private witnesses for every proof in the bundle together with proofs that they satisfy the compiled Action specification; both the adaptive and straight-line/sequential relation projections share their route's data-bearing outcome.",
+        "The direct profiles separately account for random-oracle, group, and direct-coordinate work while leaving the finite-security Vesta DLOG advantage as a caller premise.",
+        "For every numProofs at or below Orchard's consensus maximum, the generic adaptive Action knowledge endpoint bounds verifier acceptance when executable private-witness extraction fails by Adv_DLOG(2^127,2^127) + 2^-83, prices all five semantic challenge surfaces, profiles the shared extractor, and exposes an explicit PMFEventBiasLE transport conjunct; it requires no externally supplied traces, cuts, or views."
```
daira, line 473: Adv_DLOG is still not defined.

book/src/formal-verification/proof-map-embed.html
```diff
@@ -362,26 +363,26 @@ <h2>Detail</h2>
...
   { id:'hlrel', c:4,r:8, w:225, label:'high-level relation', status:'floor',
-    anchor:'hencodes',
-    role:'Semantic adequacy (hencodes): SnarkRelation — the witness satisfies the gates — implies the intended high-level Ironwood statement (note well-formed, values balance, nullifier derived, spend authorized). An assumed hypothesis over a free Prop; the composition otherwise stops at SnarkRelation. The output-side gap. Not started.' },
+    anchor:'ActionBundleWitness → Security.Ledger.Statement',
+    role:'The output-side floor, moved one step out. The composed knowledge capstone returns an ActionBundleWitness containing each actual circuit PrivateWitness and a proof that it satisfies the deployed Action circuit’s compiled specification. What stays outside Lean is identifying those circuit witnesses and that specification with the abstract Orchard Action witness relation (protocol spec §4.17.4) consumed by the ledger security games: the target is transcribed in Security/Ledger/Statement.lean, but nothing yet connects it to actionCircuit.' },
```
daira, line 385: That's a big gap, no? I don't think it is obvious to consumers of this repo that it isn't done yet.

TalDerei: tracked in https://github.com/zcash/ironwood/issues/147.

TalDerei: gap is end-to-end ledger knowledge soundness isn't demonstrated yet (only action circuit knowledge soundness and ordinary ledger soundness, and we need to bridge them)

TalDerei:
@daira can i fold these proof map changes into a broader set of refinements? don't want to kick off CI again.

Zcash/Circuits/Integration/AdaptiveActionEvent.lean
```diff
@@ -0,0 +1,1650 @@
...
+    (hroots : family.AdaptiveDeployedGoodRoots basis O witness)
+    (hshifted : family.AdaptiveShiftedValue basis O)
+    (haccepts : adaptiveActionAccepts family basis O) :
+    Option (AlgebraicRelationWitness (F := Fp) basis) :=
+  match adaptiveActionPreXIdentityWitnessOrRelationFinder pp family inputs hvk hI hchar
+      basis O hprovenance witness hsrc hroots hshifted haccepts with
+  | some (Sum.inr relation) => some relation
+  | _ => none
+
+theorem adaptiveActionPreXIdentityRelationFinder_isSome_of_witnessOrRelation
```
daira, line 248: I would really like all theorems to have a doc comment summarizing what they say.

TalDerei: fair, will track.

PR #120, opened 2026-07-27
Enforce the csimp-census and native-opt-in trust rules

daira:
Ports the two trust-enforcement rules adopted in CompElliptic (daira/CompElliptic#16, daira/CompElliptic#17) to this repository:

- **csimp census**: every `@[csimp]` replacement lemma must carry its own `assert_axioms` entry in `Zcash/TrustBoundary.lean`, because the axioms of a csimp proof are not propagated into downstream `native_decide` axiom tracking ([lean4#7463](https://github.com/leanprover/lean4/issues/7463)). The one existing csimp lemma (`Zcash.Arithmetic.Msm.evalNat_eq_evalNatFast`) is now censused at the standard tier; `scripts/check_csimp_census.sh` enforces the rule in CI.
- **native opt-in**: executing a check through a locally compiled `precompileModules` dylib (ours or the CompElliptic pin's) is a coarse-grained compiler-trust event with no axiom trace, and loading is inseparable from elaborating lane-importing modules — so the enforced invariant sits at the level of checks: no module whose import closure reaches a lane module may contain an evaluation-based check (`#eval`, `#guard`, `native_decide`) without an explicit opt-in entry. `scripts/check_native_optin.py` enforces this in CI; the local lane is parsed from `lakefile.toml`, the pin's lane is an explicit mirror re-checked on pin bumps. Currently one module in the lane import cone and no ambient checks. If #99's VK-pin checks rely on native compilation, they will need to be added to the opt-in list, with rationale.

The book's formal-verification page documents both rules; the underlying mechanism study is [`design/lean-native-trust-research.md`](https://github.com/daira/CompElliptic/blob/main/design/lean-native-trust-research.md) in the CompElliptic repository. Both guards share the non-adversarial scope of the existing fixture checks. Full `lake build Zcash` green at the census commit; the guard commits are script/CI/book only.

🤖 Claude Fable 5

### DIFF
--- a/.github/workflows/lean.yml
+++ b/.github/workflows/lean.yml
@@ -75,6 +75,12 @@ jobs:
       # native_decide axiom tracking (lean4#7463).
       - name: Verify csimp lemmas are axiom-censused
         run: scripts/check_csimp_census.sh
+      # Executing a check through a locally compiled precompileModules dylib is a
+      # coarse-grained compiler-trust event with no axiom trace, so it must be an
+      # explicit opt-in: no module reaching a lane module by imports may contain an
+      # evaluation-based check without an entry in the script's opt-in list.
+      - name: Verify native-executing checks are opt-in
+        run: python3 scripts/check_native_optin.py
           diff Stamp.lean <(./stamp.sh) \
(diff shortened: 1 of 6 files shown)


TalDerei (approved):
ACK this was a good catch

PR #122, opened 2026-07-27
fv: require fully-qualified names and native provenance in the trust census

daira:
Motivated by finding 2 of [the #99 re-audit](https://github.com/zcash/ironwood/pull/99#issuecomment-5095766985): an unqualified `assert_axioms` entry resolved silently to a generic homonym while the intended deployed theorem was not in scope at all, so the census read as covering the PR's headline result while checking a different theorem.

- The first commit qualifies every `assert_axioms` / `assert_computable` entry (569 across `Zcash/TrustBoundary.lean`, both fixture censuses, and `BridgeTests.lean`) and the eight `#guard_msgs`-pinned `#print axioms` targets, and drops the opens the census files no longer need. The guard pins now record the `native_decide` axioms' full names. An audit of every rewritten entry against its census section found no wrong-cousin resolutions on main — the finding-2 hazard was introduced on #99's branch, not latent here.

- The second commit makes `unc_thirteen_not_isSquare` public: it owns a `native_decide` axiom that appears in downstream axiom footprints, where a private owner can only be written in its mangled form.

- The third commit makes the macros enforce both conventions. `assert_axioms` / `assert_computable` error unless the written name equals the resolved constant's full name (a `_root_.` prefix is accepted), so the silent-capture case becomes a loud error: writing the intended full name of a declaration whose module is not imported is an unknown identifier, not a quiet resolution to a same-named cousin. And `+native` now requires naming the declarations that own the permitted `native_decide` axioms — `+native(D₁, ...)`, fully qualified — since a bare `+native` would accept a native axiom smuggled in by any declaration entering the dependency cone. The annotation is checked exactly: a new native axiom, or a stale entry, fails the build with the list to write. The 204 `+native` entries now each name their native trust surface (for example, `orchardActionCircuit` names its six fixed-base window-table certificates plus CompElliptic's two Pallas curve facts); five entries whose `+native` reached no native axiom at all lose the flag.

In-flight PRs will hit the new checks on rebase for any census entries they add; that is the point. The same change should be mirrored into CompElliptic's copy of `AxiomCheck.lean` (the file header notes the two are kept in sync); that is queued with the native-code split.

Full build green.

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Meta/AxiomCheck.lean
+++ b/Zcash/Meta/AxiomCheck.lean
@@ -12,6 +12,11 @@ declared tier (a `sorry`, an unexpected axiom, or `native_decide` where none was
 
 The `#guard_msgs`-pinned form remains the right tool when the *exact* axiom set is the claim.
 
+Both commands require their argument to be written fully qualified (`checkFullyQualified`):
+an unqualified name resolves through the census file's `open`s, so a same-base-name cousin in
+an opened namespace can silently capture an entry meant for a declaration that is not in scope
+at all — the assertion then reads as covering one theorem while checking another.
+
 This command is also used in CompElliptic (`CompElliptic/Meta/AxiomCheck.lean`); changes here
 should be reflected there and vice versa.
 -/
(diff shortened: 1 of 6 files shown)


Zcash/Meta/AxiomCheck.lean
```diff
@@ -29,19 +34,75 @@ the check stable across the toolchain-dependent axiom naming. -/
 def isNativeDecideAxiom (n : Name) : Bool :=
   n.components.any (· == `native_decide)
 
+/-- Assertion names must be written fully qualified. Resolution through `open`s is
+context-dependent: a same-base-name cousin in an opened namespace can silently capture an
+entry meant for a declaration that is not even in scope, so the assertion reads as covering
+one theorem while checking another. Requiring the written name to equal the resolved
```
daira, line 40: Nit: this is duplicative of the file-level comment.

daira (commented):
Self-ACK

TalDerei (approved):
ACK much better!

Issue #121, opened 2026-07-27
RedDSA ±-randomized unforgeability


daira:
Split out of #22, which now tracks only extractability — the two obligations are different theorems with different consumers.

Spend Authority consumes unforgeability of the ±-randomized signature scheme: `SpendAuthForgery` (`Zcash/Security/Ledger/SpendAuthority.lean`) packages a verifying signature under a randomization of `akV` or `-akV` (recorded by `negated`), over a message outside the victim's signing history, and the Spend Authority capstone bounds its forgery arm by a named ε hypothesis that this issue discharges.

The ± is deliberate: `ak` is consumed as a single extracted coordinate, so the statement pins ak^ℙ only up to sign, and the unforgeability hypothesis must cover both `akV` and `-akV` — a factor-of-2 loss, explicitly accepted in the design (see the note in spec §4.18.4).

The required notion is the spec's **SURK-CMA** — Strong Unforgeability with Re-randomized Keys under adaptive Chosen Message Attack ([Signature with Re-Randomizable Keys](https://zips.z.cash/protocol/protocol.pdf#abstractsigrerand)) — which the spec requires of every `SpendAuthSig` instantiation. Strong means freshness is over (message, signature) pairs: a new signature on a previously-signed message is a win. The formalized arm's `¬ Signed msg` freshness is the weaker message-level consequence, discharged a fortiori. The SURK-CMA game as specified is the spec author's own invention (including the single-oracle simplification relative to the literature's two-oracle form), so this issue includes a thorough check of the game definition itself, not only the reduction under it.

Discharging it means formalizing RedDSA (scheme definition, verification equation, randomization — all completed in #157), stating the SURK-CMA experiment with the ± extension, checking the game definition, and connecting its ε to the capstone's named hypothesis.

🤖 Claude Fable 5

PR #124, opened 2026-07-28
fv: deployed capstone discharges

daira:
The deployed instantiation's break-to-relation extraction for the ledger-game capstones. Every Balance-subset break arm now reduces deterministically to a discrete-log relation among the fixed Orchard Sinsemilla bases: a violating valid ledger computes the relation as data. This is a deterministic per-ledger extraction, not yet the probabilistic ε discharge — see the scope note below.

The headline is the key-binding arm. In ZIP 2005 its security rests on a random-oracle collision argument — a birthday bound, with `Commit^ivk` modelled as a random oracle, which is a heuristic rather than a falsifiable assumption. The deployed construction needs none of that: two `Commit^ivk` openings of the same `ivk` exhibit an honest discrete-log relation among the fixed Sinsemilla bases, computed as data, with no oracle, no birthday bound, and no oracle-machine → `PMF` bridge. The reduction is deterministic.

With this, all three Balance-subset arms extract to the same target — a nontrivial relation among the deployed Sinsemilla bases, whose intractability is the discrete-log-relation hardness the SNARK soundness stack and the binding-signature balance argument already reduce to. At the deterministic level this removes the birthday/oracle loss the ZIP 2005 key-binding argument incurs. It does **not** yet collapse the capstone's named ε hypotheses: `balanceSubset_measure_le` still takes separate per-arm ε bounds, and instantiating them under DLR hardness needs the efficient-adversary composition (pushing a `PMF` over valid annotated ledgers through the reducer, a runtime/query bound, and applying hardness). That efficiency bridge — the DLR-side analogue of the Recovery oracle-machine → `PMF` bridge (#107) — is not in this PR; the RedDSA obligations stay honestly named (#22, #121).

The second result is a single two-sided Balance theorem. `balanceOrBreak` states: in a valid ledger, either the value after the first `i + 1` transactions is accounted for exactly — both the shielded and transparent pools non-negative and summing to the minted issuance — or the ledger's own data computes a break. This ties the two Balance arguments into one statement, so a reader no longer assembles them by hand: the upper side is "no value created within a transaction" (the binding signature), the lower side is "no value spent that was not created" (spend-integrity, a new lower bound from Balance-subset). Details in the `Balance.lean` bullet below.

- `commitGen`, its linearity lemmas (plus a new `commitGen_neg`), and `NontrivialRelation.ofCombinationCollision` move to `Zcash/Common/DiscreteLogRelation.lean`: they are generic over any `(F, G)`-module and belong with the relation types they serve — the ledger's Sinsemilla reductions consume them too, and importing the SNARK soundness stack into the ledger modules would tangle the census cones. Call sites resolve outward unchanged.

- The chain-collision reducer (`SinsemillaDLR.lean`): two defined Sinsemilla chains at the same domain point, blinded by `r₁ • W` and `r₂ • W`, whose outputs agree up to sign, compute a nontrivial relation among the table, the domain point, and `W` — the shared core of the key-binding, note-commitment, and Merkle arm discharges. The negation case is unconditional (the domain-point coefficients `2^n` and `−2^n` cannot agree in the odd-order scalar field); the equality case rests on the proven `preCoeffs_inj`.

- **The binary-expansion core of spec Theorem 5.4.3 is proven** (`preCoeffs_inj`): chunk lists of equal length at most 253 are determined by their scalar-field coefficients. The ℕ-level coefficient sums cast faithfully because deployed message lengths keep them below the field order, and they determine the list by cons induction — the head's `2^n` term dominates the tail's `≤ 2^n − 1`, forcing head equality, with the tails following by cancellation; no binary-digit machinery is needed. The chunk-encoding side is `chunksOf_inj` / `commitIvkChunks_inj` (`Circuits/Specs/Sinsemilla.lean`): the `K`-bit digits determine the value by a `Nat.mod_mul` induction, and the 51-chunk `(ak, nk)` message splits by mod and cancellation.

- The deployed key-binding discharge (`KeyBindingDLR.lean`) is **hypothesis-free**: a `KeyBinding.Pool.Break` — two valid `Commit^ivk` openings of the same `ivk` disagreeing on their opening projection — computes the nontrivial relation among the Sinsemilla table, the `CommitIvk` domain point, and the randomness base, with the defined chains recovered computably (`Option.get` on the break's own hash hits). What is complete is the deterministic break-to-relation reduction; wiring it into the capstones' named-ε slot, and the onward relation-to-single-DL step, land with the composition below.

- The note-commitment arm (`NoteCommitDLR.lean`) and the Merkle arm (`MerkleDLR.lean`) instantiate the same reducer, both hypothesis-free. The note arm recovers each compressed point from its `x`-coordinate and `y`-parity bit (`eq_of_toPoint_x_eq_of_y_parity_eq`) and its message by `noteCommitChunks_inj`; the Merkle arm applies the reducer with zero blinding and converts the result to the one-generator form (`NontrivialRelation.toOne`), since a zero randomness-base coefficient carries no nontriviality.

- `DeployedCapstone.lean` composes them. `deployedBalanceSubsetOrRelation` is the deployed Balance-subset reduction: a valid deployed ledger either covers its nonzero spends, or its own data computes one of the three discrete-log relations. At the deterministic level the key-binding arm no longer routes through the random-oracle model: the extraction is ROM-free. Wiring this into the capstone's named-ε slots (instantiating `balanceSubset_measure_le`) is the efficient-adversary composition noted above, and is not done here. The relation-to-single-DL step remains the named pre-quantum hardness (independence of the deployed bases, spec Theorems 5.4.3 and 5.4.4).

The three arm modules and the composition are in the library root and censused, with the native provenance the deployed constants inherit (the Pallas curve-order witness and the fixed-base certificates) recorded per entry.

- `balanceOrBreak` (`Balance.lean`) is the two-sided Balance statement, composing the two arguments into one. In a valid ledger, either the value after the first `i + 1` transactions is accounted for exactly — both pools non-negative and summing to the minted issuance — or the ledger's own data computes a break. The shielded pool is non-negative by spend-integrity (`poolValueBalance_nonneg`, new, from Balance-subset: a sub-multiset of positioned openings sums to no more value); the transparent pool by validity (`transparent_nonneg`); the two sum to issuance by per-transaction conservation (`valueConservationOrBreak`, from the binding signature). The value identity is an equality, not a bound — value is not destroyed in the model. This makes the roles of the two Balance halves explicit rather than leaving a reader to assemble them: the upper side is "no value created within a transaction" (binding signature), the lower side is "no value spent that was not created" (spend-integrity).

Still to land: the probabilistic ε discharge — the efficient-adversary composition that instantiates `balanceSubset_measure_le` from the deterministic reducer under DLR hardness (`PMF` pushforward + runtime/query bound + hardness application; the DLR-side analogue of #107); the deployed instantiation of the full two-sided Balance (routing the value premiss to the binding-signature relation); and the NU6.3 flags in `ActionInstance`.

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Circuits/Specs/Sinsemilla.lean
+++ b/Zcash/Circuits/Specs/Sinsemilla.lean
@@ -279,6 +279,30 @@ theorem chunksOf_inj {a b n : ℕ} (ha : a < 2 ^ (K * n)) (hb : b < 2 ^ (K * n))
   have hfin := hmod n le_rfl
   rwa [Nat.mod_eq_of_lt ha, Nat.mod_eq_of_lt hb] at hfin
 
+set_option exponentiation.threshold 600 in
+/-- `merkleChunks` is the 52-chunk decomposition of its packed message. -/
+theorem merkleChunks_eq_chunksOf (l lv rv : ℕ) :
+    merkleChunks l lv rv = chunksOf (l + 2 ^ 10 * lv + 2 ^ 265 * rv) 52 := by
+  simp [merkleChunks, chunksOf, bitrange]
+
+set_option exponentiation.threshold 600 in
+/-- The 52-chunk Merkle compression encoding is injective on components within
+their widths. -/
+theorem merkleChunks_inj {l lv rv l' lv' rv' : ℕ}
+    (h1 : l < 2 ^ 10) (h2 : lv < 2 ^ 255) (h3 : rv < 2 ^ 255)
+    (h1' : l' < 2 ^ 10) (h2' : lv' < 2 ^ 255) (h3' : rv' < 2 ^ 255)
+    (h : merkleChunks l lv rv = merkleChunks l' lv' rv') :
+    l = l' ∧ lv = lv' ∧ rv = rv' := by
+  rw [merkleChunks_eq_chunksOf, merkleChunks_eq_chunksOf] at h
+  have hK : (2 : ℕ) ^ (K * 52) = 2 ^ 520 := by norm_num [K]
+  have hbound : ∀ {a b c : ℕ}, a < 2 ^ 10 → b < 2 ^ 255 → c < 2 ^ 255 →
+      a + 2 ^ 10 * b + 2 ^ 265 * c < 2 ^ (K * 52) := by
+    intro a b c ha hb hc
+    rw [hK]
+    omega
+  have hmsg := chunksOf_inj (hbound h1 h2 h3) (hbound h1' h2' h3') h
+  omega
+
 /-- Chunks below position `n` are unaffected by reducing `val` mod `2 ^ (K * n)`. -/
 theorem chunksOf_mod (val n : ℕ) : chunksOf (val % 2 ^ (K * n)) n = chunksOf val n := by
   unfold chunksOf
@@ -464,9 +488,9 @@ def noteCommitChunks (gdX gdY pkdX pkdY v rho psi : ℕ) : List ℕ :=
 theorem noteCommitMessage_inj
     {gdX gdY pkdX pkdY v rho psi gdX' gdY' pkdX' pkdY' v' rho' psi' : ℕ}
     (h1 : gdX < 2 ^ 255) (h2 : gdY < 2) (h3 : pkdX < 2 ^ 255) (h4 : pkdY < 2)
-    (h5 : v < 2 ^ 64) (h6 : rho < 2 ^ 255) (h7 : psi < 2 ^ 255)
+    (h5 : v < 2 ^ 64) (h6 : rho < 2 ^ 255) (_h7 : psi < 2 ^ 255)
     (h1' : gdX' < 2 ^ 255) (h2' : gdY' < 2) (h3' : pkdX' < 2 ^ 255) (h4' : pkdY' < 2)
-    (h5' : v' < 2 ^ 64) (h6' : rho' < 2 ^ 255) (h7' : psi' < 2 ^ 255)
+    (h5' : v' < 2 ^ 64) (h6' : rho' < 2 ^ 255) (_h7' : psi' < 2 ^ 255)
     (h : noteCommitMessage gdX gdY pkdX pkdY v rho psi
         = noteCommitMessage gdX' gdY' pkdX' pkdY' v' rho' psi') :
     gdX = gdX' ∧ gdY = gdY' ∧ pkdX = pkdX' ∧ pkdY = pkdY' ∧ v = v' ∧ rho = rho'
--- a/Zcash/Circuits/Specs/Sinsemilla.lean
+++ b/Zcash/Circuits/Specs/Sinsemilla.lean
@@ -258,6 +258,27 @@ theorem chunksOf_add (val m n : ℕ) :
   congr 2
   ring
 
+/-- `chunksOf` is injective below `2 ^ (K * n)`: the `K`-bit digits determine the value. -/
+theorem chunksOf_inj {a b n : ℕ} (ha : a < 2 ^ (K * n)) (hb : b < 2 ^ (K * n))
+    (h : chunksOf a n = chunksOf b n) : a = b := by
+  have hdig : ∀ i, i < n → bitrange a (K * i) K = bitrange b (K * i) K := by
+    intro i hi
+    have hgd := congrArg (fun l => l.getD i 0) h
+    simpa [chunksOf, List.getD_eq_getElem?_getD, List.getElem?_map, List.getElem?_range, hi]
+      using hgd
+  have hmod : ∀ m, m ≤ n → a % 2 ^ (K * m) = b % 2 ^ (K * m) := by
+    intro m
+    induction m with
+    | zero => simp [Nat.mod_one]
+    | succ i ih =>
+      intro hi
+      have hstep := hdig i (by omega)
+      simp only [bitrange] at hstep
+      rw [show K * (i + 1) = K * i + K from by ring, pow_add, Nat.mod_mul, Nat.mod_mul,
+        ih (by omega), hstep]
+  have hfin := hmod n le_rfl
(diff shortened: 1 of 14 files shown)


TalDerei:
At current head `b31baf4d`, I found one actionable implementation defect and one tracked integration boundary.

**Actionable — the move of `ofCombinationCollision` to `Common` is incomplete.** `Zcash/Common/DiscreteLogRelation.lean` adds `Zcash.NontrivialRelation.ofCombinationCollision`, but `Zcash/Snark/Soundness/Deployed/Binding.lean:27-45` still defines the old `Zcash.Snark.NontrivialRelation.ofCombinationCollision`; this PR only removes that definition's doc comment. The existing unqualified call sites under `namespace Zcash.Snark` therefore continue to resolve to the namespace-local copy, while `Zcash/TrustBoundary.lean:349` now checks the new root declaration. Please delete the old definition so those callers resolve outward, or explicitly update the callers and keep the declaration they actually use in the computability census.

**Tracked incompleteness, not a hidden algebraic hole.** `Zcash/Security/Ledger/KeyBindingDLR.lean` is not imported by `Zcash/Security.lean` or another build root/caller, so the passing default CI did not compile the main new concrete reduction (an explicit targeted `--wfail` build does pass). `relationOfKeyBindingBreak` also still requires caller-supplied `hcoeffs` and `hchunks` and has no game/capstone consumer. Thus the current head computes a fixed-basis N-generator DLR witness from a concrete break, but it does not yet discharge the ledger PMF event's `ε_kb` or reduce that relation to textbook one-DL. The PR body already acknowledges the injectivity proofs, root import, census, and capstone composition as still to land; those should land before saying that `ε_kb` has been folded into the final DL terminal.

I did not find a specification or algebraic correctness bug in the reducer itself: the fixed 510-bit message shape, `±` extraction split, coefficient equations, and deployed CommitIvk Q/R constants check out.

🤖 Claude Fable 5

daira:
Right — the move deleted only the doc comment. Fixed and folded into the move commit: the namespace-local copy is gone, so the call sites resolve to the Common declaration that the census pins. Your other point predates the current head: `preCoeffs_inj` and `commitIvkChunks_inj` are now proven and `relationOfKeyBindingBreak` takes no hypotheses. Description reworded — what is complete is the break-to-relation reduction; the capstone ε_kb wiring and the relation-to-single-DL step land with the composition.

TalDerei (approved):
optimistic ACK since it's being tracked.

TalDerei:
At live head `520c7a3`, I found one blocking soundness/claim gap, but no concrete algebraic or encoding bug.

### Soundness — blocking relative to the PR’s claims

`deployedBalanceSubsetOrRelation` is only a deterministic, per-ledger extraction theorem: a violating valid annotated ledger yields one of three fixed-basis DLR witnesses. It does **not** discharge the probabilistic capstone’s ε hypotheses or prove security under one computational assumption. [DeployedCapstone.lean](https://github.com/zcash/ironwood/blob/520c7a3e4c824a2f77adc3c95312637ffe332ff5/Zcash/Security/Ledger/DeployedCapstone.lean#L41-L64)

Concretely:

- `balanceSubset_measure_le` still requires separate `hm`, `hnc`, and `hkb` probability bounds. [Capstone.lean](https://github.com/zcash/ironwood/blob/520c7a3e4c824a2f77adc3c95312637ffe332ff5/Zcash/Security/Ledger/Capstone.lean#L193-L204)
- No efficient relation-finder machine, pushforward/event equivalence, runtime bound, or theorem applying DLR hardness appears.
- An arbitrary `PMF ValidAnnotated` cannot directly be bounded by computational DLR hardness: it may be supported on hard-to-compute relations. Deterministic extraction removes the birthday/oracle loss, but not the efficiency bridge.
- The existing documentation still explicitly says the named-ε connections are open. [Capstone.lean](https://github.com/zcash/ironwood/blob/520c7a3e4c824a2f77adc3c95312637ffe332ff5/Zcash/Security/Ledger/Capstone.lean#L20-L30)

So either narrow the PR description to “deterministic break-to-relation extraction,” or add the efficient-adversary composition and instantiate `balanceSubset_measure_le`. The relation-to-single-DL step is already honestly named as an assumption, so I do not count that explicit deferral as a hidden hole.

### Specification

No packing, domain-point, fixed-base, bit-order, or length discrepancy found. The CommitIvk, note-commitment, and Merkle encodings agree with the protocol’s Sinsemilla construction and injectivity argument. [Zcash protocol](https://github.com/zcash/zips/blob/main/protocol/protocol.tex#L9749-L9855)

The only specification-facing issue is the PR’s overstatement that the ε capstone is already collapsed to one computational assumption.

### Implementation correctness

No correctness defect found in:

- the `±` Sinsemilla collision reducer;
- coefficient/message injectivity;
- key-binding, note-commitment, or Merkle instantiations;
- `balanceOrBreak` and its indexing;
- the moved generic DLR helpers.

The four targeted modules build successfully at the exact head with `--wfail`.

Integration remains blocked: GitHub reports the PR `DIRTY` due to a `Zcash/TrustBoundary.lean` import conflict with current `main`; there are no checks, and the approval applies to older commit `86bee80`.

Verdict: **request changes** for the soundness/claim gap and rebase conflict. The deterministic extraction work itself looks correct.

With the help of Claude Fable 5.

daira:
Agreed — the description overstated it, now narrowed. `deployedBalanceSubsetOrRelation` is deterministic per-ledger extraction, and you are right that removing the birthday/oracle loss is not the efficiency bridge: an arbitrary `PMF ValidAnnotated` can be supported on hard-to-compute relations, so instantiating `balanceSubset_measure_le` still needs the pushforward, a runtime/query bound on the composite finder, and a DLR-hardness application. That is the DLR-side analogue of the Recovery oracle-machine → `PMF` bridge (#107); it is tracked as a remaining task, not claimed here. The rebase conflict is also resolved — the branch is now on current main with the `TrustBoundary.lean` import merged and a green full build.

TalDerei:
reminder https://github.com/zcash/ironwood/pull/129#pullrequestreview-4796221612.

PR #125, opened 2026-07-28
Remove the per-file copyright notices

daira:
Nine files carried copyright headers reading "Ironwood Contributors" under the Apache License 2.0 alone. This is [wrong in three ways](https://github.com/zcash/ironwood/pull/99#discussion_r3662094041):

* the attribution is "Copyright 2026 Zcash Protocol Developers";
* the licence is dual MIT / Apache 2.0;
* per-file notices are unnecessary in the first place — the README's Contribution section already places contributions under the dual licence with that copyright holder, so the notice belongs there alone.

This removes all nine headers. The only other copyright mentions in the repo are the licence texts themselves, the README's canonical notice, the vendored Mermaid bundled header, and the fonts' OFL — all deliberately untouched.

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Arithmetic.lean
+++ b/Zcash/Arithmetic.lean
@@ -1,7 +1,3 @@
-/-
-Copyright (c) 2026 Ironwood Contributors.
-Released under the Apache License, Version 2.0.
--/
 import Zcash.Arithmetic.Field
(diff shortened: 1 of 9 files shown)


TalDerei (approved):
ACK

PR #130, opened 2026-07-28
chore(fv): authenticate native axiom provenance

TalDerei:
## Context

Request changes. Live head `dc0f1c7` has one blocking correctness bug from #122.

**P1 — `+native` provenance is forgeable.** [`checkNativeAllowance`](https://github.com/zcash/ironwood/blob/dc0f1c7713130a8693dadee80795a7f125ec0a81/Zcash/Meta/AxiomCheck.lean#L65-L87) derives ownership solely from an axiom’s textual name and never resolves the alleged owner.

This compiles successfully:

```lean
axiom Spoof.owner._native.native_decide.ax_1_1 : False
theorem Spoof.victim : False := Spoof.owner._native.native_decide.ax_1_1

assert_axioms Spoof.victim +native(Spoof.owner)
```

`Spoof.owner` is not even a declaration, yet the false theorem passes the census. Thus an arbitrary axiom disguised as `…._native.native_decide.…` bypasses the new trust check.

The owner should be resolved and the axiom validated against that actual owner’s dependency footprint; add this as a negative regression test.

Otherwise, the 659 migrated assertion targets and eight guards preserve their targets, and all four census roots built successfully locally.

## Fix

- Resolve every `+native(...)` owner and require its spelling to be fully qualified.
- Require each admitted axiom to occur in the owner’s own transitive axiom footprint, come from the same module, and have the source range Lean recorded inside the owner declaration.
- Recognize the last `_native.native_decide` marker so owner names cannot be truncated by an earlier marker-like component.
- Add a default-built, test-only `MetaCheck` library covering a genuine certificate and three forged-name variants: nonexistent owner, unrelated owner, and an owner that depends on a separately declared forged axiom.

## Verification

- `lake build Zcash.Meta.AxiomCheck`
- `lake build MetaCheck`

Co-authored-by: Claude Fable 5.

### DIFF
--- a/Zcash/Meta/AxiomCheck.lean
+++ b/Zcash/Meta/AxiomCheck.lean
@@ -13,6 +13,10 @@ declared tier (a `sorry`, an unexpected axiom, or `native_decide` where none was
 
 The `#guard_msgs`-pinned form remains the right tool when the *exact* axiom set is the claim.
 
+Scope: these commands catch *inadvertent* drift — an entry silently reaching more than it
+claims. They are not a defense against a deliberately deceptive author, and there may be
+obscure elaboration corners in which an axiom is still misattributed.
+
 Both commands require their argument to be written fully qualified (`checkFullyQualified`):
 an unqualified name resolves through the census file's `open`s, so a same-base-name cousin in
 an opened namespace can silently capture an entry meant for a declaration that is not in scope
@@ -44,7 +48,7 @@ def nativeAxiomOwner? (ax : Name) : Option Name :=
   go [] ax.components none
 
 /-- A syntactically plausible `native_decide` auxiliary axiom. `checkNativeAllowance` additionally
-authenticates its owner, dependency, module, and source range before permitting it. -/
+checks its owner, dependency, module, and source range before permitting it. -/
 def isNativeDecideAxiom (n : Name) : Bool :=
   (nativeAxiomOwner? n).isSome
 
@@ -102,7 +106,7 @@ def nativeAxiomsOwnedBy (owner : Name) : CommandElabM (Array Name) := do
 
 /-- `+native` must name the owning declaration(s) of exactly the `native_decide` axioms the
 entry actually reaches, fully qualified. A bare `+native` would permit a `native_decide`
-axiom smuggled in by *any* declaration entering the dependency cone; naming the owners makes
+axiom brought in by *any* declaration entering the dependency cone; naming the owners makes
 the census state precisely which native certificates are trusted, and a new native axiom —
 or a stale annotation — fails the build with the list to write. -/
 def checkNativeAllowance (n : Ident) (axs : Array Name) (allowed : Option (Array Name)) :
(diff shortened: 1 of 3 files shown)


daira:
Okay, but the axiom check explicitly disclaims an adversarial model? It's trying to find inadvertent reliance on more than was claimed. Here `Spoof.victim` *does* depend on `+native(Spoof.owner)` and that's why it's unsound. I'm fine with the changes in this PR, I just don't think we should be implying that checks like this can close all cases. If I'd designed Lean then maybe it would be possible, but I didn't :-)

Zcash/Meta/AxiomCheck.lean
```diff
@@ -28,43 +29,86 @@ namespace Zcash.Meta
...
+    match rest with
+    | a :: b :: tail =>
+      let ownerRev? := if a == `_native && b == `native_decide then some prefixRev else ownerRev?
+      go (a :: prefixRev) (b :: tail) ownerRev?
+    | _ => ownerRev?.map fun ownerRev =>
+      ownerRev.reverse.foldl Name.append Name.anonymous
+  go [] ax.components none
+
+/-- A syntactically plausible `native_decide` auxiliary axiom. `checkNativeAllowance` additionally
+authenticates its owner, dependency, module, and source range before permitting it. -/
```
daira: "authenticates" overstates it.

Zcash/Meta/AxiomCheck.lean
```diff
@@ -74,17 +118,37 @@ def checkNativeAllowance (n : Ident) (axs : Array Name) (allowed : Option (Array
...
-/-- Extract the annotation list from an optional `+native(A, B)` flag: `none` when the flag
-is absent, `some names` with the parenthesized list otherwise. -/
-def nativeAnnotation (native : Option (TSyntax ``nativeFlag)) : Option (Array Name) :=
-  native.map fun stx => stx.raw[2].getSepArgs.map (·.getId)
+/-- Extract the identifiers from an optional `+native(A, B)` flag. -/
+def nativeAnnotation (native : Option (TSyntax ``nativeFlag)) : Option (Array Ident) :=
+  native.map fun stx => stx.raw[2].getSepArgs.map fun arg => ⟨arg⟩
+
+/-- Resolve and fully qualify every owner named by `+native`. Besides preventing namespace
+capture, resolution ensures an annotation cannot invent a nonexistent owner whose text happens
```
daira: "cannot" is too strong I think. We should be framing this as "there might be obscure cases where this could happen inadvertently", and avoid any implication of "we're trying to prevent an attacker from doing sneaky axiom-smuggling".

daira (approved):
ACK modulo comments about framing.

daira (approved):
[Claude] The spoof is real and this closes it. `nativeAxiomOwner` derived the owner from the axiom's textual name and `isNativeDecideAxiom` only matched the marker, so a hand-declared `axiom Foo._native.native_decide.ax` with `+native(Foo)` passed the census with no check that `Foo` exists or owns it. Good catch on our own #122 macro.

The fix is well-designed. The four authentication conditions in `nativeAxiomsOwnedBy` — the axiom in the owner's transitive footprint, the owner-prefix match, same module, and start-position inside the owner's declaration range — together defeat all three spoofs. The range check is the load-bearing one: a hand-written `axiom` is its own top-level command declared before the owner it feeds, so its start never lands inside the owner's range, while a compiler-emitted `native_decide` auxiliary does. The `nativeAxiomOwner?` last-marker fix is a genuine correctness improvement too. CI green, and `MetaCheck` is a default target so the four negative/positive cases actually run.

Two notes, neither blocking:

- Threat-model boundary. The range check authenticates against accidental name collision and the tested spoofs. A fully adversarial source author with a custom elaborator could in principle emit an axiom with a faked declaration range inside the owner; that seems out of scope here, consistent with the "not attempting to cover adversarial injection" framing on the native-opt-in check. Worth a one-line confirmation that casual/accidental forgery is the intended threat model.

- Cross-package owners. Several real census entries name owners in the CompElliptic pin (e.g. the Pallas curve-order witnesses that the deployed reductions inherit), so `nativeAxiomsOwnedBy` relies on imported `DeclarationRange`s being present in the `.olean`. Your note that the 659 migrated targets still pass covers the ones on `main`; #124 adds more cross-package `+native` entries, and I will confirm those still pass when I rebase it over this.

Approving.

PR #131, opened 2026-07-28
feat(fv): RedDSA obligations

TalDerei:
RedDSA, formalized abstractly, and the two signature obligations the ledger games consume — the Spend Authority forgery arm's ε and knowledge of `bsk` for the binding signature — stated with their failures as data. Originally @TalDerei's branch, handed off and extended; now based on main with #139 merged.

## Front fix

The first commit corrects an overclaim that shipped with #139: the Orchard capstone docs described `ε_bindsig` as "the binding-signature advantage" when nothing in the tree connects the conservation events to a binding-signature break. The connection begins below; the front commit makes the base's snapshot accurate first (#107 tracks the remaining glue).

## RedDSA and the obligations

`Security/RedDSA/Basic.lean`: an abstract RedDSA scheme over an `(F, G)`-module — the verification equation, the spec's re-randomization axioms proven at this instantiation, completeness, and special soundness (`forkDlog_spec`, `verify_fork_dlog`). `RedDSA/SURK.lean`: the SURK-CMA win as data with the ± extension, and the game definition audited in the module doc — the single-oracle simplification is an identity-randomizer repackaging, and a key-blind challenge hash provably forfeits the game (`verify_shifted_of_key_blind_hash`), so the key-prefixing is load-bearing.

`Ledger/ForgeryArm.lean`: with the signature primitives pinned to RedDSA (`SpendAuthShape`), every forgery-arm sample exhibits a SURK-CMA win against `akV` or `−akV`, so the arm's ε is the SURK-CMA advantage at the two ± keys (`spendAuthority_measure_le_surk` — the design's factor-of-2 made formal). `Ledger/ExtractionArm.lean`: the transaction-balance premiss discharge restated with a fallible extractor — nothing assumed of it, failures exhibited as `RedDSA.ExtractionFailure` — so the conservation capstones are bounded by `εdlr + κ` (`balanceConservation_measure_le_kerr`, `shieldedBalanceCap_measure_le_kerr`).

[...]

PR #139, opened 2026-07-28
fv: the all-prefixes probabilistic capstones and Orchard bounds

daira:
On top of the merged docs-and-naming PR: the generic all-prefixes probabilistic capstones, their instantiation at the Orchard-protocol bases and parameters, the Orchard Spend Authority probability bound, and the book's security-definitions page brought up to date.

## The all-prefixes probabilistic capstones

Balance-subset, Balance-conservation, the shielded-balance cap, and Balance integrity over every prefix below a bound, at no extra cost in the bound: each ε hypothesis is named on one break event shared across prefixes — every prefix's violation lands in it — rather than summed per prefix. A naive union bound would pay `k · ε`.

The all-prefixes events follow EWD 831: half-open ranges `i < k` over unshifted families, with the exclusive bound as the parameter. The value-side events are prefix-indexed and named `*Before`; the Balance-subset events are step-indexed and keep `*UpTo`. The one step/prefix crossing — non-negativity after the first `i + 1` transactions rests on subset step `i` — is confined to lemmas carrying `_succ` in their names; every other statement is same-index, and the empty prefix cannot violate anything.

Balance integrity is the intended property, mirroring `balanceIntegrityOrBreak`'s conclusion: the shielded pool is non-negative and the pools sum to the minted issuance (the transparent conjunct cannot fail on the valid sample space). Its bound composes a non-negativity ε with the transaction-balance premiss arm's `ε_conservation`; the earlier interval form `pool ∈ [0, issuanceTotal]` is the weaker consequence and survives only as the separate shielded-balance-cap capstones.

## The Orchard probability bounds

The relation event is the branch preimage of `orchardBalanceSubsetOrRelation`, and every Balance capstone violation is bounded by it: one `ε_sinsemilladlr` replaces the abstract capstones' three per-arm ε's, and the all-prefixes bounds cost no factor of `k`, because the arms and the prefixes all land in the same event — a nontrivial discrete-log relation among the same fixed Sinsemilla bases. `ε_sinsemilladlr` reduces tightly to discrete-log hardness of those bases (Jaeger–Tessaro, eprint 2020/1213 Lemma 3, re-proved as `relation_prob_le_of_textbookDL`). The witness-level model abstracts away Halo 2 knowledge soundness, a separate, lossy reduction on the different Halo 2 bases; that is where the end-to-end reduction loss lives. `ε_bindsig`, the binding-signature advantage, bounds the conservation side (its RedDSA discharge is #22). The census pins the new bounds with the reduction's native provenance.

## The Orchard Spend Authority bound

At the Orchard `keyBinding` interface, the Spend Authority reduction's key-binding break is the same `CommitIvkCollision` that the Balance key-binding arm reduces. `relationOfSpendAuthorityKBBreak` sends it to the same discrete-log terminal at the `CommitIvk` bases, `orchardSpendAuthorityOrRelation` composes that conversion into the Spend Authority reduction, and `orchardSpendAuthority_measure_le` names the key-binding arm's hypothesis on the composed reduction's relation event — so the Spend Authority violation (some Action spends a note addressed to the victim key over a sighash its holder never signed) is bounded by the forgery arm's ε plus a discrete-log-relation advantage, with no oracle model. The forgery arm's ε is RedDSA ±-randomized unforgeability, a named hypothesis (#22). Connecting the named ε's to adversary experiments — the oracle-machine layer, including recovering the firing Action from the adversary's announced indices — is tracked in #155.

## The book

The connected picture's top-level nodes carry the probabilistic capstones where they exist, the Orchard-protocol key-binding arm gains its edge to the Sinsemilla discrete-log-relation terminal, and that terminal moves from named-hypothesis to machine-checked (its inbound reductions and the tight onward reduction to discrete log are all proven; RedDSA stays a named hypothesis). A new glossary section covers Balance integrity's definition, events as branch preimages, the all-prefixes no-factor-of-`k` bounds, and the Orchard single-ε collapse. The Key binding node also drops its blanket ROM tag: the model split is visible on its outgoing edges — the ZIP 2005 route through the shifted-oracle collision, and the oracle-free Orchard-protocol `CommitIvkCollision` route.

Full build green at every commit.

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Security/Ledger/OrchardCapstone.lean
+++ b/Zcash/Security/Ledger/OrchardCapstone.lean
@@ -292,7 +292,10 @@ def orchardSpendAuthorityOrRelation
 /-- The Orchard Spend Authority relation event, for a victim key witness `wV` with
 signing history `Signed`: the samples on which the reduction, run on an Action
 spending a note addressed to `wV` over an unsigned sighash, computes a nontrivial
-discrete-log relation at the `CommitIvk` bases. -/
+discrete-log relation at the `CommitIvk` bases. The firing Action is selected
+existentially; #155 tracks the oracle-machine layer that connects the ε named on
+this event to an adversary experiment and recovers the Action from the adversary's
+announced indices. -/
 def orchardSpendAuthorityRelationEvent (wV : KeyBinding.Pool.Witness Fq PallasGroup Fp)
     (hKB : keyBinding.KB wV) (Signed : MSG → Prop) :
     Set (OrchardAnnotated verify bverify issuance maxActions) :=
--- a/Zcash/Security/Ledger/OrchardCapstone.lean
+++ b/Zcash/Security/Ledger/OrchardCapstone.lean
@@ -270,4 +270,74 @@ def relationOfSpendAuthorityKBBreak (b : KeyBindingBreakData keyBinding) :
     NontrivialRelation (F := Fq) pallasS ivkQpt commitIvkRpt :=
   relationOfKeyBindingBreak b.h
 
+/-- The Orchard Spend Authority reduction with its key-binding arm routed to the
+discrete-log terminal: as `spendAuthorityOrBreak`, with a key-binding break converted
+to its nontrivial relation at the `CommitIvk` bases by
+`relationOfSpendAuthorityKBBreak`. -/
+def orchardSpendAuthorityOrRelation
+    {ledger : Ledger _ Fq PallasGroup Fp Fp Fp Encoding MSG SIG _}
+    (hval : ValidLedger (primitives (MSG := MSG) (SIG := SIG) verify bverify) keyBinding
+      issuance maxActions ledger)
+    {tx} (htx : tx ∈ ledger) {a} (ha : a ∈ tx.actions)
+    {wV : KeyBinding.Pool.Witness Fq PallasGroup Fp} (hKB : keyBinding.KB wV)
+    (hrecv : a.w.note_old.pkd
+      = (primitives verify bverify).emb (keyBinding.ivk wV) • a.w.note_old.gd)
+    {Signed : MSG → Prop} (hfresh : ¬ Signed tx.sighash) :
+    SpendAuthForgery (primitives verify bverify) (keyBinding.akP wV) Signed
+      ⊕' NontrivialRelation (F := Fq) pallasS ivkQpt commitIvkRpt :=
+  match spendAuthorityOrBreak hval htx ha hKB hrecv hfresh with
+  | .inl f => .inl f
+  | .inr b => .inr (relationOfSpendAuthorityKBBreak b)
+
+/-- The Orchard Spend Authority relation event, for a victim key witness `wV` with
+signing history `Signed`: the samples on which the reduction, run on an Action
+spending a note addressed to `wV` over an unsigned sighash, computes a nontrivial
+discrete-log relation at the `CommitIvk` bases. -/
+def orchardSpendAuthorityRelationEvent (wV : KeyBinding.Pool.Witness Fq PallasGroup Fp)
+    (hKB : keyBinding.KB wV) (Signed : MSG → Prop) :
+    Set (OrchardAnnotated verify bverify issuance maxActions) :=
+  {ω | ∃ tx, ∃ htx : tx ∈ ω.1, ∃ a, ∃ ha : a ∈ tx.actions,
+    ∃ hrecv : a.w.note_old.pkd
+      = (primitives verify bverify).emb (keyBinding.ivk wV) • a.w.note_old.gd,
+    ∃ hfresh : ¬ Signed tx.sighash, ∃ r,
+    orchardSpendAuthorityOrRelation verify bverify issuance maxActions ω.2 htx ha hKB
+      hrecv hfresh = .inr r}
+
+/-- A break-arm sample lands in the relation event: on it, the reduction's key-binding
+break converts to a relation, so the composed reduction returns one. -/
+theorem spendAuthorityBreakEvent_subset_relation
+    (wV : KeyBinding.Pool.Witness Fq PallasGroup Fp) (hKB : keyBinding.KB wV)
+    (Signed : MSG → Prop) :
+    spendAuthorityBreakEvent (P := primitives verify bverify) wV hKB Signed
+      ⊆ orchardSpendAuthorityRelationEvent verify bverify issuance maxActions wV hKB
+          Signed := by
+  rintro ω ⟨tx, htx, a, ha, hrecv, hfresh, b, hb⟩
+  refine ⟨tx, htx, a, ha, hrecv, hfresh, relationOfSpendAuthorityKBBreak b, ?_⟩
+  unfold orchardSpendAuthorityOrRelation
+  rw [hb]
+
+/-- **The Orchard Spend Authority probability bound.** For any adversary over valid
+Orchard ledgers, the probability that some Action spends a note addressed to `wV` over
+a sighash the holder of `wV` never signed is at most the forgery arm's ε plus the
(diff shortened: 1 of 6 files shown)


daira (commented):
Self-ACK

TalDerei:
Comparing against my previous comments, PR #139 resolves the Balance-subset event-level gap, but its Spend Authority DLR discharge seems missing.

### Blocking soundness finding

`relationOfSpendAuthorityKBBreak` correctly converts an explicit break into a DLR witness, but nothing consumes it. `deployedSpendAuthority_measure_le` still accepts the old opaque bound

`measure(spendAuthorityBreakEvent) ≤ εkb`

and merely calls the generic `spendAuthority_measure_le`. It neither defines a DLR output event nor derives `hkb` from a DLR-hardness bound. [DeployedCapstone.lean](https://github.com/zcash/ironwood/blob/8d7764281055e49b95a44980d7aa58d79d82b9e4/Zcash/Security/Ledger/DeployedCapstone.lean#L141-L162), [generic theorem](https://github.com/zcash/ironwood/blob/8d7764281055e49b95a44980d7aa58d79d82b9e4/Zcash/Security/Ledger/Capstone.lean#L227-L262)

Therefore the PR body, theorem docstring, and TrustBoundary note incorrectly say that `εkb` is now the CommitIvk DLR advantage. This is blocking relative to one of PR #139’s headline claims.

The fix needs a computed Spend Authority relation event/finder and a containment such as:

`spendAuthorityBreakEvent ⊆ spendAuthorityRelationEvent`

followed by `measure_mono`, with the theorem taking a bound on that relation event. Merely having `b ↦ relationOfSpendAuthorityKBBreak b` is insufficient because the current ledger event existentially hides which action and break fired.

### Previous comments

- [Comment 5099193054](https://github.com/zcash/ironwood/pull/124#issuecomment-5099193054): the duplicate definition, build-root coverage, injectivity hypotheses, and Balance consumer are resolved. The relation-to-textbook-one-DL step remains an explicit assumption.
- [Comment 5102451614](https://github.com/zcash/ironwood/pull/124#issuecomment-5102451614): resolved for Balance-subset at the PMF event layer. `deployedRelationEvent` is the preimage of the total reducer, and the new theorem correctly proves violation-event containment. [Balance theorem](https://github.com/zcash/ironwood/blob/8d7764281055e49b95a44980d7aa58d79d82b9e4/Zcash/Security/Ledger/DeployedCapstone.lean#L70-L101)
- The standard computational-efficiency bridge remains unformalized; the PR explicitly acknowledges that this layer does not model runtime. I treat that as a declared boundary, not a hidden bug.
- The old merge-readiness problem is resolved: #139 is clean, mergeable, and CI-green.

### Other audit buckets

- Specification: no encoding, fixed-base, domain, indexing, or Spend Authority semantic mismatch found.
- Implementation correctness: rename is complete; reducers and Balance probability/value proofs look correct. Exact-head `lake build --wfail Zcash.Security.Ledger.DeployedCapstone` passed, as did `git diff --check`.
- Soundness: Balance-subset is sound within the stated event-level model. Spend Authority is not yet connected to its claimed DLR terminal.

Verdict: **request changes** for the Spend Authority probability composition. No other concrete defect found.

daira (commented):
Self-ACK

Zcash/Security/Ledger/OrchardCapstone.lean
```diff
@@ -65,4 +76,268 @@ def orchardBalanceSubsetOrRelation {ledger : Ledger _ Fq PallasGroup Fp Fp Fp En
...
+discrete-log relation at the `CommitIvk` bases. -/
+def orchardSpendAuthorityRelationEvent (wV : KeyBinding.Pool.Witness Fq PallasGroup Fp)
+    (hKB : keyBinding.KB wV) (Signed : MSG → Prop) :
+    Set (OrchardAnnotated verify bverify issuance maxActions) :=
+  {ω | ∃ tx, ∃ htx : tx ∈ ω.1, ∃ a, ∃ ha : a ∈ tx.actions,
+    ∃ hrecv : a.w.note_old.pkd
+      = (primitives verify bverify).emb (keyBinding.ivk wV) • a.w.note_old.gd,
+    ∃ hfresh : ¬ Signed tx.sighash, ∃ r,
+    orchardSpendAuthorityOrRelation verify bverify issuance maxActions ω.2 htx ha hKB
+      hrecv hfresh = .inr r}
```
TalDerei, line 307: existentially selects the transaction, action, and proofs (prop-based). shouldn't this invoke `orchardSpendAuthorityOrRelation` (otherwise it's not computable)?

Zcash/Security/Ledger/OrchardCapstone.lean
```diff
@@ -65,4 +76,268 @@ def orchardBalanceSubsetOrRelation {ledger : Ledger _ Fq PallasGroup Fp Fp Fp En
...
+    (hbind : A.toOuterMeasure (balanceConservationViolationBefore
+        (P := primitives verify bverify) (kv := keyBinding) (issuance := issuance)
+        (maxActions := maxActions) k) ≤ ε_bindsig) :
+    A.toOuterMeasure (balanceIntegrityViolationBefore (P := primitives verify bverify)
+        (kv := keyBinding) (issuance := issuance) (maxActions := maxActions) k)
+      ≤ ε_sinsemilladlr + ε_bindsig :=
+  le_trans
+    (toOuterMeasure_le_add₂ A
+      (balanceIntegrityViolationBefore_subset_relation verify bverify issuance maxActions k))
+    (add_le_add hsin hbind)
```
TalDerei, line 257: this bound is assumed rather than going through concrete binding-signature break event.

TalDerei (commented):
some events still assert an existential witness.

Zcash/Security/Ledger/OrchardCapstone.lean
TalDerei: intended to be handled by #131 (and will need to modify `OrchardCapstone.lean`).

daira:
Status against this review, after the hand-off work. The blocking Spend Authority finding was right, and it survived the branch reconstruction until we re-checked this review: the rebuilt bound had reproduced the opaque `εkb` shape. Fixed on the branch now: `orchardSpendAuthorityOrRelation` composes the break conversion into the reduction, `orchardSpendAuthorityRelationEvent` is its relation-branch preimage, `spendAuthorityBreakEvent_subset_relation` is the containment, and `orchardSpendAuthority_measure_le` takes its key-binding hypothesis on the relation event. Also since this review: the relation-to-textbook-DL step is proven (`relation_prob_le_of_textbookDL`), and the efficiency bridge remains a declared boundary. Paths moved: DeployedCapstone.lean → OrchardCapstone.lean.

Zcash/Security/Ledger/OrchardCapstone.lean
daira: The probability layer's events are Prop-level branch preimages; the computable object is the reducer, and `orchardSpendAuthorityRelationEvent` is defined as `orchardSpendAuthorityOrRelation`'s `.inr`-branch preimage at the quantified action. The existential selection of which Action fires is the known gap shared with the Model's Spend Authority events — the composite machine recovers it oracle-free from the adversary's announced indices (`kwAt`; see the census's key-binding-arms section). Tracked in #155 (the oracle-machine layer) rather than this PR.

Zcash/Security/Ledger/OrchardCapstone.lean
daira: Agreed, and confirmed for the stacked #131 branch: the extractor-plus-knowledge-error restatement there is Model-level so far, and its Orchard-protocol instantiation — routing the conservation arm's ε through the binding-signature relation and extraction-failure events — will modify `OrchardCapstone.lean` on that branch.

TalDerei (approved):
ACK

PR #141, opened 2026-07-28
fv: derive the domain root's order from the certificate, not a native bridge

mitschabaude:
`omegaOf_eq_certifiedRootPow` proved, by one `native_decide`, that all 33 supported `omegaOf k = powFast 5 ((p−1)/2^k)` agree with powers of CompElliptic's certified Pasta root — 33 full-width native modular exponentiations standing in for one missing provenance fact. This PR removes the theorem and its `native_decide` entirely.

`omegaOf` is now defined *from* the root: `rootOfUnityFp` is the Pasta `ROOT_OF_UNITY` hex literal, `omegaOf k := powFast rootOfUnityFp (2^(32−k))` — squaring down exactly as `EvaluationDomain::new` does. `rootOfUnityFp_eq_certified` ties the literal to `pallasBase.rootOfUnity` by `rfl` (the two numerals are definitionally equal), and `omegaOf_isPrimitiveRoot` derives the primitive-root fact from the certificate's `rootOfUnity_order` through it. The generator-5 provenance moves to the docstring; faithfulness to the deployed key's omega remains pinned empirically by `VkMatch` against the captured VK, and all captured-key native checks decide the same values (the definition change is value-preserving).

Why the intermediate literal instead of naming the CompElliptic constant in the definition: `assert_axioms` collects axioms through every referenced constant, and `pallasBase` carries its validity proof — with the certificate's own `native_decide` axiom — inside its value. Defining `omegaOf` directly from `pallasBase.rootOfUnity` put that axiom into the `+native` lists of two dozen pure definitions (`vkAt`, the semantic failure events, …). The bare literal keeps definitions axiom-clean; only proofs that actually use the certificate's order fact reach `pallasBase`, exactly as before.

Census diff is pure removals: the eight `+native` entries naming `omegaOf_eq_certifiedRootPow` drop that name and nothing gains an axiom.

Verified locally with `lake build Zcash CircuitCheck` (includes the census and the recomputed native checks in `Keygen/Certificate` and `SingleAction/Fixture`); the remaining `FixtureCheck` lane is left to CI.

### DIFF
--- a/Zcash/Arithmetic/Domain.lean
+++ b/Zcash/Arithmetic/Domain.lean
@@ -5,9 +5,9 @@ import Zcash.Arithmetic.Field
 # Evaluation-domain scalars (pasta `Fp` constants and facts)
 
 halo2's domain data as pure functions: binary exponentiation (`powFast`), the size-`2^k`
-domain root of unity `omegaOf` (pasta `Fp::GENERATOR = 5`, `ROOT_OF_UNITY` squared down,
-`EvaluationDomain::new`), `Fp::DELTA`, and the domain facts (primitive-root, power
-injectivity, size nonvanishing) bridged once to CompElliptic's certified Pasta root.
+domain root of unity `omegaOf` (CompElliptic's certified Pasta `ROOT_OF_UNITY` squared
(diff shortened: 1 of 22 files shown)


mitschabaude:
Moved onto #140 (cf8c039) at Gregor's request.

PR #143, opened 2026-07-29
Minimize soundness-stack Mathlib imports

mitschabaude:
## What changed

- replace umbrella `import Mathlib` statements throughout the soundness-stack dependency closure with `Mathlib.Tactic` or narrower imports
- add precise imports for downstream consumers that had relied on those transitive umbrella imports
- make `Zcash.Common.RelationWitness` use core `Nat`, eliminating its Mathlib dependency entirely

## Why

Importing the Mathlib umbrella made each affected Lean process peak around 6.5 GB RSS. During a parallel Lake build, several such processes could create severe memory and GC pressure; one observed `PermutationInstantiation` build took 648 seconds despite compiling in about five seconds in isolation.

For `PermutationInstantiation`, isolated peak RSS dropped from 6.57 GB to 3.77 GB and wall time dropped from 5.42 seconds to 4.00 seconds.

## Validation

- `lake build` on a fresh branch based on `origin/main`
- all 8,961 jobs completed successfully

### DIFF
--- a/Zcash/Arithmetic/Domain.lean
+++ b/Zcash/Arithmetic/Domain.lean
@@ -1,4 +1,5 @@
 import CompElliptic.Curves.Pasta
+import Mathlib.RingTheory.RootsOfUnity.PrimitiveRoots
 import Zcash.Arithmetic.Field
 
 /-!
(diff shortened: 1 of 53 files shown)


mitschabaude:
Superseded by #144, using a branch directly in zcash/ironwood.

PR #144, opened 2026-07-29
Minimize soundness-stack Mathlib imports

mitschabaude:
- replaces umbrella `import Mathlib` statements throughout the soundness-stack dependency closure with `Mathlib.Tactic` or narrower imports
- add precise imports for downstream consumers that had relied on those transitive umbrella imports

## Why

Importing the Mathlib umbrella made each affected Lean process peak around 6.5 GB RSS. During a parallel Lake build, several such processes could create severe memory and GC pressure; one observed `PermutationInstantiation` build took 648 seconds despite compiling in about five seconds in isolation.

For `PermutationInstantiation`, isolated peak RSS dropped from 6.57 GB to 3.77 GB and wall time dropped from 5.42 seconds to 4.00 seconds.


TalDerei (approved):
optimistic ACK on passing CI.

should we maybe consider adding a CI guard rejecting bare umbrella imports?

mitschabaude:
> should we maybe consider adding a CI guard rejecting bare umbrella imports?

good idea, done! (it caught many more existing umbrella imports)

PR #149, opened 2026-07-29
fv: suppress the library-suggestions indexer's heartbeat panic

daira:
On toolchains before v4.31.0, the `symbolFrequency` environment extension's export function runs `whnf` over every local theorem signature with the default heartbeat limit and panics on a timeout (leanprover/lean4#12989, fixed upstream by leanprover/lean4#13202); the limit cannot be raised from outside. The deployed AGM modules' signatures exceed it, so builds print panics and the Lean InfoView becomes unusable.

This adds a `Zcash` entry to the library-suggestions name deny list in a new module that the offending modules import, so the indexer skips their declarations before reading the signatures. Only opt-in suggestion-driven tactics consult the deny list — nothing in the repository enables a suggestion selector — and `exact?`, `apply?`, and `aesop?` use separate machinery and are unaffected. The export fold only visits the current module's declarations, so the entry cannot affect anything outside this repository.

Full build green with zero panics (previously two per build). The suppression is only needed before v4.31.0 and can be removed when the toolchain moves (Clean is expected to port to v4.31.0).

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Meta/LibrarySuggestionsDeny.lean
+++ b/Zcash/Meta/LibrarySuggestionsDeny.lean
@@ -1,0 +1,26 @@
+import Lean
+
+/-!
+# Suppress the library-suggestions symbol-frequency indexer
+
+On toolchains before v4.31.0, the `symbolFrequency` environment extension's export
+function runs `whnf` over every local theorem signature with the default heartbeat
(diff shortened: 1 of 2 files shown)


daira (commented):
Self-ACK. Tested locally.

PR #150, opened 2026-07-30
chore(fv): axiom laundering bypass

TalDerei:
**AI Summary:**

This PR closes a bypass of the build-time axiom census, extends the census to the endpoints it silently missed, and removes two avoidable compiler-trust axioms. I'd recommend reviewing this commit by commit.

## What Changed

- **Provenance bypass:** `+native(…)` accepted a macro-emitted `axiom` as a genuine `native_decide` certificate. Macro expansion gives every emitted declaration the invocation site as its range, so an auxiliary-named `axiom` and the theorem using it shared one range and passed the non-strict check automatically — laundering any axiom, up to `False`, past the census. Provenance now requires the auxiliary to start strictly after its owner.
- **Census regressions:** `MacroForged` pins that rejection. `SecondCertificate` and `AliasedOwner` pin two properties no existing case exercised: a stale `+native` must fail when a second certificate enters the cone, and a declaration whose own name imitates the marker path must not collapse onto the legitimate owner.
- **Uncovered endpoints:** `collectAxioms` walks a declaration's dependencies, so a top-level capstone nothing pinned depends on was checked by nothing. `scripts/check_endpoint_census.sh` requires a direct entry per endpoint and pins the 13 it found uncovered.
- **Wrong census tier:** the plain-`def` statement-or-relation terminals and the three `AlgebraicWfProof` IPA dichotomies move from `assert_axioms` to `assert_computable` — the tier that rejects `noncomputable`, and so witnesses that break coefficients were computed rather than conjured from propositional existence via choice. Several carried no entry at all.
- **Vacuity as a theorem:** `nonempty_nontrivialRelation_vesta` exhibits a nontrivial discrete-log relation among arbitrary Vesta generators from the group order alone. The `⊕'`-with-data shape is therefore necessary but not sufficient, which makes the computable pin load-bearing rather than stylistic.
- **Trusted base:** `pallas_base_card_lt_scalar_card` and the Pallas `neg_five_not_isCube` become kernel proofs (`decide`, and the `reduce_mod_char` technique its Vesta twin already used), dropping the now-stale `+native` entries naming them and taking both certificates out of the ledger capstone's trusted base.
- **Degenerate generator:** the straight-line DLOG profiles now require `B ≠ 0`. At `B = 0` the programmed basis collapses, so `hardness` was satisfiable only at advantage ≈ 1 and every bound it fed went trivial while still reading as a hardness premise.

## Remaining Floor

Computable extraction at the deployed artifacts is still a conjunction of two theorems rather than one: the rewind route is `noncomputable` because `openedMemberDecode_of_x1Prob` turns an accept-measure bound into transcripts by `Classical.choose`, and the computable straight-line route lives at a quantified basis instead. Joining them needs a straight-line family reproducing the captured proof, which does not exist yet.

CompElliptic's inherited compiler trust is now exactly its two curve point counts and the Tonelli–Shanks root-of-unity data; every other `+native` owner is first-party. Provenance rejects the macro forgery but cannot reject deliberate declaration-range manipulation under `run_cmd`, which has no innocuous reading in a diff.

Every assertion above is enforced by `lake build` (the `Lean` job), not by review.


mitschabaude (approved):
ACK for the new tests and trust boundary changes

Issue #155, opened 2026-07-30
Oracle-machine layer for the capstone ε slots


daira:
The ledger capstones name their probability bounds (ε_sinsemilladlr, ε_bindsig, the SURK-CMA ε's, κ) as hypotheses on events over `PMF (ValidAnnotated ...)`. Connecting them to adversary experiments needs the oracle-machine layer: a composite machine that runs the ledger adversary, recovers the arm witnesses oracle-free (`kbPairOf` for Balance; `kwAt` at the announced indices for Spend Authority), and inherits the arm bounds at an unchanged query count — extending what #68 did for the key-binding layer to the full capstone slots. This also covers the existential selection in the arm events (which Action fires), raised in #139's review threads.

🤖 Claude Fable 5

daira:
We should also check for other existentials violating breaks-as-computed-data.

TalDerei:
> We should also check for other existentials violating breaks-as-computed-data.

cross referencing motivation in https://github.com/zcash/ironwood/issues/156.

TalDerei:
From the audit follow-up: when this layer lands, the all-prefix ledger ε premises (`txBalanceBreakEventBefore` and its siblings — see the "from one bound on the shared break event" docstrings in `Security/Ledger/Capstone.lean`) should discharge through a first-bad-prefix machine: run the adversary once, scan for the first violating prefix, emit that break. The run is single-shot and straight-line, so the shared-event bound carries no factor of `k`, and the per-transaction break data the `⊕'` arms already compute is exactly what the machine scans.

daira:
Source check against the current tree (the PR #171 head): the Balance half of this issue is delivered, and the remaining scope is Spend Authority, plus a Spendability precondition.

Delivered by #171:

* The Balance capstone slots (conservation, cap, integrity) now sit over a challenge-oracle experiment: `challengeExperiment`/`sampledLedgerEvent` (`Ledger/ConservationExperiment.lean`), with one combined coin-consuming finder (`conservationRelFinder`) whose discrete-log hypothesis is `TextbookDLWithCoinsAdvantageLE` with the challenge table as coins, at an unchanged query count. Endpoints: `balanceConservationBefore_measure_le_experiment`, `shieldedBalanceCapBefore_measure_le_experiment`, `balanceIntegrityBefore_measure_le_experiment`, and the Orchard/deployed forms `orchardBalanceIntegrity_measure_le_deployed` and siblings.
[...]

PR #157, opened 2026-07-31
fv: RedDSA core, extractability, and the Orchard conservation bounds

### LINKED ISSUE #22: RedDSA extractability

daira:
The parts of the handed-off #131 that survive the SURK-CMA modelling question, rebuilt on a fresh branch: the RedDSA extractability discharge for the balance argument (#22). The binding signature extracted from is the adversary's own, so no honest-party model is needed; the SURK-CMA experiment and the Spend Authority forgery arm are planned for a separate PR (#121, #155).

## ε_bindsig overclaim fix

The front commit corrects an overclaim that shipped with #139: the Orchard capstone docs described `ε_bindsig` as "the binding-signature advantage" when nothing in the tree connects the conservation events to a binding-signature break. They are connected later in this PR.

## RedDSA, abstractly

`Security/RedDSA/Basic.lean`: an abstract RedDSA scheme over an `(F, G)`-module. This includes the verification equation, the spec's re-randomization axioms proven at this instantiation, and completeness under original and re-randomized keys; the SURK-CMA work will share it. The possible discharge routes are described neutrally in the module doc: the random-oracle forking argument, whose loss is essentially optimal for ROM-only reductions from discrete log (Seurin [eprint 2012/029](https://eprint.iacr.org/2012/029)), and the tight straight-line AGM+ROM reduction for plain Schnorr (Fuchsbauer–Plouviez–Seurin [eprint 2019/877](https://eprint.iacr.org/2019/877), section 3, Theorem 1). The latter's adaptation to RedDSA —the key-prefixed hash, the strong notion of freshness, and re-randomized keys— has not been worked out. The doc records why RedDSA hashes the verification key into the challenge (a key-blind hash would let a verifying signature transport across re-randomizations of the key).

## Extraction arm

`Ledger/ExtractionArm.lean` is the transaction-balance premiss discharge restated with a fallible extractor. The conservation and cap capstones are bounded by `εdlr + κ`, per prefix and at all prefixes with no factor of `k`. It is the sibling of the forgery arm to come; the extractor recovers the signing key, and extraction is general to any use of RedDSA.

## Retiring the total-extractor premiss

The earlier total-extraction hypothesis forms (`premissOrBreak`, `conservationOrBreak`, `capOrBreak`) are removed. They assumed a *total* extractor —classically satisfiable in a cyclic group, so vacuous as a security hypothesis— and had only census consumers. The fallible forms above supersede them as specializations of the generic value-break capstone layer, and are the actual discharge rather than a restatement of a deleted thing.

## Orchard conservation arm

`orchardBalanceConservationBefore_measure_le_kerr` and `orchardBalanceIntegrity_measure_le_kerr` instantiate the extraction arm's bounds at the Orchard-protocol primitives: the integrity bound becomes `ε_sinsemilladlr + εdlr + κ`, replacing the opaque `ε_bindsig` with the exhibited binding-signature arms' named bounds. The no-overflow premiss is bounded against the Pallas scalar order `r_ℙ`.

## Value commitment base rename

`R` clashed as a RedDSA signature component and as the value-commitment randomness base (the spec's ℛ). The bases are renamed `Vbase` and `Rbase` throughout the binding-signature and value layers, with 𝒱 and ℛ kept for prose. The book's security-definitions page tracks the change.

## Binding-signature extraction

`Security/RedDSA/Extraction.lean` holds the extraction notions and the deterministic core of the extraction arm's κ discharge (#22). `Extractor` and `ExtractionFailure` (moved in from `Basic.lean`) state the obligation as data: a candidate extractor is a total function and its failures are exhibited, so κ bounds an exhibited event rather than a total extraction hypothesis. The binding signature is a signature of knowledge of `bsk = Σ rcv`, the discrete log of `bvk` with respect to ℛ. The balance argument extracts from the adversary's own signature under an arbitrary key, so this is proof-of-knowledge extraction — not unforgeability, which would be defined for an honestly generated key. In the algebraic group model the signature's commitment `R` and key `bvk` carry coefficient vectors over the `m` public bases the adversary has received (`QueryRep`), with ℛ at a distinguished slot. Substituting them into the Schnorr verification equation assembles coefficients whose evaluation against the basis is zero, so whenever some coordinate is nonzero they are a shared `AlgebraicRelationWitness` (`bindingSig_relation_of_nontrivial`). Extraction failure forces a pivot — a non-ℛ slot with a nonzero key coefficient — and away from the pivot equation's one root (`QueryRep.badChallenge`) some assembled coordinate is nonzero.

## Documenting the reference-string heuristic

The reductions to discrete log treat the value-commitment, Sinsemilla, and inner-product bases as independent random elements, while deployed they are fixed hash-to-curve outputs. The Security Definitions book page gains a section stating the heuristic that carries the random-basis proof to the fixed bases, and its caveat: an adversary has the protocol's whole lifetime to attack that one reference string, and a single discrete-log computation breaks binding chain-wide — the sharp edge of the quantum threat. The caveat is protocol-wide, so it lives in the book; `Extraction.lean` references it.

## Knowledge error bounded

`Security/RedDSA/KnowledgeError.lean` gives κ ≤ (qH + 1)/|F| + (ε_DL + 1/|F|) (`kappaEvent_measure_le`) — the FPS straight-line AGM+ROM extraction in the key-only setting: no signing oracle, none of that proof's simulation terms. The adversary is a `LabeledOracleComp`, the online-AGM machinery the Halo 2 soundness stack already uses: each challenge query carries the adversary's representation of its group elements as a label the oracle never sees, and the output announces the representation of the returned signature. The representation in effect at the output's query point is the run's first annotation there, with the announced one as fallback — the FPS table discipline made structural, closing the trivial-representation trap. The bad-challenge arm is bounded by the existing labeled squeeze (`finalBadWithoutRelation_measure_le`) at (qH + 1)/|F|; the fallback branch plays the game's own final challenge query, so there is no completion step and no hypothesis that the adversary queries its output's point. The relation arm replays the adversary as a computable relation finder, and `relation_prob_le_of_textbookDL` gives the tight Jaeger–Tessaro accounting. The arms live on independent factors of the product sample space, so the fibre-wise Fubini bounds compose them with no cross term. Degeneracy needs no side condition: at base `0` the DL hypothesis itself forces ε ≥ 1 − 1/|F| (`textbookDLAdvantageLE_base_zero`). The random-basis DL game transfers to the deployed fixed bases by the reference-string heuristic above.

## Extraction-failure arm discharged

`Ledger/ExtractionKappaArm.lean` ties the bound into the Balance games, following the key-binding arm's oracle-model pattern: for any `qH`-query-bounded labeled algebraic ledger adversary, the probability that its output ledger is valid and the conservation reduction lands in the extraction-failure arm is at most (qH + 2)/|F| + ε_DL. The primitives at a sample read the challenge hash off the table at the query point and commit at two distinguished slots of the `m` sampled bases (`kappaPrimitivesAt`, a record update, so the ledger type does not depend on the sample). The adversary outputs an annotated ledger — each transaction paired with the announced representation of its commitment and binding key — and being algebraic is the pair of hypotheses that query-time labels and announced representations evaluate to the transactions' actual elements at their query points. The extractor (`kappaExtractor`) reads the key's ℛ coefficient off the representation in effect at the signature's query point; it is defined per sample and obtains that representation from the run's own data, so none appears among the `Extractor` interface's arguments. The failing transaction and its representation are recovered oracle-free from the annotated output (`failTxOfAnn`), so the composite machine is covered by the knowledge-error bound at an unchanged query count; validity is a conjunct of the measured event, and the all-prefixes form costs no factor of `k`.

## Census rejects unneeded `+choice`

`assert_computable X +choice` — the breaks-as-computed-data check, permitting `Classical.choice` — passed even when `X` did not reach `Classical.choice`, so a stale or over-cautious flag silently over-stated the trusted base. It now fails the build (drop the flag), mirroring how `+native` already rejects an owner list that names no present certificate. Placed early, it validates every later census addition; the one flag it surfaced, `Zcash.Snark.finForallOrRelationWitness`, is dropped in the same commit. Every `+choice` this PR adds is genuine, reaching `Classical.choice` only in erased positions.

## Extraction in the connected picture

The security-definitions page's connected picture gains the knowledge-error node (checked). This is fed by the binding-signature balance arm, computing a relation over the presented basis away from each query's one bad challenge, and resting on the challenge-as-random-oracle model, with query-time representation labels pinning each query's one bad challenge before the answer is drawn. Spendability gains two missing dependency edges — on the spend-auth forgery arm (a valid adversarial spend of an honest note is a forgery, per the game's without-the-honest-user proviso) and on Balance integrity (an adversary draining the shielded pool would leave the honest transaction unminable under the ZIP 209 pool-balance rules). The knowledge-soundness boundary is restated in current terms: the named `TopLevelCircuitCorrectness` component conditions, replacing the retired opaque `hencodes` implication. The legend and glossary track the discharge.

## What remains

For #22, nothing beyond the standing assumptions, which are named rather than discharged:

- `ε_DL` — discrete-log hardness at the sampled bases;
- the algebraic-adversary hypothesis — the AGM itself;
- the reference-string transfer to the deployed fixed bases — the documented book heuristic.

For Balance integrity as deployed, four formalizable pieces remain beyond this PR, tracked in #107 (the first three are planned for the next PR):

- the conservation relation arm's oracle-model discharge: `εdlr` is still a named hypothesis. This is the easier sibling of the κ arm — the event already carries the computed relation;
- composing the per-arm oracle-model discharges — key binding, the relation arms, κ — into one experiment instantiating the capstone's single sample space;
- the Orchard/Pallas instantiation of the knowledge-error layer, which is abstract over the group, the query type, and the number of presented bases;
- wiring the ledger model's witness requirement to the SNARK extractor (the dashed knowledge-soundness edge) — separate from the circuit-correctness conditions themselves.

The SURK-CMA experiment and the Spend Authority forgery arm —which needs the honest-party oracle model that Spendability and Spend Authority currently lack— follow in a separate PR (#121, #155).

Closes #22.

Full build green at every commit.

🤖 Claude Fable 5 and Opus 4.8


daira (commented):
Ready for review.

daira:
Critical self-review at tip 72358f61 — gaps, overclaims, and modelling deficiencies, clustered by severity. Fixes for most items are in progress on this branch per triage (noted per item).

**Convention:** a struck-out finding has been addressed by a fix folded into the branch since this review was posted, and no longer applies to the current head. A partially struck finding is addressed only in the struck part.

Current status: all items are addressed. Items 2–10 were fixed by folds into the earlier head; item 1 — and with it the item-2 nuance — by the relayering at the current head, which rebuilds the κ layer on the online-AGM `LabeledOracleComp` machinery with full-basis representations. The first-label table discipline is structural there, so the per-triple normalization dichotomy and the representation-collision reduction are no longer needed, and the collision lemma is gone.

## Substantive (modelling)

1. ~~**Representation basis restricted to**~~ $\{\mathcal{V}, \mathcal{R}\}$ ~~**.** `QueryTags` and `halg` restrict the algebraic adversary's representations to the two value-commitment bases; FPS/AGM representations span all received elements, and a deployed adversary knows every public point (the Sinsemilla bases,~~ $\mathcal{G}$ ~~, the URS). Internally the model is consistent — the experiment hands the adversary only~~ $\mathcal{V}$ ~~and~~ $\mathcal{R}$ ~~— but the transfer to the deployed protocol needs the enlarged basis (the relation machinery is generic over~~ $\iota$ ~~) or an explicit heuristic extension, and neither is documented. *Triage: modelling gap that impedes composability; likely fixable; design to be worked out (extraction failure no longer forces*~~ $A \neq 0$ ~~*once extra coordinates exist, the bad-challenge target needs a pivot rule, and a challenge-independent relation arm appears).*~~

2. ~~**The challenge is a function of the query point; the deployed hash is a function of**~~ $(R, \mathsf{bvk}, m)$ ~~**only.** Representations are forced to be a function of the element triple (`queryOf`), and the embedding of a deployed adversary — normalize it to never re-query a triple under different representations — is sound but written nowhere. There is also something to prove: re-querying under a different representation of the same element itself reveals a nontrivial DLR, so the normalization loses nothing; that argument should be formal, not prose. `queryOf` injectivity is an undocumented instantiation obligation (collisions shrink the covered adversary class through `halg` — vacuity, not unsoundness). *Triage: formalize the representation-collision lemma (`bindingSig_relation_of_representation_collision`); document the rest*.~~

3. ~~**`hdl` is a supremum over all challenge tables.** `∀ O, TextbookDLAdvantageLE … ε` asks one $\varepsilon$ to bound the replayed finder's advantage for *every* table — arbitrary unbounded advice, a non-uniform-flavoured hypothesis strictly stronger than "this reduction's advantage". It is an artefact of the fibre-wise composition; the joint experiment repairs the quantifier (the reduction samples the table lazily). The PR description's body glosses this:~~ $\\varepsilon_{\mathrm{DL}}$ ~~— discrete-log hardness at the sampled bases. *Triage: fix the prose.*~~

## Substantive (gaps)

4. ~~**The arm bound is per-prefix; the all-prefixes κ event is uncovered.** `balanceConservation_extractFailArm_measure_le` fixes the prefix `i` (and `kappaComposite` is a different machine per prefix), but `balanceConservationBefore_measure_le_kerr` consumes `hκ` on the all-prefixes event; a naive union pays the factor of $k$. The repair is the first-imbalanced-transaction stability argument: the failure arm at any prefix breaks at the ledger's first imbalanced transaction — the same transaction for every prefix containing it — so one machine covers all prefixes. *Triage: fix in this PR.*~~

5. ~~**Joint-vs-conditional.** The capstones' `hκ` is over `PMF (ValidAnnotated …)` (conditional-on-valid); the arm bound is~~ $\Pr[\mathrm{valid} \wedge \mathrm{arm}]$ ~~in the oracle experiment — the same bridge #107 records for~~ $\varepsilon_{\mathrm{kb}}$, ~~where the phrasing is "at the reductions' own events". The κ-arm module doc's "discharges it" glosses the distinction. *Triage: doc fix.*~~

## Modelling (minor)

6. ~~**Uniform-challenge idealization.** The model's oracle is exactly uniform on 𝔽; the deployed~~ $H^\circledast$ ~~is 512 bits reduced mod~~ $r_\mathbb{P}$ ($\approx q_H \cdot 2^{-257}$ ~~total statistical slop). Unaccounted; same class as the GenRandom note in `Basic.lean`. *Triage: doc fix.*~~

7. ~~**Degenerate parameters.** At `gen = 0` the DL hypothesis is unsatisfiable below~~ $1 - 1/\vert\mathbb{F}\vert$, ~~so nothing is unsound, but only prose in `AGM/Probability` says so. *Triage: prove non-degeneracy (exhibit the trivial finder forcing*~~ $\varepsilon \geq 1 - 1/\vert\mathbb{F}\vert$ ~~*at `gen = 0`), replacing documentation with a theorem.*~~

8. ~~**`tags` is data.** Over colinear model bases every element has some representation, so `halg` alone is cheap to satisfy with dlog-aware tags; the theorem's content rests on `tags` being the algebraic adversary's actual, efficiently-computed representations, with `tagExtractor` computable relative to them. True to the AGM, but implicit. *Triage: doc fix.*~~

## Test coverage / cosmetic

9. ~~**No negative regression** in `Meta/Tests/AxiomCheck.lean` for the unneeded-`+choice` rejection (the `+native` counterparts have them). *Triage: add.*~~

10. ~~**`kappaPrimitivesAt` samples only `valueCommit`/`bindingVerify`** while the other primitives stay fixed — deliberate for this arm (#31 dissolves the record update), but unstated. *Triage: say so.*~~

## Checked, no finding

The trivial-representation trap is closed (tags precede the oracle's answer; the FPS U-table discipline is structural here). $A \neq 0$ is derived from extraction failure via `halg`, not assumed. The $(q_H+2)/\vert\mathbb{F}\vert + \varepsilon$ arithmetic and the $+1$ completing accounting. The `find?`-vs-fold identification on all three branches. Whole-table sampling sidesteps reprogramming. `hr`/`valueBound` transfer through the record update. The census entries are minimal under the strict `+choice` checker.

## Triage summary

Items 1–3 are documentation-weight now but block honest instantiation later; 4 is the one provable gap and lands in this PR; 5–10 are small. None invalidates "Closes #22" under the issue's own scope — the extractability input in extractor-plus-error form — with the residuals tracked here and in #107.

🤖 Review by [Claude Code](https://claude.com/claude-code) (Claude Fable 5), requested by @daira.

Zcash/Security/RedDSA/KnowledgeError.lean
```diff
@@ -0,0 +1,290 @@
...
+  rw [not_not] at hbad
+  simp only [QueryTags.badChallenge]
+  rw [← neg_div]
+  exact eq_div_of_mul_eq hA (eq_neg_of_add_eq_zero_right hbad)
+
+open Classical in
+/-- The relation finder the discrete-log reduction runs: replay the adversary at the presented
+basis, and if its output verifies on a good challenge, return the extracted relation
+(`bindingSig_relation_of_goodChallenge`, bridged to the AGM relation type). -/
+noncomputable def relFinder (O : Q → F) (b : Fin 2 → M) :
```
TalDerei: should this be a computable def since it references a reduction?

book/src/formal-verification/security-definitions.md
```diff
@@ -125,7 +131,13 @@ This picture is a deliberate approximation, and is likely to change as the forma
 proceeds. The RedDSA node is a named hypothesis rather than a terminal assumption: its
 discharge edge names the reduction for security of signatures with re-randomizable keys
 [<a href="https://eprint.iacr.org/2015/395">FKMSSS2016</a>, section 3], adapted to the
-±-randomized variant, together with forking extraction of the Schnorr witness.
+±-randomized variant, together with forking extraction of the Schnorr witness. The
+binding-signature extractability node, by contrast, is discharged via the straight-line
+AGM+ROM extraction of Fuchsbauer–Plouviez–Seurin
```
TalDerei, line 136: ACK reuses our existing generic adversary model.

Zcash/Security/RedDSA/KnowledgeError.lean
```diff
@@ -0,0 +1,290 @@
...
+    (uniformOfFintype_prod_fiber_bound_right relFiber hrel)
+
+end Composition
+
+section Discharge
+
+variable {Q F M : Type*} [Fintype Q] [DecidableEq Q] [Field F] [Fintype F] [Nonempty F]
+  [AddCommGroup M] [Module F M]
+
+variable (gen : M) (tags : Q → QueryTags F)
```
TalDerei: This currently reuses only the back half of the AGM stack:

- `scalarBasis`
- `AlgebraicRelationWitness`
- `relSet`
- `TextbookDLAdvantageLE`
- `relation_prob_le_of_textbookDL`

See the imports in [`KnowledgeError.lean`](https://github.com/zcash/ironwood/blob/3b95b70b70b97f387c872e1360bd92445e5e460a/Zcash/Security/RedDSA/KnowledgeError.lean#L1-L15).

It does not reuse the existing adaptive AGM execution machinery:

- [`LabeledOracleComp`](https://github.com/zcash/ironwood/blob/3b95b70b70b97f387c872e1360bd92445e5e460a/Zcash/Snark/Soundness/AGM/AdaptiveOnline.lean#L13-L57), which attaches representations at query time while preserving the actual oracle key;
- `AlgebraicPoint` / full-basis coefficient vectors;
- the online representation log.

Instead, this defines its own `QueryTags`, global `tags : Q → QueryTags F`, and `halg` hypothesis. Thus it reuses the AGM relation-to-DL/probability code, but not the standard online AGM adversary representation code.

It need not reuse the Halo2-specific `ComputedAdaptiveOnlineAGMFSFamily` wholesale. The clean repair would reuse generic `LabeledOracleComp` and `AlgebraicPoint`, with a RedDSA-specific label/output type. Otherwise the result should be scoped as a restricted two-base algebraic model, rather than the standard FPS AGM+ROM discharge.

_posted with the help of claude._

TalDerei: this makes sense because we need to show these were algebraically constructed from known public basis points, not just that some representation exists, and we don't currently enforce that here.

Zcash/Security/RedDSA/KnowledgeError.lean
daira: Yes — as of the relayered head it is: `relFinder` is a computable `def` given `[DecidableEq M]` and `[DecidableEq F]` (the branch conditions are equalities in `M` and in `Fin m → F`), and the census pins it with `assert_computable`. The `Classical.choice` it still reaches sits in erased positions, which the census records with the `+choice` flag.

_drafted with the help of Claude Code._

Zcash/Security/RedDSA/KnowledgeError.lean
daira: Adopted wholesale — thanks, this was the right repair. The κ layer is rebuilt on `LabeledOracleComp` with a RedDSA-specific label and output type: `QueryRep` (full-basis coefficient vectors for `R` and `bvk`) rides each challenge query as a label the oracle never sees, and the output announces the returned signature's representation (`KappaOutput`). The representation in effect at the output's query point is the run's first annotation there, so the first-representation table discipline is structural. When the run never queried that point —an adversary can win that way, since verification just reads the table— the announced representation stands in as the fallback. That is sound because such a run cannot depend on the oracle's value at the unqueried point: reprogramming it changes nothing about the output (`run_update_of_findLabel_eq_none`), so the announced representation is fixed independently of the answer there and its one bad challenge is still a `1/|F|` target. The bad-challenge arm is then bounded by the existing `firstLabelOrFallbackBad`/`finalBadWithoutRelation_measure_le` squeeze at `(qH + 1)/|F|` —as in FPS's accounting— with the game's own final challenge query played by the fallback branch. That is why the separate completion step and the guessing bounds (`Common/Guessing.lean`, `RedDSA/ExtractionBound.lean`) are gone entirely.

On your second point: the ledger arm's algebraic hypotheses now say exactly that. `halgLabel` and `halgOut` require query-time labels and announced representations to evaluate to the transactions' actual elements over the presented basis, so representations are enforced to be constructed from the known public basis points at query time rather than merely existing.

_drafted with the help of Claude Code._

daira (commented):
Self-ACK

TalDerei (approved):
ACK properly uses breaks-as-computed-data now; honest-party side for forgeability is follow-up work.

PR #160, opened 2026-07-31
Draft: Action: link with Rocq

clarus:
This is a Work In Progress.

I need to check that the statement of equivalence is correct, and that the translation from Lean to Rocq is reliable.

The general technique is to have a definition for the Action statement in "simple Lean" holding into a single file without dependencies. This definition is shown in Lean as equivalent to the Ironwood Action statement. We also translate this simple sub-set of Lean in Rocq, and show the equivalence to the Garden Action statement in Rocq, in the companion pull request https://github.com/formal-land/garden/pull/90


book/src/formal-verification/ironwood-garden-action-bridge.md
```diff
@@ -0,0 +1,138 @@
...
+   the 64 and 1,024-row constant tables;
+5. stamps both generated files with the exact Lean SHA-256;
+6. checks the committed complete source diff.
+
+A body-only source change therefore changes the corresponding generated Rocq
+body or is rejected before any output is written; updating only the source
+hash cannot make a stale translation pass.
+
+The generated constants use Rocq primitive arrays. Their signed-`Z` accessor
+preserves Lean's behavior: negative indices select row zero and indices past
```
daira, line 96: I didn't know that the Lean code had this behaviour. It seems wrong: we should prove that negative-index accesses do not occur. Otherwise there's the risk of a soundness error not being caught.

TalDerei: cc @mitschabaude

Issue #110, opened 2026-07-26
Reduce reliance on `native_decide`


TalDerei:
Reduce compiler trust by replacing `native_decide` with ordinary `decide` where practical and structural proofs where necessary. Some candidates across 86 `native_decide` calls:

<img width="1426" height="332" alt="Image" src="https://github.com/user-attachments/assets/0679c95a-c9a4-412e-8f89-6dcb7fca3c3b" />

TalDerei:
Supplemental pre-publication audit context (origin/main efd4e08):

The top adaptive Action knowledge-soundness theorem reaches 50 generated native_decide axioms owned by 35 declarations. All owners are enumerated exactly by +native pins, and the cone contains no sorry, admit, or handwritten project axiom. The remaining risk is therefore not silent axiom drift but the accepted native TCB: Lean compiler/code generation, runtime and bignums, plus Task externs used by parallel key certification and the proved csimp fast-MSM replacement.

The highest-value structural replacements are the Action copy/fixed/gate/permutation coherence checks, numeric budget/degree checks, keygen certificate, and curve/fixed-base certificates. Until replaced, publication should describe these facts as compiler-checked rather than kernel-checked.

TalDerei:
_native owners for the Action capstones_

**Circuit copy, layout, and coherence**
• Zcash.Snark.actionCopyBounds
• Zcash.Snark.actionCopyActiveRowFailures_eq_nil
• Zcash.Snark.actionCopyAddressFailures_eq_nil
• Zcash.Snark.actionMissingConstantAllocations_eq_nil
• Zcash.Snark.actionConstantSites_fit
• Zcash.Snark.actionConstantCellAddressFailures_eq_nil
• Zcash.Snark.ActionFixedCoherence.queryCoverageFailures_eq_nil
• Zcash.Snark.ActionFixedCoherence.realizationFailures_eq_nil
• Zcash.Snark.ActionGateCoherence.adviceQueryColumnsAllocated
• Zcash.Snark.ActionGateCoherence.domainExponent_lt
• Zcash.Snark.ActionGateCoherence.gateData_eq
• Zcash.Snark.ActionGateCoherence.selectorDegree
• Zcash.Snark.ActionPermutationDomain.instanceQueryLayout_columns_lt • Zcash.Snark.ActionPermutationDomain.numInstanceColumns_eq
• Zcash.Snark.ActionPermutationDomain.permutationColumnCount_eq
[...]

Issue #66, opened 2026-07-19
Instantiate the Fiat–Shamir oracle with Halo2's concrete Blake2b byte encoding


TalDerei:
see context below; @daira not sure we want to do this. i've already accepted something like this (ie. not modeling Blake2b) as a floor in our code. The fixture also includes the transcript ordering already. Specifically, we don't just capture transcript ordering,  but it's modeled as executable Lean and checked against the capture. What we don't model is: byte encoding, domain-separator bytes, the hash itself.

--------------------

`deriveChallenges` (`Zcash/Snark/Verifier/FiatShamir.lean`) already models Halo2's deployed absorb/squeeze order (advice → θ; lookup-permuted → β, γ; … ; per IPA round (Lⱼ, Rⱼ) → uⱼ), and the fixtures check the captured challenges against it. So the ordered `transcript → challenges` map is **already** verified — what is *not* modeled is only the byte layer beneath the schedule.

### Already checked (single- and multi-action)
- `deriveChallenges_matches_captured_schedule` — the captured `ch` is exactly what the ordered schedule squeezes from `capturedInit ++ proof`.
- `nonInteractiveFingerprint_matches` — the fingerprint run through that schedule matches the captured MSM.
- Ordering fidelity: `capturedScheduleIncludesInit_eq_true`, `capturedInit_startsWith_vkTranscriptRepr`, `capturedInit_has_vk_scalar_and_instance_commitments` (VK representation, then instance commitments, in order).
- Binding sensitivity: `swapped_advice_absorb_breaks_schedule`, `swapped_lookup_permuted_breaks_schedule` — permuting the proof's commitments changes the derived challenges (`deriveChallenges … ≠ ch`).

### Remaining gap
`FiatShamir.squeeze : List (TranscriptElt F G) → F` is treated as an abstract random oracle; `capturedFs` instantiates it as a table returning the captured challenge at each captured transcript prefix. The unmodeled step is the concrete byte encoding beneath the typed schedule:

```text
transcript elements → Halo2 byte encoding (point/scalar serialization + domain-separator bytes) → Blake2b → challenge
```

The domain-separator bytes include Halo2's transcript-tag encodings (`challenge = 0`, `point = 1`, `scalar = 2`).

A strengthening would replace the abstract `squeeze` with Halo2's concrete Blake2b encoding and re-check the derived challenges against both captured fixtures.

### Why this may not be worth pursuing
[...]

PR #162, opened 2026-07-31
Strengthen evidence of deployed and model verifier equivalence

ebfull:
See also https://github.com/zcash/halo2/pull/924 and https://github.com/zcash/orchard/pull/541.

This PR doesn't find any gaps in the analysis or in the equivalence between the deployed and model verifiers, at least for the purposes of establishing the guarantee we're seeking: **if the deployed verifier has a soundness bug that is ever exploited, the exploiting proofs can be identified retroactively from chain data.** But, there are things we can do to strengthen the evidence and reduce the risk of such a deviation in the first place.

After these changes, the Rust-to-Lean comparison has several distinct kinds of evidence:

* Valid captures show that the Lean model accepts concrete proofs produced by the deployed protocol.
* Sensitivity sweeps show that a tamper to each named proof field or challenge changes the fingerprint at the captured fixtures.
* Random match-only captures exercise distinctions that valid proofs cannot expose.
* The derived-key boundary statements ensure that the comparison uses the same verifying key as the soundness theorems.
* The rational-function development proves that a competing coefficient family differing from Lean's agrees with the assembled MSM at the sampled point with probability at most ε ≈ 2⁻²⁴⁰ — per family, at the captured key and group-valued inputs, at shapes `numProofs ∈ {1, 2, 3}`. The `Perm`→positional bridge and the challenge-coordinate layer are theorems; the full product-space reading of ε rests on the stated seeded-expansion idealization, and the quantifier order makes this a pre-registered comparison rather than a bound on future edits.
* The trust documentation enumerates what still depends on audited source review, stated idealizations, or future byte-level modeling.

This provides a stated, checkable basis for treating the Lean model as the reference verifier used by a historical audit.


book/src/formal-verification/trust-boundary.md
```diff
@@ -0,0 +1,418 @@
...
+> Lean model does not — then some fingerprint-match theorem must fail, **except with
+> probability ≤ ε over the randomness of the captured inputs, conditional on the enumerated
+> premises below**. Both ε and the premises are themselves enumerated, never implicit.
+
+Three scoping rules give the invariant its precise shape:
+
+- **One-way correctness.** Only *deployed-accepts ⊆ model-accepts* is soundness-relevant. The
+  model's accept set is `assemble? = some m ∧ m.eval urs = 0` — rejection is `none`, never a
+  zero MSM. A transcription error that makes the deployed verifier reject proofs the model
+  accepts is out of scope; completeness is not a goal.
```
daira: Completeness *is* a goal, otherwise I'd be wasting my time formalizing Spendability, which depends on completeness. I would say that Spendability is necessary to make Balance integrity even meaningful, because *both* are necessary in order for you to able to spend your funds without their value going to zero.

Also, completeness is necessary to keep the formalization on-track. As you pointed out in the call today, an implementation that doesn't have to maintain completeness could achieve soundness trivially, just by rejecting every proof.

We can't fully prove completeness of the Rust prover without proving correctness of the Rust witness generator. But we should still prove completeness of the Lean model, which then implies witness existence. Given witness existence, we know that if the Rust witness generator has bugs, they can be fixed without needing to change the circuit. So we prove Spendability to the extent that "if there is a bug that prevents you accessing your funds, we can fix it". That is an important and credible level of assurance given that there are all kinds of circuit-independent wallet bugs that can also cause temporary inaccessibility of funds.

book/src/formal-verification/trust-boundary.md
```diff
@@ -0,0 +1,418 @@
...
+> can be identified retroactively from chain data.
+
+The identification procedure is an audit: for every committed bundle, decode the proof bytes,
+re-derive the challenges with the schedule model, and evaluate `DeployedAccepts` at the
+Lean-derived verifying key against the committed public inputs. Every bundle receives either
+a **certificate** (the model accepts) or a **flag** (the model rejects, or the bytes fail to
+decode). Today this is an *algebraic adjudication predicate plus the premises on this page*,
+not an operational audit — byte-level Blake2b, proof-byte decode, public-input derivation,
+and a chain-sweep harness are exactly the enumerated premises — but the predicate and its
+trusted remainder are what this repository can pin down.
```
daira: I absolutely intend to pin down byte-level issues in this repository.

book/src/formal-verification.md
daira: We're going to remove the native-executing checks entirely. The performance gain is not worth the extension of the trusted base. If I'd understood how the native execution worked, I'd never have ACKed them in the first place. They should be documented as temporary.

TalDerei: tracked by https://github.com/zcash/ironwood/issues/110.

TalDerei (commented):
taking over this PR; after upstream lands, we'll need to: pin the released Orchard version and regenerate fixtures, rebase this PR over #168's APIs.

scripts/regenerate-fingerprint-fixtures.sh
```diff
@@ -36,9 +36,9 @@
 set -euo pipefail
 cd "$(dirname "$0")/.."
 
-# The `fingerprint-random-capture` branch heads (the capture tooling), each exactly one
-# commit atop the release pin used by CI's honest regeneration (fixtures.yml).
-HALO2_COMMIT=7ef444fe1e5740838bf7aaaa5858e54e5edf3e64   # ebfull/halo2 fingerprint-random-capture
+# The capture-tooling branch heads and release commits used by CI's honest regeneration
+# (fixtures.yml).
+HALO2_COMMIT=091e54786009f23c49a0531cbb980b8d05f7099f   # zcash/halo2#924 head
```
TalDerei: temporary pin.

book/src/formal-verification/trust-boundary.md
TalDerei, line 1: inclined to tighten up this book entry.

scripts/regenerate-fingerprint-fixtures.sh
TalDerei, line 1: didn't seem necessary to include this kind of script, because on CI failure (indicating fixture divergence) you could just ask the AI to regenerate the fixture from the upstream orchard tag we've pinned in our fixture yml file. agnostic on keeping this around.

Zcash/Snark/Fixtures/MultiAction/Random/Epsilon.lean
TalDerei, line 1: ACK, this shows that each final verifier coefficient is a rational function of the proof’s scalar data and challenges.

book/src/formal-verification/trust-boundary.md
TalDerei: tracked by https://github.com/zcash/ironwood/issues/66.

book/src/formal-verification/trust-boundary.md
TalDerei: changed the language to reflect that.

book/src/formal-verification/trust-boundary.md
TalDerei: ended up deleting entirely, we can iterate on a polished version.

TalDerei (approved):
pushed a bunch of updates and organized things, Self-ACK.

book/src/formal-verification/glossary.md
```diff
@@ -82,7 +83,7 @@ development. Anchors point to a module + name under `Zcash/`.
...
+<div class="g"><div class="g-head"><span class="term">checked trust boundary</span><span class="anchor">Zcash.TrustBoundary</span> (and <code>Zcash.Snark.Fixtures.*.TrustBoundary</code> modules)</div><div class="def">Build-time pins on what a theorem may rest on: <code>assert_axioms</code> asserts a bound on the axioms used by a definition, so that a stray <code>sorry</code> or a new axiom fails the build instead of silently widening the trusted base. <code>assert_computable</code> additionally asserts that the definition is a plain <code>def</code>, ensuring constructivity of security reductions. Some of the <code>TrustBoundary</code> modules also use <code>#guard_msgs</code>-pinned <code>#print axioms</code> checks, e.g. to pin specific native axioms. See <a href="../formal-verification.html#trust-discipline">Trust discipline</a>; what the fixture boundaries <em>check</em> is each family's <code>Boundary.lean</code> statement of record.</div></div>
```
daira, line 86: > Some of the <code>TrustBoundary</code> modules also use <code>#guard_msgs</code>-pinned <code>#print axioms</code> checks, e.g. to pin specific native axioms.

All of these can and should be replaced with `assert_axioms` or `assert_computable`.

book/src/formal-verification.md
```diff
@@ -126,17 +130,25 @@ modules and are enforced by `FixtureCheck`. The obligations use two commands fro
...
-The boundaries kept as literal pins are the two fixture censuses,
-`Zcash.Snark.Fixtures.SingleAction.TrustBoundary` and `…MultiAction.TrustBoundary`, which belong to
+The boundaries kept as literal pins are the four fixture censuses —
+`Zcash.Snark.Fixtures.SingleAction.Honest.TrustBoundary`, `…MultiAction.Honest.TrustBoundary`,
+and their two `…Random.TrustBoundary` siblings — which belong to
 the `FixtureCheck` target (kept out of `lake build Zcash` because the captures are large and slow).
 Each states its tier with `assert_axioms` like the rest of the development, and *additionally*
-retains a `#guard_msgs`-pinned `#print axioms fingerprint_matches` documenting precisely *which*
+retains `#guard_msgs`-pinned `#print axioms` checks on `fingerprint_matches` and the derived
+boundary theorems, documenting precisely *which*
```
daira, line 139: This implies that `#print_axioms`/`#guard_msgs` are doing a stricter check than `assert_axioms ... +native(...)`, but they aren't. The latter is actually the stricter check (it does additional verification of where the native axioms are declared).

Zcash/Circuits/Fixtures/PROVENANCE.md
```diff
@@ -89,3 +89,46 @@ event plus the `CircuitCheck` equality checks (the projected Lean CS must equal
...
+- Publish the instrumented halo2/orchard branches (or upstream the dump tooling
+  behind a feature flag), record the exact commits here, and add a CI
+  job that regenerates the JSON fixtures from those pins and diffs them against the
+  checked-in bytes — the circuit-side analogue of what `fixtures.yml` already does
+  for the `Zcash/Snark/Fixtures/` verifier-fingerprint captures (see
+  `Zcash/Snark/Fixtures/PROVENANCE.md`).
+- For the base (pre-NU 6.3) circuit, a stronger check is available: verify the
+  dumped constraint system against the actual mainnet Orchard verifying key — a
+  fixed public artifact — tying the chain to the deployed network rather than to
+  any checkout.
```
daira, line 134: Why is this restricted to the pre-NU6.3 circuit? It's equally true of the post-NU6.3 circuit.

daira (approved):
There's a lot of code here; I can't review it all. It seems ok from spot-checks ¯\\\_(ツ)\_/¯

PR #163, opened 2026-07-31
chore(fv): census pin hardening

TalDerei:
The way we enforce trust census pins with regex is fragile, because a new theorem that falls outside our semi-standard naming convention (like these top-level capstones) doesn't get pinned. This widens the regex, but i'll separately do a pass at standardizing our naming conventions and hardening the regex.


TalDerei:
expanded scope to fix some other scattered annoyances.

daira (approved):
utACK

PR #164, opened 2026-07-31
chore(fv): harden multi-action fixture permutation check

TalDerei:
This was missing, but it would've been caught by the single-action fixture, so it's not really a concern and may actually be unnecessary; this only hardens the check on the multi-action fixture side.

Closes a gap in the multi-action fixture's permutation checks, which accepted uneven chunk widths and duplicated common-eval indices, by adding width / index coverage conditions and pinning `chunkLen` and the chunking to `actionCircuit`.


daira (approved):
utACK

Issue #119, opened 2026-07-27
Make the proof-journey validator check that symbols exist in the tree


TalDerei:
`book/validate-proof-journey.py` checks the proof journey's graph shape — it reports `44 nodes, 49 edges, 10 stages, 36 capstones` and passes. What it does not check is whether the `symbol` entries in `book/src/formal-verification/proof-journey-data.json` still exist in the Lean tree, so deleting a theorem leaves a stale reference that CI accepts silently.

Seven symbols are currently stale, all fallout from #106/#112 removing the rewind stack:

- `relation_of_foldGens`
- `orchardDeployedAlgebraicForkingFixedSlot`
- `fixedSlotExtractOrMiss`
- `orchard_verifier_vesta_forking_constraint_adaptive_rewind_xgood`
- `honestCompletenessSupply_or_relation`
- `orchard_verifier_vesta_member_constraint_budgeted_hfold_derived`
- `snarkExtractionFailureEventDeployed_measure_le`

#96 removed the one entry it invalidated itself (`orchard_verifier_sound_vesta_computed`) and rewrote the glossary's "capstones" entry, which described the `legacy_orchard_verifier_vesta_*` ladder that PR retired. The seven above predate it.

## Scope

- Extend the validator to resolve each `symbol` against the tree — ideally against its declared `path`, so a symbol that moved files is caught too.
- Clear the seven stale entries: repoint each to the endpoint that replaced it, or drop it if the stage no longer exists.

TalDerei:
closing as stale on current main.

Issue #51, opened 2026-07-11
Deduplicate capstones


TalDerei:
Stacking PRs left duplicated and legacy capstones behind: earlier forms of a result that a later PR superseded but did not remove, so the same fact is proven at several endpoints and it is not obvious which one is current.

**The bar is not exactly one top-level capstone.** A handful is expected and correct — the route legitimately ends in several endpoints that differ in their hypothesis sets (discrete-log flavour against adversary model), and a caller picks the one whose assumptions it can supply. What this issue targets is *redundancy*: two endpoints proving the same thing, or a legacy form kept alive only because a stacked branch never cleaned it up.

**Scope**
- Remove capstones that a later result strictly supersedes — same conclusion under weaker or equal hypotheses.
- Remove legacy ladders left behind by superseded routes.
- Keep endpoints that genuinely differ in what they assume, and say so where they live, so the plurality reads as intentional rather than as residue.

A future single overall capstone could sit above this layer and compose Fiat–Shamir, DLR hardness, multiopen decoding, and circuit semantics into an end-to-end security theorem. But it should be a computational/game-based reduction, not merely `soundness ∨ ∃` relation, which is vacuous for these groups.

TalDerei:
Follow-up from #106: as part of deduplicating the capstones, remove the remaining **bare-existential relation endpoints in the legacy rewinding stack**.

The rewind-free path introduced by #106 charges an explicit, computable relation finder to DLOG, so the legacy endpoints are not on that active path. However, older compatibility capstones still branch directly on `HasNontrivialRelation`, notably:

- `orchard_verifier_vesta_member_constraint_derived`
- `deployed_value_check_node_binding`
- `deployed_member_node_binding`

These should either be deleted with the superseded rewinding capstones or rewritten so the relation branch carries explicit `NontrivialRelation` / `AugmentedRelationWitness` data produced by a computable finder.

Completion criterion: no security capstone treats the bare existence of a relation as a DLOG-break branch. Any remaining compatibility-only surfaces in `GoodX`, `Bridge`, `Quotient`, or `Multiopen/Compat` should be removed or clearly isolated from the final capstone.

PR #165, opened 2026-07-31
chore(fv): reorganize SNARK modules

### LINKED ISSUE #170: Remove stale hencodes language
after https://github.com/zcash/ironwood/pull/99, there's still stale hencode references.

TalDerei:
Organizes the SNARK tree by dependency layer, cuts the advertised endpoint surface to what the development actually claims, and tightens the doc comments around both.

Rebased onto current main, which had independently reorganized Fixtures/ by capture kind. That split is orthogonal and is preserved: Fixtures/ keeps main's Honest/ and Random/ structure, and only the Action circuit's own endpoints move out of it.

## Layout

- **`Soundness/`** held 30 loose modules beside its subdirectories. They now factor into five dependency layers: `Ipa/` and `Constraint/` are independent roots, `Argument/` proves the permutation and lookup arguments over `Constraint/`, then `Pricing/` and `Relation/`. One loose module is left — the acceptance predicate `Main.lean`.
- **`Capstones/`** holds the Action circuit's advertised statements, previously inside `Fixtures/`: `ActionEvents` → `ActionChecks` → `ActionBudgets` → `Action`. Endpoints about the verifier's algebra rather than the circuit statement stay with the layer that proves them.
- `Verifier/Key`, `Verifier/ConstraintSystem`, `Circuits/Integration/VerifierCS`, `Keygen/Commitments`, `Soundness/Circuit/Terminal`, and `Soundness/StraightLine/{Terminal,Event}` split out of oversized modules.

## Endpoints

`Capstones/Action.lean` advertised twelve endpoints across three "settings", of which only one was an adversary model. It now has four, in 563 lines rather than 2685.

- `captured` took the four surface bounds as hypotheses and did the addition — pinned as a top-level leaf while `sequential` called it.
- `sequential` bounded a prover that runs in order. That was never the security claim, and `adaptive` proves its own surfaces and never went through it.
- The one-action pair was the degenerate case of the bundle endpoints, since `actionProofParams = actionProofParamsFor 1`.

What remains is a 2×2 — soundness and knowledge soundness, each as the compositional formula and its resource-accounted evaluation, both at arbitrary bundle size and both landing on `Adv_DLOG(2^127, 2^127) + 2^-83`.

## Names

Action endpoints read `<setting>_<property>_<form>`, ending in `_error_bound` or `_finite_security`, replacing spellings that encoded the event and work factor in the identifier; `_prob_le` and `_measure_le` become `_probability_bound`. The census matched by protocol prefix alone, so a standardized name outside those prefixes could go unpinned in silence — it now also accepts those suffixes. Endpoints get their own `Capstone` namespace, so the deliverables are no longer pinned as `Zcash.Snark.Fixture.orchard_*`. The capture namespaces are left alone: halo2's `dump_vesta_lean_fixture` writes them into each generated `Fixture.lean`, which the fixture CI regenerates and diffs.

## Not pure code motion

- **Eight endpoints deleted**, as above. Scope reduction rather than layout, and the part most worth arguing about.
- **Fourteen supporting lemmas lose `private`** — each is used from a later module in the split chain, and Lean has no narrower visibility. None is censused or endpoint-named.
- **Four modules deleted**: three compatibility shims the first commit added that nothing ever imported, and the `Soundness/FiatShamir.lean` umbrella, whose five modules `Zcash/Snark.lean` now lists directly.
- **Doc comments**: issue numbers and development-history narration removed across 24 files; the worst module docstrings cut hard, ones already at the standard left alone. Fixes #170, and #119's substring hole in `validate-proof-journey.py`.

## Verification

`lake build --wfail` over all default targets, `check_endpoint_census.sh` (32 endpoints, all pinned), `check_csimp_census.sh`, `check_no_umbrella_imports.sh`, `validate-proof-journey.py`, and `typos`. The four generated `Fixture.lean` captures are byte-identical to `main`. Doc-only commits additionally check that stripping comments leaves the code byte-identical, and a pin audit resolved all 1271 `assert_axioms`/`assert_computable` entries against their declarations.


Zcash/Snark/Fixtures/MultiAction/Honest/ActionCapstone.lean
```diff
@@ -1417,8 +1417,9 @@ theorem action_dlog_groupWork_le_2pow126
 
 /-- **The exact captured Action soundness bound.**  Its left-hand event is literal deployed
 acceptance with a false `BundleStatement`.  The combined profile prices IPA, unbatching, quotient,
-and Action-terminal relation branches once. -/
-theorem orchard_action_acceptFalseStatement_prob_le_captured
+and Action-terminal relation branches once.  This is the compositional error formula used by the
+finite-security capstone. -/
+theorem orchard_action_captured_soundness_error_bound
```
daira: Soundness or knowledge soundness? (Same comment for all "soundness" names.)

TalDerei: that's just soundness, and there are separate ones for knowledge soundness. there are 4 top-level capstones: two soundness and two knowledge soundness.

daira: Why do we even need soundness? Knowledge soundness obviously and tightly implies it.

Zcash/Snark/Fixtures/MultiAction/Honest/ActionCapstone.lean
```diff
@@ -2017,8 +2022,9 @@ theorem orchard_action_knowledgeFailure_prob_le_adaptive_for
 
 /-- **Concrete bare-adaptive Action capstone.**  At `Q <= 2^123`, the complete adaptive finder
 fits a conservative `2^127` random-oracle/group-work envelope (eight uncached represented runs),
```
daira: The DL security is at most $2^{126}$ group operations. As a success probability bound for Pollard rho, it's $ε ≤ \frac{m \cdot n \cdot (n-1)}{2 \cdot |G|}$ for $n$ group operations over a group $G$ with automorphism group of size $m$ (for Pasta, $m = 6$).

Zcash/Snark/Fixtures/MultiAction/Honest/ActionCapstone.lean
TalDerei: it was just a simpler capstone to reason about as i built up to full knowledge soundness, this was iterative.

scripts/check_endpoint_census.sh
```diff
@@ -47,7 +52,7 @@ cd "$(dirname "$0")/.."
 # `orchard_verifier_*` currently matches nothing: those rungs were retired with the legacy rewind
 # paths. It is retained as a guard, so a reintroduced name in that family is demanded rather than
 # silently unpinned.
-ENDPOINT_RE="^(orchard_verifier_|orchard_action_|orchard_deployed_|competing_|nonInteractiveFingerprint_matches_derived)|bundleStatement_or_relation|workFactor|fingerprint_matches_positional"
+ENDPOINT_RE='(^orchard_(verifier|action|deployed)_)|(^competing_)|(^nonInteractiveFingerprint_matches_derived)|(bundleStatement_or_relation)|(workFactor)|(fingerprint_matches_positional)|(_(error_bound|finite_security|capstone)$)'
```
daira: This should include `*_measure_le` (the top-level probabilistic capstones).
```suggestion
ENDPOINT_RE='(^orchard_(verifier|action|deployed)_)|(^competing_)|(^nonInteractiveFingerprint_matches_derived)|(bundleStatement_or_relation)|(workFactor)|(fingerprint_matches_positional)|(_(error_bound|finite_security|measure_le|capstone)$)'
```
(Or I can rename those, but this does the right thing until I have.)

Btw shouldn't this also include `_probability_bound`?

Zcash/TrustBoundary.lean
```diff
@@ -94,8 +94,10 @@ Both commands are built on `Lean.collectAxioms`, which walks a declaration's tra
 depends on is checked by nothing, however prominent it is. Deliverable endpoints are exactly the
 top-level leaves, so they must be pinned *directly* — never left to inherit coverage from some
 dependent, which would vanish the moment that dependent is refactored.
-`scripts/check_endpoint_census.sh` enforces this in CI for the capstone naming families, and
-lists them; extend its patterns when a new endpoint family appears.
+`scripts/check_endpoint_census.sh` enforces this in CI for the capstone naming families. A new
+public endpoint must either belong to one of the listed protocol families or end in the semantic
+suffix `_error_bound`, `_finite_security`, or `_capstone`; the latter keeps new protocol families
```
daira:
```suggestion
suffix `_error_bound`, `_finite_security`, `_measure_le`, or `_capstone`; the latter keeps new protocol families

File: scripts/check_endpoint_census.sh (hunk 3 above)
Comment (TalDerei) [reply scripts/check_endpoint_census.sh]: it should, renaming.

File: Zcash/Snark/Verifier/Assemble.lean [hunk 5]
```diff
@@ -1894,12 +1866,7 @@ private theorem filterMap_getD_idxOf_of_forall_isSome {α β : Type*} [Decidable
         simpa using ih (fun x hx => hsome x (List.mem_cons_of_mem _ hx)) hjt hgj
 
 open Classical in
-/-- **General slot routing.** Every query is routed by the grouping to a member of one point set:
-the member's recorded identity is the query's slot, the query's point appears among the set's
-points, and the member's recorded evaluation at that point's position is the query's claimed
-evaluation — under `hdet`, one claimed evaluation per slot and point. The many-point companion of
-`constructIntermediateSets_unique_comm_routed`, and the eval-faithfulness fact the constraint
-feed bindings read. -/
+/-- Every query is routed while preserving its commitment, point, and claimed evaluation. -/
```
daira: Blank line between doc comment and theorem/definition. Check for other cases.

Zcash/Snark/Fixtures/MultiAction/Honest/ActionCapstone.lean
TalDerei: yeah sorry, i know the pollard rho bound here, but for some reason my doc skill was modifying this incorrectly.

Zcash/Snark/Fixtures/MultiAction/Honest/ActionCapstone.lean
TalDerei: will keep them for now (because removing them would involve other more involved cleanups and splitting things into corollaries) and we can reduce to just two knowledge soundness capstones separately: one definition, and then one concrete security argument.

Zcash/Security/KeyBinding/Probability.lean
```diff
@@ -60,7 +60,7 @@ an `OracleComp` from the Fiat–Shamir layer):
   tables drawn independently, the tables uniformly; same `(Q+4)·(Q+3)/|RIVK|` bound.
 
 The remaining modelling steps toward the full ZIP 2005 key-binding theorem are tracked in
-issue #68.
+a separately tracked issue.
```
daira: This is closed, but referenced as though it isn't. I think this sentence can just be removed.

Zcash/Snark/Fixtures/MultiAction/Honest/ActionCapstone.lean
TalDerei: tracking in https://github.com/zcash/ironwood/issues/174.

mitschabaude:
I had my Fable analyze the folder structure of `Soundness/`, I think this is good feedback.

TL;DR, the folder structure is not fully logical yet, ideally folders don't have mutual dependencies and split cleanly along responsibilities.
Right now, some folders like `Canonical/` are in a weird place. Some of the feedback below seems easy to fix and should be done

----

The five-layer story this PR tells for `Soundness/` — `Ipa/` and `Constraint/` as independent roots, `Argument/` over `Constraint/`, then `Pricing/` and `Relation/` — is the right taxonomy, and I want to keep it. But I checked it against the actual import graph at `2c32444a`, and it currently holds only file-locally, not at directory granularity. I think a handful of cheap changes, mostly within this PR's stated scope, make it true.

**The directory-level import matrix** (each row lists every `Soundness/` neighbor the directory imports from):

| Directory       | Imports from                                                                  | vs. the five-layer story                                                |
| --------------- | ----------------------------------------------------------------------------- | ----------------------------------------------------------------------- |
| `Ipa/`          | —                                                                             | ✅ root, as advertised                                                  |
| `Constraint/`   | `Composition`, `Multiopen`                                                    | ❌ "root" with two down-edges into the legacy blob                      |
| `Argument/`     | `Constraint`                                                                  | ✅ as advertised                                                        |
| `Relation/`     | `Argument`, `Constraint`, `Ipa`                                               | ✅ as advertised                                                        |
| `Pricing/`      | `Argument`, `Canonical`, `Constraint`, `FiatShamir`                           | ⚠️ entangled with `Canonical` and `FiatShamir`                          |
| `Canonical/`    | `Argument`, `Constraint`, `Deployed`, `Multiopen`, `Pricing`, `Relation`      | ❌ not mentioned in the narrative; woven both below and above `Pricing` |
| `Multiopen/`    | `Canonical`, `Deployed`, `Ipa`, `Main`, `Pricing`, `Relation`                 | ⚠️ mutual with `Canonical` and `Deployed`                               |
| `Deployed/`     | `Ipa`, `Main`, `Multiopen`                                                    | ⚠️ mutual with `Main` and `Multiopen`                                   |
| `Main.lean`     | `Deployed`                                                                    | ⚠️ loose file, participates in a cycle                                  |
| `FiatShamir/`   | `AGM`, `Deployed`, `Main`, `Pricing`                                          | ⚠️ mutual with `Pricing` and `AGM`                                      |
| `Composition/`  | `AGM`, `Deployed`, `FiatShamir`, `Multiopen`, `Pricing`                       | ⚠️ mutual with `AGM`                                                    |
| `AGM/`          | `Composition`, `Constraint`, `Deployed`, `FiatShamir`, `Multiopen`, `Pricing` | ⚠️ mutual with `Composition` and `FiatShamir`                           |
| `Circuit/`      | `Canonical`, `Multiopen`                                                      | ✅ strictly above                                                       |
| `StraightLine/` | `AGM`, `Circuit`, `Composition`                                               | ✅ strictly above                                                       |
| `Action/`       | `AGM`, `Composition`, `StraightLine`                                          | ✅ strictly above                                                       |

✅ stratifies as the narrative says · ⚠️ inside the legacy SCC, no claim made about it · ❌ contradicts or undercuts the advertised layering

Computing SCCs on this graph: **eleven of the fifteen nodes form a single strongly connected component** — everything except `Ipa/` (true root) and the `Circuit/ → StraightLine/ → Action/` chain. Four of the five new layer directories participate in the blob, and `Canonical/` — which the layer narrative doesn't mention — is woven both below `Pricing/` (its semantics feed `ChallengePricing`) and above it (it imports `GoodChallenge`). File-level is acyclic, of course; the point is that the directories don't yet correspond to the layers the PR names.

The good news is that tracing each cycle-forming edge to its file shows most of them are shallow:

**Suggested for this PR** (all are the kind of move it already does):

1. **Dead imports.** `Canonical/LookupSemantics.lean` and `Canonical/LookupRows.lean` import `Pricing/GoodChallenge` but use none of its declarations; `Constraint/ConstraintCore.lean` imports `Composition/Bridge` and references no declaration from it (it only wants Bridge's transitive imports — import those directly instead). Verified: removing all three imports builds. Three of the worst edges cost three deleted lines, and removing the ConstraintCore edge is half of making `Constraint/` a real root.
2. **`Deployed/` mixes two altitudes.** `Verification`/`Binding`/`Fold` are low-level, but `Vesta.lean` and `VacuityWitness.lean` (moved in here by this PR) are top-of-stack — Vesta is the deployed decoded capstone and imports `Main` + `Multiopen`. Moving those two files up dissolves the `Deployed ↔ Main ↔ Multiopen` mutualities.
3. **`FiatShamir/` mixes two altitudes.** `Oracle.lean` is a leaf (the uniform-challenge PMF) that `Pricing/GoodChallenge` rightly consumes; `Adversary/*` sits high (imports `AGM` and `Pricing/UniformMeasure`). Splitting Oracle from the adversary models removes the `Pricing ↔ FiatShamir` cycle.
4. **`Constraint/VanishingSlotFold.lean` is misfiled.** It consumes `Multiopen/Deployed` (which imports `Main`), so it's mid-stack living in the root directory. Moving it up is the other half of making `Constraint/` a root — after which `Constraint → Argument → Relation` stratifies exactly as the PR description claims.
5. **Split `Canonical/` along the seam the imports reveal.** After (1), its real `Pricing` dependency is a handful of measure-facing statements in `PermutationSemantics` (`uniformChallenge_szBadSet`, `…_union`) and `LookupRows` (the `additiveZeroBadSet` lemmas) that belong up in `Pricing/`; the deterministic semantics stay put. This is statement relocation rather than a file move, but it's small and it's the step that actually resolves the `Canonical ↔ Pricing` cycle — it belongs in a PR whose stated purpose is the layer structure.

**Agreed follow-up, not this PR:**

6. **Re-home the canonical-decode cluster.** `Multiopen/{Decode, Compat, CanonicalRelation, ConstraintResolver}` and `Canonical/{Terminal, InstanceCommitment, PermutationInstantiation, ConstraintModel, …}` are one interleaved unit wearing two directory names — this is the entire `Canonical ↔ Multiopen` cycle. It needs one decision about where the unit lives (own layer above both, or folded together); mechanical after that. `Compat.lean` should get a real name in the process — it's load-bearing, not a shim.

**Explicitly out of scope:** the `AGM ↔ Composition ↔ FiatShamir/Adversary` triangle (~26k lines). Those mutualities reflect the real entanglement of the staged decode with the adversary game and need interface work, not file moves — its own arc.

With 1–4 the SCC drops from eleven directories to two small local tangles; with 5–6 only the AGM triangle remains, and the five-layer story in this PR's description becomes literally true of the tree.

Zcash/Security/Ledger/Capstone.lean
```diff
@@ -36,7 +36,7 @@ probability is then bounded by a named ε hypothesis. The key-binding arm's ε i
 discharged in `KeyBindingArm.lean` against the key-binding layer's probability bound
 (`toInterface_break_measure_le`), for a ledger adversary in the oracle model.
 Connecting that bound to this layer's named-ε slot — an oracle machine against a `PMF`
-over valid annotated ledgers — is still open (#155), as are the Merkle and
+over valid annotated ledgers — is still open, as are the Merkle and
```
daira: Discussed on Signal: retain references to open issues (that should be open and that describe tracked gaps).

Zcash/Snark/Capstones/Action.lean
```diff
@@ -313,131 +178,16 @@ theorem orchard_action_adaptive_bundle_knowledge_error_bound
...
-
 /-- Consensus-generic knowledge capstone: extractor failure is bounded by one profiled Vesta-DLOG
 advantage plus `2^-83`.  This is the concrete resource-accounted finite-security capstone for
-knowledge soundness. -/
-theorem orchard_action_adaptive_bundle_knowledge_soundness_finite_security
+knowledge soundness.
+
+The profile is evaluated at `2^127` queries and group work, which is *above* Vesta's own discrete-log
+security of at most `2^126` group operations.  The advantage is left symbolic, so the statement holds
+regardless; but the resource figure records a work-factor target and cannot be read as a security
```
daira: I still don't understand where the work factor target came from.

("Understanding the Security of Zcash" gave $2^{125}$ as the security target we used in the design, before we had a proof of anything other than pencil-and-paper proofs of a few specific primitives. It's ok —and inevitable— to have a higher design target for the underlying curves and primitives than the achieved security bound for the overall protocol, although the presentation did not make the distinction clear, and $2^{125}$ could have been read as the latter when it meant the former.)

Zcash/Snark/Soundness/AGM/StraightLineFiniteSecurity.lean
```diff
@@ -8,6 +8,14 @@ The probability capstone and the work-factor interpretation are deliberately sep
 module records random-oracle queries, group work, and the actual direct-coordinate field/data work
 as distinct quantities.  It does not assert a generic-group DLOG success formula.  A caller
 supplies the finite-security Vesta DLOG advantage for the resulting concrete solver cost.
+
+The solver budgets below are work-factor targets, not security levels.  A five-bit overhead takes a
+`2^122` adversary budget to `2^127`, which is above Vesta's own discrete-log security of at most
```
daira: https://github.com/zcash/ironwood/pull/165/changes/46e7977e3b4906f26d386a063adcef4045321af1#r3711136531 also applies here.

daira (commented):
Reviewed with comments. The one about not removing open issue references is blocking; the rest are either non-blocking or have already been addressed.

Not currently building successfully.

Zcash/Snark/Soundness/AGM/StraightLineFiniteSecurity.lean
TalDerei: this is touching the AGM straight-line modules, and this was deferred to #167 where this code will be removed.

Zcash/Snark/Capstones/Action.lean
TalDerei: this was accidentally stated as one bit above the specified security level and is being corrected in #167, which removes unnecessary derivation chains and tightens some bounds.

TalDerei:
> I had my Fable analyze the folder structure of `Soundness/`, I think this is good feedback.

this is great feedback, implemented in latest batch of commits.

TalDerei:
what, why is our fixtures CI failing. investigating

Zcash/Snark/Capstones/Action.lean
```diff
@@ -0,0 +1,593 @@
...
+Vesta's own ceiling sits below that.  For `n` group operations over `G` with automorphism group of
+size `m` (`m = 6` for Pasta), Pollard rho succeeds with probability at most
+`m · n · (n − 1) / (2 · |G|)`; at `|G| ≥ 2^254` that puts `n` a little above `2^126`.  Against a
+four-bit overhead the ceiling-respecting adversary budget is therefore `2^126 / 2^4 = 2^122`, one
+bit below the `2^123` stated here — so the advantage is being requested slightly above what the
+group can offer.
+
+The advantage is left symbolic, so the statement holds regardless.  The figure records the
+accounting scale the reduction was evaluated at, which is a design target for the underlying curve
+rather than a proven end-to-end bound on the protocol. -/
```
daira:
```suggestion
accounting scale the reduction was evaluated at, not a proven end-to-end bound on the protocol. -/
```
The design target for the "headline cost" (expectation) of the best known pre-quantum DL attack was $2^{125}$, which was met. I only caught this error in Claude's suggestion after you had applied it.

Zcash/Snark/Capstones/Action.lean
```diff
@@ -0,0 +1,593 @@
...
```
daira:
```suggestion
Vesta's own ceiling sits below that at about `2^126` — see the discussion for
`orchard_action_adaptive_bundle_soundness_finite_security` above. Against a four-bit
overhead the ceiling-respecting adversary budget is therefore `2^126 / 2^4 = 2^122`, one
bit below the `2^123` stated here — so the advantage is being requested slightly above what the
group can offer.

The advantage is left symbolic, so the statement holds regardless.  The figure records the
accounting scale the reduction was evaluated at, not a proven end-to-end bound on the protocol. -/
```

TalDerei: https://github.com/zcash/ironwood/issues/175#issuecomment-5182621785 – these are dead paths being pruned in #167 (as a side-effect, no stale 2^127 security references).

TalDerei: for some context, there's stale proof/API surface associated with the old extractor implementation that #167 cleans up. we've already done capstone pruning per https://github.com/zcash/ironwood/issues/51 and in this PR, and now i'm doing this pass at the lower layers those capstones consumed.

daira: There's an incorrect historical claim about the design process of the Pasta curves in the last sentence: that $2^{123}$ was their security design target (for pre-quantum attacks). The design target was actually $2^{125}$. Removing that incorrect information is the bit I care about; the rest of the suggestion is just removing duplication.

TalDerei: so there was an 8x reduction which was a loose bound i didn't think i could get around in the formalism, which is why it's 2^123 (was hitting walls with fable), but i think i found a way to get that down to a 2x reduction to 2^125. so this claim should carry in some form until this is attempted in follow-up work to #167 (though i'll update the doc comment to make that clear here). will fold that work into https://github.com/zcash/ironwood/issues/172.

Zcash/Snark/Soundness/AGM/StraightLineFiniteSecurity.lean
```diff
@@ -8,6 +8,28 @@ The probability capstone and the work-factor interpretation are deliberately sep
...
+
+The five-bit profile is stated at `2^122` rather than `2^121`, so its solver cost is `2^127` — one
+bit above the ceiling, meaning an advantage supplied at that scale is being supplied above what the
+group can offer.  `AGM.StraightLineOrchardConsensusBounds` works that one-bit loss through at the
+consensus target.
+
+None of these figures is an achieved end-to-end security level.  The advantage is left symbolic, so
+the statements hold whatever it is; the number records only the accounting scale the reduction was
+evaluated at, which is a design target for the underlying curve rather than a proven bound on the
+protocol.
```
daira:
```suggestion
evaluated at, not a proven bound on the protocol.
```
Same issue as a previous comment.

TalDerei: resolved in https://github.com/zcash/ironwood/pull/165/changes/899a9fdf3ddaae2d73a9266fb570e778ccafebf4.

Zcash/Snark/Capstones/Action.lean
TalDerei: resolved in https://github.com/zcash/ironwood/pull/165/changes/899a9fdf3ddaae2d73a9266fb570e778ccafebf4. note, this basically will get deleted in #167.

daira (approved):
ACK

PR #167, opened 2026-08-01
feat(fv): strengthen adaptive soundness notion

### LINKED ISSUE #175: Finite-security action endpoints overclaim the adversary budget by one bit
The resource-accounted finite-security action endpoints state an adversary budget one bit above what the reduction certifies against Vesta's discrete-log ceiling.

## The overclaim

`orchard_action_adaptive_bundle_knowledge_soundness_finite_security` (and its `soundness` sibling) in `Zcash/Snark/Capstones/Action.lean` is stated at adversary budget `T = 2^123`. The endpoint's reduction has a four-bit resource overhead (`solverCost_le` gives `16 · T`), so the DLOG solver cost is `16 · 2^123 = 2^127`.

Vesta's discrete-log ceiling is `≈ 2^126` group operations. For a group `G` with an efficiently computable automorphism group of order `m` (`m = 6` for Pasta, the full automorphism group of a `j = 0` curve), the equivalence-class Pollard rho runs over the `|G|/m` classes: its expected cost is `√(π · |G| / (2m)) ≈ 2^126`, and after `n` operations it finds a collision with probability at most `m · n · (n − 1) / (2 · |G|)`.

So the endpoint requests DL hardness at `2^127`, which the curve misses by one bit. The ceiling-respecting adversary budget for a four-bit overhead is `2^126 / 2^4 = 2^122`, one bit below the `2^123` stated. `AGM.StraightLineOrchardConsensusBounds` currently carries the one-bit loss through at the consensus target rather than removing it.

## Overhead inconsistency

Three different overheads appear across the two files:

- the inner straight-line reduction is three-bit (`straightLineDlogGroupWork_le_eight_mul`, `8 · T`), which would land `2^123` exactly at the ceiling;
- the action-bundle endpoint wraps it and is four-bit (`16 · T`);
- the `StraightLineFiniteSecurity` module doc separately narrates a five-bit envelope stated at `2^122`, also one bit above its own `2^121` ceiling-respecting budget.

## Fix options

Either:

- state the endpoint at `2^122` (four-bit → solver cost `2^126`, at the ceiling), accepting one fewer bit of adversary coverage; or
- tighten the action-bundle wrapper to the three-bit inner reduction, so `2^123` lands at the ceiling.

The stated budget is itself a legitimate bit-security convention — the reduction's overhead fixes it, as a block cipher's key length fixes its nominal security — so the defect is only the one-bit overshoot beyond what the curve provides, not the practice of quoting a bit level.

## References for the DL ceiling

- Birthday bound and expected cost: Handbook of Applied Cryptography, Fact 2.27 (https://cacr.uwaterloo.ca/hac/about/chap2.pdf).
- Order-`m` equivalence-class walk, including the `j = 0` CM case: Duursma–Gaudry–Morain, ASIACRYPT 1999 (https://inria.hal.science/inria-00511639v1/document).
- Negation-map component (`m = 2`) achievable near-optimally despite fruitless cycles: Bernstein–Lange–Schwabe, PKC 2011 (https://eprint.iacr.org/2011/003).

Related: #172 (certifying the work-limit premise; its motivation quotes the advantage at `2^126`, while the endpoint code supplies it at `2^127` — the same one-bit discrepancy).

Documented as a known defect in the docstrings of `AGM/StraightLineFiniteSecurity.lean` and `Capstones/Action.lean` (in #165), pending this issue.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code)

TalDerei:
Stacks on the verifier side public-instance work in #168. Also closes https://github.com/zcash/ironwood/issues/175 as a side-effect.

This implements a staged cost model that counts explicit Vesta group operations across the AGM adversary and extractor/reduction. We count every operation explicitly recorded by the staged program: additions, negations, scalar multiplications, and MSM terms across them. We count all explicitly staged group operations, but there may be additional unstaged operations that the current model cannot detect (we assume ordinary Lean helpers and callbacks do not perform unrecorded group operations).

## Summary

PR #167 strengthens the security proof for deployed Orchard Action verification:

- It lets the adversary choose the public statement and proof together after
  random-oracle queries.
- It binds that statement into the transcript through the vk and instance
  commitments, preventing later statement substitution.
- It proves the deployed endpoint for the `2^123` adversary-work profile with
  `Adv_DLOG(2^126, 2^126) + 2^-83`.
  - At this target,
    `orchard_action_knowledgeFailure_adaptiveStatement_2pow123_workFactor_generatorRO_for`
    bounds acceptance without extraction with separate resource/locality
    conditions and a bias-transfer form. This is 123-bit computational coverage
    with an independent 83-bit statistical error at the tail.
- It accounts for oracle access, three decoding passes, and assembly/basis work.
- It removes the obsolete fixed-statement adaptive proof path.

In short, it proves a stronger chosen-statement knowledge-soundness result for
the deployed verification path.

## Why

The existing adaptive capstones proved a weaker game than the verifier actually
plays: the statement was fixed before the random oracle was sampled, so the
initial transcript `init` stayed universally quantified and the vk/instance
absorption verified in #168 was never load-bearing.

PR #167 instead adds a chosen-statement adaptive game under
`Soundness/Action/AdaptiveStatement*`. After querying the oracle, the adversary
outputs the public statement and proof together. The transcript prefix is
derived from the vk representation and the chosen statement's instance
commitments, binding `init` to deployed absorption and preventing statement
replacement after challenge sampling.

More stringent security accounting representation to follow in https://github.com/zcash/ironwood/issues/172.


TalDerei:
rebasing this over #165.

book/src/formal-verification/glossary.md
```diff
@@ -69,14 +69,14 @@ development. Anchors point to a module + name under `Zcash/`.
...
+<div class="g"><div class="g-head"><span class="term">capstones</span><span class="anchor">Capstones/Action.lean</span></div><div class="def">The top-level results a route ends at. The consensus-generic adaptive-statement knowledge capstone lets the adversary output the public statement and proof together, binds the VK and selected instance commitments before <code>theta</code>, uses an executable Type-valued private-witness-or-relation branch, and bounds acceptance when witness extraction fails by profiled DLOG advantage plus the statistical remainder. The profile gives <code>Adv_DLOG(2^126, 2^126) + 2^-83</code> at the <code>2^123</code> computational work target. This is a 123-bit computational work-factor statement with an independent 83-bit statistical soundness error, not an unqualified 123-bit failure-probability claim.</div></div>
```
daira:
```suggestion
<div class="g"><div class="g-head"><span class="term">capstones</span><span class="anchor">Capstones/Action.lean</span></div><div class="def">The adaptive-statement capstone proves knowledge soundness for the deployed Action verifier, for algebraic adversaries (the online Algebraic Group Model), with the challenge derivation modelled as a random oracle. The adversary chooses the public statement and the proof together after making its oracle queries. The statement is bound into the challenge derivation through the vk representation and its instance commitments. For every consensus-valid Action count, the probability that the verifier accepts while executable witness extraction fails is at most <code>Adv_DLOG(2^126, 2^126) + 2^-83</code> at the <code>2^123</code> computational work target. The arguments to <code>Adv_DLOG</code> are upper bounds on the number of random-oracle queries and group operations respectively. This states that within a <code>2^123</code> adversary work envelope, the constant term of the soundness error does not exceed <code>2^-83</code>; it is not an unqualified failure-probability claim.</div></div>
```

Zcash/Snark/Soundness/Action/AdaptiveStatementProfile.lean
```diff
@@ -0,0 +1,380 @@
...
+      (adaptiveActionStatementInstanceCommitment pp basis (family.runOutput basis O).inputs)
+      (family.runProof basis O).proof.1 (family.runRecord basis O) <
+        Zcash.Arithmetic.scalarFieldOrder)
+    (B : VestaG) where
+  proverGroupWork : Nat
+  reductionGroupWork : Nat
+  advantage : Nat → Nat → ENNReal
+  advantage_mono : ∀ {q q' g g'}, q ≤ q' → g ≤ g' →
+    advantage q g ≤ advantage q' g'
+  hardness : TextbookDLWithCoinsAdvantageLE B (family.relationFinder hchar)
```
daira: This field bounds the named relation finder's winning measure and is generic in `advantage` — it asserts nothing about hardness. The name invites the reading that the theorem takes DL hardness as a premiss, and we initially misread it that way. Consider `finderAdvantageLE` or similar.

book/src/formal-verification/proof-journey-data.json
```diff
@@ -443,34 +455,35 @@
...
         "The direct profiles separately account for random-oracle, group, and direct-coordinate work while leaving the finite-security Vesta DLOG advantage as a caller premise.",
-        "For every numProofs at or below Orchard's consensus maximum, the generic adaptive Action knowledge endpoint bounds verifier acceptance when executable private-witness extraction fails by Adv_DLOG(2^127,2^127) + 2^-83, prices all five semantic challenge surfaces, profiles the shared extractor, and exposes an explicit PMFEventBiasLE transport conjunct; it requires no externally supplied traces, cuts, or views."
+        "For every numProofs at or below Orchard's consensus maximum, the generic adaptive-statement Action knowledge endpoint bounds verifier acceptance when executable private-witness extraction fails by Adv_DLOG(2^126,2^126) + 2^-83 at a 2^123 computational work target, prices all five semantic challenge surfaces, profiles the shared extractor, and exposes an explicit PMFEventBiasLE transport conjunct; the 2^-83 term is an independent statistical soundness error, and the endpoint requires no externally supplied traces, cuts, or views."
```
daira: "prices all five semantic challenge surfaces" — the five surfaces are the four semantic-promotion surfaces (theta, beta, gamma, y) plus the constraint (x) surface, so "five semantic" miscounts.
```suggestion
        "For every numProofs at or below Orchard's consensus maximum, the generic adaptive-statement Action knowledge endpoint bounds verifier acceptance when executable private-witness extraction fails by <code>Adv_DLOG(2^126,2^126) + 2^-83</code> at a <code>2^123</code> computational work target. It prices the four semantic challenge surfaces (<code>θ, β, γ, y</code>) and the constraint surface (<code>x</code>), profiles the shared extractor, and exposes an explicit <code>PMFEventBiasLE</code> transport conjunct. The <code>2^-83</code> term is an independent statistical soundness error, and the endpoint requires no externally supplied traces, cuts, or views."
```

book/src/formal-verification/proof-map-embed.html
```diff
@@ -279,8 +279,8 @@ <h2>Detail</h2>
...
+    anchor:'Capstones.Action.orchard_action_knowledgeFailure_adaptiveStatement_2pow123_workFactor_generatorRO_for · Forking.Oracle.event_measure_le_of_bias',
+    role:'The consensus-generic adaptive-statement Action endpoint bounds the event that the verifier accepts while the executable extractor fails to return private witnesses for the entire bundle, for every numProofs at or below Orchard\'s consensus maximum. The adversary selects the public statement and proof together; the route binds the selected instance prefix before theta, prices all five Action challenge surfaces and the retained provenance execution, and transports the ideal-generator-RO event through an explicit PMFEventBiasLE premise.' },
```
daira: "prices all five Action challenge surfaces and the retained provenance execution" — the retained execution is not an event carrying a probability; suggest separating the clause, e.g. "... and reuses one retained provenance execution".
```suggestion
    role:'The consensus-generic adaptive-statement Action endpoint bounds the event that the verifier accepts while the executable extractor fails to return private witnesses for the entire bundle, for every consensus-valid number of proofs. The adversary selects the public statement and proof together; the route binds the selected instance prefix before theta, and prices all five Action challenge surfaces. It reuses a retained provenance execution, and transports the ideal-generator-RO event through an explicit statistical-distance premiss (<code>PMFEventBiasLE</code>).' },
```

daira (approved):
This PR closes a real gap between what the capstones proved and what the deployed verifier actually faces, and it does so carefully. Claude read every module doc and theorem statement in the new `AdaptiveStatement*` stack, worked through the load-bearing proofs in detail, and checked the deleted files for anything downstream still depending on them (nothing does; the replacements strictly subsume the old capstones). It did not find a correctness problem.

The heart of the change is easy to describe. In the old game, the public statement was fixed before the random oracle was sampled, so the fact that the verifier absorbs the vk and instance commitments never carried any weight in the proof. Now the adversary chooses the statement and the proof together, after making its oracle queries — and the statement still cannot be swapped afterwards, because the first thing the transcript hashes is the statement itself. The key step is definitional: the first oracle query literally *is* the statement-bearing transcript (`prefixesPre_zero_val` closes by `rfl`; Claude really liked this), and the initial transcript has a fixed, shape-determined length, so equal squeeze points force equal statements (`initialTranscript_eq_output`). The whole no-swap argument rests on those two facts, and they are exactly where a reviewer wants them: small, visible, and kernel-checked.

Some other things worth calling out, because they are what make a proof of this size reviewable. The extractor is proved to succeed whenever none of the named events fire (`adaptiveStatementKnowledgeExtractor_isSome_of_no_events`), which is what makes the failure bound a statement about something, rather than a bound on an event that might be empty for accidental reasons. Statement selection introduces no second `(Q+1)` query factor — every surface sits under the factor that was already there. And the doc comments state plainly where certification against code ends and declaration begins (the prover and reduction group-work fields), which saves time for future readers.

One thing to keep in view: everything here is for algebraic adversaries, with the challenge derivation modelled as a random oracle. Within that model, the theorem turns an adversary at the `2^123` work target into a discrete-log finder within `2^126` oracle queries and group operations —a three-bit overhead— with the `2^-83` statistical remainder proved at the full covered budget, not at a typical workload. Nothing about discrete-log hardness is assumed anywhere: the profile's advantage bound is the interface where the reader's own judgement about Vesta plugs in. The mathematics gets all of this right; our suggestions are mostly about making the surrounding prose as clear as the theorems.

The suggestions, none of them blocking:

- `AdaptiveStatementDlogProfile.hardness` names an interface that is generic in the bound and asserts nothing about hardness. We initially misread this — the field name invites "the theorem takes DL hardness as a premiss". Something like `finderAdvantageLE` would prevent that reading at the source.
- `ProbabilityVesta.lean` twice describes identifying halo2's hash-to-curve with the oracle as an assumption; "heuristic identification" is more accurate to the kind of step it is.
- Two small fixes in the book card data (line comments): "all five semantic challenge surfaces" is four semantic-promotion surfaces plus the constraint surface, and one clause applies a probability to an execution where it means an event.
- The proof-map and journey cards for the new capstones need a fuller rework than line comments can carry; we would treat that as follow-up rather than grow this PR. The next paragraph explains why.

The card descriptions need considerable work to live up to the standard of the mathematics. For example, the knowledge-capstone card says:

> The consensus-generic adaptive-statement knowledge capstone proves a 2^126 Vesta DLOG oracle/group resource ceiling for a 2^123 computational work target, with an independent 2^-83 statistical soundness error, for every consensus-valid Action count. Its instance, six-way pre-IPA, k-way IPA, five-way semantic, and source checks reuse one retained output/annotation execution before the quotient, identity, and terminal stages. The finite-security Vesta DLOG advantage at that ceiling remains a caller-supplied cryptographic premise, not a formula asserted by Lean.

This is extremely jargon-heavy, which obscures some real imprecision and omissions. (It's not the only example of undefined jargon, for example the journey card "exposes an explicit `PMFEventBiasLE` transport conjunct".)

A capstone does not prove a ceiling — it proves a bound, and the quoted numbers are coverage parameters. No card ever says which adversaries the guarantee covers (algebraic ones, with the challenge derivation modelled as a random oracle) — the model lives in the adversary's type, where a summary written from the theorem statement does not see it. The sentence enumerating six-way and k-way checks describes the proof's internal anatomy, when what a reader needs at this altitude is its meaning and how it fits into the development. And "the finite-security Vesta DLOG advantage at that ceiling remains a caller-supplied cryptographic premise" makes the same misreading the `hardness` field name invites: it reads as a careful concession, yet still implies the theorem takes hardness as a premiss, which it does not. It is a reduction in the AGM and does not need to do that, even if DL hardness could be stated without running into modelling issues.

None of this blocks the PR: the register is book-wide rather than new here, and the mathematics underneath is right. But the cards are the first thing a reader meets, and they deserve prose at the level of the theorems they describe — we would like a follow-up pass (coordinated with the planned book-side work) that gives every capstone card a statement of what is guaranteed, for which adversaries, at which coverage, in words that do not require reading the proof first.

Strong work, approved to merge.

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

daira (approved):
~~ACK~~ Failing CI.

Zcash/Snark/Soundness/Action/AdaptiveStatementProfile.lean
TalDerei: Renamed the field to `finderAdvantageLE` in e4e61995 so the API describes the bound it carries without implying a hardness premise.

TalDerei:
> This is extremely jargon-heavy, which obscures some real imprecision and omissions

agreed, and we've discussed offline about improving the representation of these which will be it's own focused effort. will fold this and other follow-up work described here into https://github.com/zcash/ironwood/issues/172.

PR #168, opened 2026-08-01
chore(fv): formalize public-instance handling in the verifier

TalDerei:
Largely lifting from summary shared OOB: Our FS model and squeeze schedule were verified against halo2 verifier's actual behavior (importantly absorbing the vk / instance commitments before squeezing the first challenge). The soundness capstones though technically prove weaker adaptive soundness: the statement is fixed before the RO is sampled, so the theorem never relies on that absorption (we’re not pinning what’s in the prefix “init”). I noticed this when inspecting [this code](https://github.com/zcash/ironwood/blob/90385ee234ba5b2975ad63ffe8039634219b2a40/Zcash/Snark/Soundness/AGM/AdaptiveOnline.lean#L1147-L1148) (where "init" is unconstrained in the sense that it’s universally quantified).

https://github.com/zcash/ironwood/pull/167 will add a stronger adaptive-soundness game where the adversary outputs the statement and proof jointly, after querying the oracle. In that game the transcript prefix is derived from the vk representation and the instance commitments of the statement it chose, so "init" stops being a free parameter.


Zcash/Snark/Core/ProofString.lean
```diff
@@ -46,6 +47,7 @@ structure Shape where
   numPermutationSets : ℕ
   numPermutationColumns : ℕ
   numQuotientPieces : ℕ
+  numInstanceColumns : ℕ := 1
```
TalDerei: this matches the fixture, but should be changed in https://github.com/zcash/ironwood/pull/162 i'd imagine. It's just metadata

Zcash/Snark/Verifier/FiatShamir.lean
```diff
@@ -70,6 +70,122 @@ def absorbLookup {F G : Type*} (e : LookupEval F) : List (TranscriptElt F G) :=
...
+    (instanceCommitment : Fin shape.numProofs → ℕ → G) : List (TranscriptElt F G) :=
+  (List.ofFn fun p : Fin shape.numProofs =>
+    List.ofFn fun column : Fin shape.numInstanceColumns =>
+      TranscriptElt.point (instanceCommitment p column)).flatten
+
+/-- The complete verifier-controlled Fiat--Shamir prefix: the VK transcript representation followed
+by every configured public-instance commitment. -/
+def initialTranscript {shape : Shape} {F G : Type*} (vkTranscriptRepr : F)
+    (instanceCommitment : Fin shape.numProofs → ℕ → G) : List (TranscriptElt F G) :=
+  TranscriptElt.scalar vkTranscriptRepr :: absorbInstanceCommitments instanceCommitment
```
TalDerei: defines the absorption: `initialTranscript = [VK] ++ instance commitments` for #167.

Zcash/Snark/Fixtures/MultiAction/FiatShamir.lean
TalDerei, line 1: there's some duplicated code across two `FiatShamir.lean` files in the fixtures that will be deduplicated and extracted into a shared file in #167.

Zcash/Snark/Core/Shape.lean
```diff
@@ -19,6 +19,7 @@ structure CircuitShape where
   numPermutationSets : ℕ
   numPermutationColumns : ℕ
   numQuotientPieces : ℕ
+  numInstanceColumns : ℕ := 1
```
mitschabaude, line 24: I recommend against giving this a default value. That just enables consumers to silently set a wrong field

Zcash/Snark/Fixtures/MultiAction/FiatShamir.lean
```diff
@@ -64,12 +65,39 @@ def capturedFs : FiatShamir Fp G := {
...
+  fun p => List.ofFn fun column : Fin shape.numInstanceColumns =>
+    capturedPublicInstances.getD
+      (p.val * capturedNumInstanceColumns + column.val) []
+
+/-- The composed rejecting verifier reproduces checked assembly at the captured challenges. -/
+theorem assembleNonInteractiveInstances?_matches_captured :
+    assembleNonInteractiveInstances? capturedFs capturedVkTranscriptRepr vk
+      capturedRawInstances commitLagrange ps =
+        assemble? vk derivedInstanceCommitment ps ch := by
+  native_decide
```
mitschabaude: This native decide seems to mix two statements:
- validateInstance on the LHS returns 'some'
- assemble? agrees between the two sides

The first one should be much cheaper to check, and the second might be something we already check elsewhere. If so, they should be separated to avoid running assemble through native_decide a second time.
In general we should keep our native_decides to minimal statements

Zcash/Snark/Fixtures/MultiAction/FiatShamir.lean
```diff
@@ -64,12 +65,39 @@ def capturedFs : FiatShamir Fp G := {
 theorem deriveChallenges_matches_captured_schedule :
     deriveChallenges capturedFs capturedInit ps = ch := by native_decide
 
+/-- The captured verifier prefix is exactly the canonical VK-and-instance prefix. -/
+theorem capturedInit_eq_initialTranscript :
+    capturedInit =
+      initialTranscript capturedVkTranscriptRepr derivedInstanceCommitment := by
+  native_decide
```
mitschabaude: This native decide seems to be unnecessary, follows by a simple proof from instance_commitments_derived. Let's not add new native decides carelessly

mitschabaude (commented):
Partial review

Zcash/Snark/Core/Shape.lean
TalDerei: https://github.com/zcash/ironwood/pull/168#discussion_r3696990003 removing the specialization.

Zcash/Snark/Fixtures/MultiAction/FiatShamir.lean
TalDerei: yeah this doesn't need to be a native_decide, proving it compositionally.

Zcash/Snark/Verifier/FiatShamir.lean
```diff
@@ -70,6 +70,122 @@ def absorbLookup {F G : Type*} (e : LookupEval F) : List (TranscriptElt F G) :=
...
+
+theorem instanceCommitment_mem_initialTranscript {shape : Shape} {F G : Type*}
+    (vkTranscriptRepr : F) (instanceCommitment : Fin shape.numProofs → ℕ → G)
+    (p : Fin shape.numProofs) (column : Fin shape.numInstanceColumns) :
+    TranscriptElt.point (instanceCommitment p column) ∈
+      initialTranscript vkTranscriptRepr instanceCommitment := by
+  simp [initialTranscript, absorbInstanceCommitments]
+
+/-- Equality of canonical verifier prefixes pins every configured instance commitment. -/
+theorem instanceCommitment_eq_of_initialTranscript_eq
```
ebfull, line 158: Consider adding the trivial companion `vkTranscriptRepr_eq_of_initialTranscript_eq` (`List.cons.inj … |>.1` plus `TranscriptElt.scalar` injectivity), so the generic layer states the complete binding fact: equal canonical prefixes determine *everything* absorbed, not only the instance half. #167's model fixes the VK at deployment and so never needs it, but this file is where the complete statement belongs.

Zcash/Snark/Verifier/Deployed.lean
```diff
@@ -0,0 +1,59 @@
...
+namespace Zcash.Snark
+
+open Zcash.Arithmetic (Msm)
+
+/-- Checked non-interactive MSM assembly from raw public-instance columns.
+
+This is the generic verifier entry corresponding to Halo2's `verify_proof` control flow: reject
+malformed instance shapes, derive their commitments, bind the VK and all instance commitments into
+Fiat--Shamir, derive challenges, and finally run the checked MSM assembler. -/
+def assembleNonInteractiveInstances? {shape : Shape} {F G : Type*}
```
ebfull, line 30: `assemble` reads the commitment family at the columns named by `vk.instanceQueryLayout`, while the transcript absorbs only columns `< shape.numInstanceColumns` — the `finFnG` docs already note that faithfulness rests on the VK's query indices being in range. In this composed entry the gap is bounded: out-of-range columns hit the constant `commitColumn []`, so prefix-determinism of the output survives regardless. But connecting general `ℕ → G` commitment families to this entry (as the adaptive layer does) wants the reusable form: a congruence lemma that families agreeing on columns `< numInstanceColumns` assemble identically when the layout is in range. #167 proves the layout bound concretely for the Action circuit (every column `< 1`); the generic congruence would sit naturally next to `assembleQueries`.

Zcash/Snark/Verifier/Deployed.lean
```diff
@@ -0,0 +1,59 @@
...
+# Halo2-faithful non-interactive verifier entry
+
+This module composes the raw-instance rejection path with the statement-bound Fiat--Shamir
+schedule.  Raw public columns are checked before they are committed.  The VK transcript
+representation and every resulting instance commitment are then absorbed before proof-controlled
+advice commitments and the first challenge.
+
+The column commitment operation remains a parameter here.  For Orchard it is instantiated by the
+existing Lagrange-basis commitment model; this entry point controls when it may run and how its
+outputs enter both the transcript and opening-query assembly.
```
ebfull: This docstring is where the binding claim is made, so it is the natural home for one sentence each on what the claim does *not* cover (with a one-clause pointer from the new `TrustBoundary.lean` section):

- after the absorption sentence: binding the key means binding its transcript representation — nothing ties the opaque `vkTranscriptRepr` scalar to `vk`'s fields; halo2's Blake2b hash of the pinned key, and its collision resistance, stay below the abstraction line.
- appended to the `commitColumn` paragraph: binding is at commitment level — the deployed Lagrange commitment zero-pads columns to the usable length, so raw columns differing only in trailing zeros commit, and verify, identically (as in halo2).

ebfull (commented):
*Note: this review was prepared with Claude Code assistance.*

As the base of the #167 stack, the preparation layer here looks sound: the canonical prefix as a function of the statement (`initialTranscript`), the prefix-injectivity lemma (consumed twice in #167's adaptive-statement reduction), the pre-`theta` membership lemmas, and the constant `initialTranscript_length` (which the soundness oracle-domain type depends on) are the right ingredients — and pinning them in `TrustBoundary.lean` before anything consumes them keeps their axiom bases constrained from day one.

Three non-blocking suggestions inline: a companion injectivity lemma for the VK repr, a reusable in-range congruence for `assemble`, and a docstring note delimiting what the binding claim does not cover.

Style nit while in these files: the new module docstrings write `Fiat--Shamir` (TeX double hyphen); house style is the en dash, `Fiat–Shamir`.

ebfull (approved):

Zcash/Snark/Core/Shape.lean
TalDerei: i remembered now why i had to specialize it; generated fixture doesn't emit `numInstanceColumns := 1`, so it inherited that default. will add a todo comment to fix this in the next upstream halo2 release in https://github.com/zcash/halo2/pull/924.

TalDerei: https://github.com/zcash/ironwood/pull/168/changes#diff-28e14fb042d444237d7dee9478847abb4db6aa986e8dc9d4e5e7eeb4bb6572b2R22-R24

daira (commented):
Post-hoc ACK

PR #169, opened 2026-08-02
Certify top-level lookup registration against the raw configure result

ebfull:
Clean derives the keygen constraint system of a `TopLevelCircuit` by closing the raw configure result under the circuit's synthesis stream; `closeWithOperations` appends any synthesis-enabled lookup that configure never registered. Halo 2 performs no such repair — synthesis can only enable a lookup already established by configuration — so a circuit that needs it is unlawful, and the closure hides that instead of rejecting it. The generic lookup bridge drew its registration facts from `TopLevelCircuit.keygenCoherent`, which holds by construction of exactly that closure, so nothing on the live path ruled out a silently repaired lookup list: gate-list and selector-count closure were already pinned inactive by `gateData_eq`, but the lookup list had no counterpart, and the `TopLevelGateCoherence` docstring presented the omission as sound by construction.

Close the lookup half of the gap at the static boundary:

* `TopLevelGateCoherence` now requires `lookups_eq_configure`: the synthesis-closed lookup list equals the raw configure lookup list, so closure cannot smuggle an unregistered lookup into the derived verifying key. The new Action certificate `lookupData_eq` discharges it for the deployed circuit by native computation. Together with `gateData_eq` this forces both missing-argument lists empty — hence the query-registration folds inactive — so every mutating component of the closure is provably inert on the Action path.
* The new `enabledLookup_configured` transports every synthesis-enabled lookup back to the raw configure list, and `projectedValues` now proves its route-index membership through that transport rather than a bare `List.getElem_mem`, making the configure-boundary check load-bearing in the generic lookup projection.
* Every `+native` trust-census entry reached by the Action capstone and terminal theorems names the new computation; the exact-match census keeps those owner lists build-enforced.

The generic compiler still performs the repair for an arbitrary `FormalCircuit`, and the Action discharge is a whole-circuit computation rather than a compositional configure/synthesis law, so obligation 26 remains in the lawfulness-cleanup inventory (updated: live-path computation count 23, row 26 marked enforced but computational). Gate-list and selector-count closure inactivity are likewise still certified only through the Action instance, not demanded generically by the boundary.


Zcash/Circuits/Integration/ActionGateCoherenceCompute.lean
TalDerei, line 1: we _may_ be able to cleanly get rid of all the `native_decide` calls in this file to avoid unnecessarily increasing our trust base, or it may require a clean-side change. cc @mitschabaude

TalDerei (approved):
afaict doesn't seem like an soundness hole that you could exploit, just a proof architecture gap.

mitschabaude:
I have a PR stack which eliminates this entire problem, it gets rid of `closeWithOperations` and proves the corresponding lawfulness locally in each circuit bundle

Zcash/Circuits/Integration/ActionGateCoherenceCompute.lean
mitschabaude: yes, and those changes as well underway

Issue #147, opened 2026-07-29
Compute the circuit-to-ledger Action witness bridge


TalDerei:
PR #114 now returns executable `ActionBundleWitness` data and bounds acceptance when that extractor fails. However, `Bridge.specPost_to_ledger` still forgets the successful conversion into:

```lean
∃ inst w, PublicProjection ... ∧ ActionSatisfied ...
```

Because this lives in `Prop`, the extracted circuit witness cannot be passed as data into the witness-annotated ledger games. We therefore have Action-circuit knowledge soundness, but only ordinary ledger soundness.

Required:

- Define a `Type`-valued ledger success object containing `inst`, `ActionWitness`, and its validity certificates.
- Compute either that object or the existing `ActionDLBreak` from each extracted Action witness.
- Lift this conversion across `ActionBundleWitness`.
- Compose it with #114’s extractors and bound acceptance when ledger extraction fails.
- Retain `specPost_to_ledger` as the proposition-level corollary.
- Add computability/trust-boundary checks and connect the result to the ledger games.

This is distinct from #22, which tracks RedDSA `bsk` extraction, and belongs under the #16 verifier-knowledge-soundness umbrella.

TalDerei:
Depends on #114 and #139.

daira:
Source check against the current tree (the PR #171 head):

* `Bridge.specPost_to_ledger` was renamed: the bridge theorem is now `Bridge.actionSpec_to_ledger` (`Zcash/Security/Ledger/Bridge.lean`). The structural findings are unchanged: it is Prop-valued, concluding `ActionBreak … ∨ ∃ inst w, PublicProjection … ∧ ActionSatisfied … ∧ CrossAddressSatisfied … ∧ EnableFlagsSatisfied …`, so the constructed instance, witness, and certificates are discarded under the existential.
* The circuit side now ends in Type-valued data: `ActionBundleWitness` (`Snark/Soundness/Action/StraightLineTerminal.lean`), with `actionKnowledgeExtractor` returning `Option (ActionBundleWitness inputs)` — but its only ledger-facing export is the Prop projection `ActionBundleWitness.statement` (the `ActionSpec` statement), which is what the bridge consumes. The data is dropped at the seam.
[...]

PR #171, opened 2026-08-04
fv: the conservation relation arm and the conservation experiment

### LINKED ISSUE #107: Binding-signature glue and the games' named hypotheses

daira:
The Balance-closure follow-up to #157, continuing #107's residual list. Three units: the conservation relation arm placed in the oracle model, the knowledge error restated with a single randomized reduction, and the conservation experiment composing both arms in one sample space through one combined finder. Later commits factor the shared experiment abstractions, add the integrity experiment composing the non-negativity and conservation arms in one sample space, discharge that experiment's non-negativity side at the Orchard instantiation, bundle the algebraicity hypotheses as a structure, strip the per-arm bounds the combined finder superseded, pin the experiment's apparatus at its deployed choices, and restate the ledger's no-overflow bound in the binding-signature layer's vSum shape.

## The binding-signature relation at basis slots

`toAlgebraicRelationWitnessAt` places the binding-signature relation's coefficients at two slots of the presented basis, vanishing elsewhere. It is the adapter between the reduction's break type and the relation-to-discrete-log finder input, and it lives with the binding-signature reduction. The fixed two-base form had no consumer beyond its census entry and is removed.

## The conservation relation arm in the oracle model

`Ledger/ValueRelationArm.lean` places the relation arm —the `εdlr` slot— in the challenge-oracle model, for any labeled ledger adversary. No query-budget term arises from this arm: its witness is oracle-free data, so there is no bad-challenge accounting. The relation branch fires at the prefix's first imbalanced transaction, and the selection is computed data. The finder replays the adversary, recomputes the extractor's value from the run's own data, and rebuilds the relation behind decidable guards; on every relation-arm sample it returns a relation, so validity is a conjunct of the measured event rather than a hypothesis. The conservation experiment's combined finder consumes the arm.

## The knowledge error with a single randomized reduction

`kappaEvent_measure_le_of_coins` restates #157's $\kappa$ bound with one discrete-log bound for the coin-consuming relation finder. This removes the supremum over challenge tables that the fibre-wise composition had noted as its artefact. The bad-challenge arm is unchanged.

## The conservation experiment

`Ledger/ConservationExperiment.lean` composes the capstones' `εdlr + κ` split in one experiment, at the reduction layer. Both arms' finders return relations over the same presented basis, so one combined machine (`conservationRelFinder`) replays the adversary and returns whichever relation the sample yields. Over the adversary's coins, the challenge table, and the logs of the presented bases, a valid output ledger violates balance conservation (or the shielded-balance cap) at some prefix with probability at most $`\varepsilon_{\mathrm{dl}} + \frac{q_H + 2}{\#\mathbb{F}}`$: every violation sample lands in the combined machine's relation event or in the knowledge-error bad-challenge fibre, and one textbook discrete-log bound covers both relation slices at once, so the two arms' programmed-basis losses merge. The discrete-log hypothesis is a single bound for the combined coin-consuming finder, per fixed adversary coins. No supremum over challenge tables remains anywhere in the experiment. The per-arm measure bounds this composition superseded are stripped; the arm modules keep the finders, selections, and containments the combined finder consumes.

## The integrity experiment

`Ledger/IntegrityExperiment.lean` composes the non-negativity and conservation arms of balance integrity in the same experiment. A premiss-free containment (`balanceIntegrityViolationBefore_subset_conservation`) splits an integrity violation into the three non-negativity arms (merkle, note-commitment, key-binding) and the conservation violation — all per-primitives events. The conservation arm is bounded wholesale by the conservation experiment above; the non-negativity side is taken as a single named bound $`\varepsilon_{\mathrm{nonneg}}`$ on the combined (unioned) arm event — rather than three separate arm bounds — which is what lets an instantiation discharge all three by one advantage. So the bound is $`\varepsilon_{\mathrm{nonneg}} + \left(\varepsilon_{\mathrm{dl}} + \frac{q_H + 2}{\#\mathbb{F}}\right)`$, with no factor of the prefix count.

The composition is done at the reduction layer, not the probability layer. The arms live over different sample spaces, so rather than combine probability statements, the decomposition is a distribution-independent set-level containment lifted to a probability bound. The lift `sampledLedgerEvent` is a monotone join-homomorphism (`sampledLedgerEvent_mono` preserves $\subseteq$, `sampledLedgerEvent_union` preserves $\cup$); composed with the outer measure's subadditivity —the measure of a union is at most the sum of the measures of its parts— it turns the containment into the sum of per-arm bounds. The sample distribution (`challengeExperiment`) and the event lift (`sampledLedgerEvent`) are factored into shared defs —also applied to the conservation experiment— to keep the statements readable.

## The Orchard integrity experiment

`Ledger/OrchardIntegrityExperiment.lean` discharges the non-negativity side at the Orchard instantiation: the three Balance-subset arms collapse onto the single advantage $`\varepsilon_{\mathrm{sinsemilladlr}}`$ of computing a nontrivial discrete-log relation among the fixed Sinsemilla bases — matching `orchardBalanceIntegrity_measure_le` at the deployed capstone layer, but now over the challenge experiment. A valid output Orchard ledger then violates balance integrity at some prefix with probability at most $`\varepsilon_{\mathrm{sinsemilladlr}} + \left(\varepsilon_{\mathrm{dl}} + \frac{q_H + 2}{\#\mathbb{F}}\right)`$ — one discrete-log advantage per side.

The enabler is that `kappaPrimitivesAt` is a record update of the deployed primitives touching only the value-commitment and binding-verification fields, and the three arms read only the note-commitment, key-binding, and Merkle structure, which the sampling leaves fixed. So a sampled Balance-subset break is the same data as at the deployed primitives, and the Orchard reducers apply unchanged. The discharge is deterministic and needs no value/binding-side hypotheses, so it is independent of the conservation side, which the experiment still carries as the combined coin-consuming finder's bound.

A concluding endpoint (`orchardBalanceIntegrity_measure_le_idealizedks`) pins the experiment's apparatus at the deployed parameters: two presented bases, the standard Pallas generator as the discrete-log base, the challenge query as the literal $`(R, \mathsf{bvk}, \mathsf{sighash})`$ triple (injective by construction), signatures decoded as themselves, and binding verification as RedPallas at $`\mathcal{R}^{\mathsf{Orchard}}`$ (for which we require only extractability here). Its docstring classifies every remaining hypothesis as a resource parameter, a documented idealization, or a cited consensus rule. The action cap is the dedicated rule on the action count (applying to either Orchard-protocol pool), and the binding challenge hash is modelled as a random oracle. The spend-authorization predicate stays universally quantified — Balance relies on no property of it, and in particular does not need the sighash to commit to the transaction effects.

The no-overflow side is stated in the binding-signature layer's vSum shape —the primitives carry a declared-balance bound `vBalanceBound` (deployed $`2^{63}`$) beside `valueBound`— so the Orchard discharge is monotonicity into that layer's one numeric anchor, with no duplicated arithmetic. The conservation and cap experiments have the same forms (`orchardBalanceConservation_measure_le_idealizedks`, `orchardShieldedBalanceCap_measure_le_idealizedks`).

The adversary model is a named type, `IdealizedKSBalanceAdversary`. Its docstring separates the accepted idealizations (the random-oracle challenge hash, the programmed bases with the reference-string heuristic, and the elided encodings) from the critical gap, i.e. that the witness annotations idealize knowledge soundness of the Action circuit. The latter is tracked by #147, and the `_idealizedks` in the names marks that dependence.

## The algebraicity hypotheses, bundled

Satisfying a review comment, the two algebraicity hypotheses the challenge-oracle experiments shared are bundled as the `AlgebraicAtBindingPoints` structure, with `atLabel` and `atOutput` as named fields; each experiment takes one bundled hypothesis per adversary coin. The structure lives beside `bvkAt` and `extractFail_mem_kappaEvent` in `Ledger/ExtractionKnowledgeError.lean` — renamed from `ExtractionKappaArm.lean` to say what the file does.

## Documentation and notation

The TrustBoundary census section summaries are raised to high-level interpretations — dropping restated bound formulas, which go stale as the bounds change, while keeping the qualitative tightness. Summaries are added to the ledger sections that lacked one. The `[k] P` bracket notation for scalar multiplication is replaced by Lean's `•` across the circuit and security docstrings, and the notation for the field size is made consistent (`#F` in the branch's comments, spelled using the `\FieldSize` macro in the book's formal-verification pages). The Bridge layer's verification-predicate binders are named `spendAuthVerify` and `bindingVerify`, matching the `Primitives` fields.

## Census patterns for qualified endpoint names

The endpoint-census patterns matched the semantic suffixes only bare or with a trailing `_for`, so the `_experiment`- and `_idealizedks`-qualified endpoint names escaped the census obligation — they were pinned, but nothing demanded it (reported by Tal in review). Both pattern sides (`ENDPOINT_RE` and its Lean port `isEndpointBaseName`) now allow the qualifier trailers, raising the demanded endpoint count from 154 to 161, all already pinned. A stale comment on the no-overflow hypothesis is also dropped.

Closes #107.

🤖 Claude Fable 5 and Claude Opus 4.8 (1M context)


TalDerei (approved):
did a pass over this and it looks good so far. optimistic ACK

Zcash/Security/Ledger/ValueRelationArm.lean
```diff
@@ -0,0 +1,359 @@
...
+import Zcash.Common.RelationProbabilityCoins
+
+/-!
+# The conservation relation arm, discharged in the oracle model
+
+The capstone layer bounds the conservation reduction's relation arm by the named `εdlr`. This
+module bounds that arm in the challenge-oracle model, at the reduction's own events: for any
+labeled ledger adversary, the probability that the output ledger is valid *and* the reduction
+computes a nontrivial `(𝒱, ℛ)` relation is at most `ε_DL + 1/|F|` (from
+`relation_prob_le_of_textbookDL`). No query budget appears: the arm's witness is oracle-free
```
dannywillems: Nit: `relationWithCoins_prob_le_of_textbookDL`? There is no reference of `relation_prob_le_of_textbookDL`.

Zcash/Security/BindingSignature/DiscreteLog.lean
```diff
@@ -19,42 +19,35 @@ namespace Zcash.Security.BindingSignature
...
-      · rcases hαβ with hα' | hβ'
-        · exact hα' hα
-        · exact hβ' hβ
-    relation := by
-      rw [representationEval_bindingSignatureBasis]
-      simpa [Zcash.commitGen] using r.relation }
+/-- Convert a binding-signature relation to the generic relation type over a basis that
+carries `Vbase` and `Rbase` at two distinct slots. The relation's coefficients are placed at
+those slots and vanish elsewhere. -/
+def NontrivialRelation.toAlgebraicRelationWitnessAt {m : ℕ} {Vbase Rbase : M}
```
dannywillems, line 25: Reviewer notes: this generalizes the previous version to any finite base size (basis, Fin m), the coefficients being in a type `NontrivialRelation` now, parametrized by the bases V and R.
LGTM.

Zcash/Security/Ledger/ConservationExperiment.lean
```diff
@@ -0,0 +1,216 @@
...
+basis logs, the probability that the output ledger is valid and violates balance conservation
+at some prefix `i < k` — the capstone's `balanceConservationViolationBefore`, at the sampled
+primitives — is at most `(ε_rel + 1/|F|) + ((qH+2)/|F| + ε_κ)`. The two discrete-log
+hypotheses are single bounds for the coin-consuming finders, per adversary coin. -/
+theorem balanceConservationBefore_measure_le_experiment {ι : Type u} (p : PMF ι)
+    {LA : ι → (Fin m → G) → LabeledOracleComp Q (ZMod r) (fun _ => QueryRep (ZMod r) m)
+      (List (Tx KW (ZMod r) G RHO PSI MHASH MENC MSG SIG P₀.depth × QueryRep (ZMod r) m))}
+    (hne_idx : v_idx ≠ r_idx)
+    {qH : ℕ} (hQ : ∀ j b, (LA j b).QueryBound qH)
+    (halgLabel : ∀ (j : ι) (O : Q → ZMod r) (s : Fin m → ZMod r) (q : Q)
```
dannywillems: Nit/suggestion: `halgLabel` and `halgOut` are defined twice. An agent refactorized in https://github.com/daira/ironwood/pull/2 to bind it to a struct.

Zcash/Security/Ledger/ConservationExperiment.lean
```diff
@@ -0,0 +1,216 @@
...
+  · exact Or.inr ⟨hval, i, hik, e, he⟩
+
+/-- **The cap experiment.** As `balanceConservationBefore_measure_le_experiment`, for the
+shielded pool exceeding the minted issuance at some prefix `i < k`. -/
+theorem shieldedBalanceCapBefore_measure_le_experiment {ι : Type u} (p : PMF ι)
+    {LA : ι → (Fin m → G) → LabeledOracleComp Q (ZMod r) (fun _ => QueryRep (ZMod r) m)
+      (List (Tx KW (ZMod r) G RHO PSI MHASH MENC MSG SIG P₀.depth × QueryRep (ZMod r) m))}
+    (hne_idx : v_idx ≠ r_idx)
+    {qH : ℕ} (hQ : ∀ j b, (LA j b).QueryBound qH)
+    (halgLabel : ∀ (j : ι) (O : Q → ZMod r) (s : Fin m → ZMod r) (q : Q)
```
dannywillems: Same comment as https://github.com/zcash/ironwood/pull/171/changes#r3834373231

Zcash/Security/Ledger/ConservationExperiment.lean
```diff
@@ -0,0 +1,216 @@
...
+namespace Zcash.Security.Ledger.Model
+
+open Zcash.Common Zcash.Security.BindingSignature Zcash.Security.RedDSA Zcash.Snark
+open Zcash.Common.LabeledOracleComp
+open scoped ENNReal
+
+universe u
+
+variable {r : ℕ} [Fact (Nat.Prime r)]
+variable {G : Type*} [AddCommGroup G] [Module (ZMod r) G] [DecidableEq G]
```
dannywillems, line 37: Pure question, no suggestion: I can see that we redefine aliases for each structure in each module, scoped to the actual requirement of the module. Is it something intended? Would it be possible to centralize the definitions somewhere, and reuse it, if it eases visibility? We could keep the same naming convention for each object we're manipulating.

dannywillems: Though, I like having to read the beginning of the file to get the "context" of all the mathematical objects we would manipulate in this module.

Zcash/Security/Ledger/ConservationExperiment.lean
```diff
@@ -0,0 +1,253 @@
...
+The composition is the capstone layer's own: at each sample, a violation lands in the
+transaction-balance premiss's relation arm or its extraction-failure arm
+(`balanceConservationViolationBefore_subset_fallible`), and each arm's measure is the bound
+already proved for it — the relation arm at `ε_rel + 1/|F|`
+(`balanceConservationBefore_valueRelationArm_measure_le`) and the extraction-failure arm at
+`(qH+2)/|F| + ε_κ` (`balanceConservationBefore_extractFailArm_measure_le_of_coins`, the
+coins form proved here). Both discrete-log hypotheses are single bounds for coin-consuming
+finders (`TextbookDLWithCoinsAdvantageLE` at `ρ` := the challenge table), quantified only
+over the adversary's coins: no supremum over challenge tables remains anywhere in the
+experiment.
```
daira, line 25: There are some long sentences and some awkward phrasing here.

Zcash/Security/Ledger/OrchardIntegrityExperiment.lean
```diff
@@ -0,0 +1,191 @@
...
+balance-integrity violation by `ε_nonneg + conservation` over the challenge-oracle sample space,
+taking the union of the three Balance-subset arms (merkle, note-commitment, key-binding) as one
+named bound `ε_nonneg`. This module discharges `ε_nonneg` at the Orchard instantiation: all three
+arms collapse onto the single advantage of computing a nontrivial discrete-log relation among the
+fixed Sinsemilla bases, exactly as `orchardBalanceIntegrity_measure_le` does at the deployed
+capstone layer, but now over the challenge experiment's sample space.
+
+The enabler is that `kappaPrimitivesAt` is a record update of the deployed primitives touching only
+`valueCommit` and `bindingVerify`, and the three non-negativity arms read only the note-commitment,
+key-binding, and Merkle structure — which the sampling leaves fixed. So at the sampled primitives a
```
daira: This could be a bit more clearly stated. It's just a [programmed-basis reduction](https://zcash.github.io/ironwood/formal-verification/security-models.html#what-a-reduction-in-these-models-says).

Zcash/Security/Ledger/ValueRelationArm.lean
daira: Fixed. `relation_prob_le_of_textbookDL` does exist (it's the non-coins version in `Common/RelationProbability.lean`), but this arm's finder is coin-consuming, and the proofs use `relationWithCoins_prob_le_of_textbookDL` (`Common/RelationProbabilityCoins.lean`), so the docstring now cites that. Folded into the introducing commit.

🤖 Reply via [Claude Code](https://claude.com/claude-code)

Zcash/Security/Ledger/ConservationExperiment.lean
daira: Yes it's intended. I like this style.

Zcash/Security/Ledger/ConservationExperiment.lean
daira: Adopted — thanks for the pointer. Bundled the two into your `AlgebraicAtBindingPoints` (the per-coin `atLabel`/`atOutput` form from daira/ironwood#2), so each experiment now takes one `AlgebraicAtBindingPoints` per adversary coin instead of the `halgLabel`/`halgOut` pair. I moved it to `ExtractionKnowledgeError.lean` (renamed from `ExtractionKappaArm.lean`), beside `bvkAt` and `extractFail_mem_kappaEvent`, whose module docstring covers the assumption. Pushed.

Zcash/Security/Ledger/OrchardIntegrityExperiment.lean
```diff
@@ -0,0 +1,277 @@
+import Zcash.Security.Ledger.IntegrityExperiment
+import Zcash.Security.Ledger.OrchardCapstone
+import Zcash.Security.BindingSignature.Orchard
+
+/-!
+# The Orchard integrity experiment: the non-negativity arms collapse to one Sinsemilla DLR
+
+The abstract integrity experiment (`balanceIntegrityBefore_measure_le_experiment`) bounds a
+balance-integrity violation by `ε_nonneg + conservation` over the challenge-oracle sample space,
+taking the union of the three Balance-subset arms (merkle, note-commitment, key-binding) as one
```
daira:
```suggestion
taking the union of the three Balance-subset arms (Merkle, note-commitment, key-binding) as one
```

Zcash/Security/Ledger/ConservationExperiment.lean
daira: Rewritten with shorter sentences in 9567cc76.

Zcash/Security/Ledger/OrchardIntegrityExperiment.lean
daira: Restated in 9567cc76: the paragraph now opens by naming it a programmed-basis reduction, links the security-models section, and keeps the record-update detail as the mechanism.

Zcash/Security/Ledger/OrchardIntegrityExperiment.lean
daira: Applied in 9567cc76, also in the sibling arm lists.

daira (commented):
Self-ACK

daira (commented):
Self-reACK

Zcash/Security/Ledger/Model.lean
```diff
@@ -196,9 +196,10 @@ structure ValidLedger (P : Primitives F G IVK NK RHO PSI MHASH MENC MSG SIG)
   /-- Each transaction's declared value balance is in range. The consensus rule
   (§7.1.2, [NU5 onward]) is `valueBalanceOrchard ∈ {-MAX_MONEY .. MAX_MONEY}` (inclusive;
   MAX_MONEY ~2^51 zatoshi), i.e. `|vBalance| <= MAX_MONEY`. This conjunct states the
-  weaker `natAbs < valueBound` (= 2^64 at the intended instantiation), a loose consequence
-  that suffices for the no-overflow bound in the integer-balance lift. -/
-  vbalance_bound : ∀ tx ∈ ledger, tx.vBalance.natAbs < P.valueBound
+  weaker `natAbs ≤ vBalanceBound` (= 2^63, the signed two's-complement endpoint, at the
+  intended instantiation), a loose consequence that suffices for the no-overflow bound in
+  the integer-balance lift. -/
+  vbalance_bound : ∀ tx ∈ ledger, tx.vBalance.natAbs ≤ P.vBalanceBound
```
dannywillems, line 202: Reviewer note: this is related to the other transformations from the strict inequality to inequality at other places.

Zcash/Security/Ledger/Completeness.lean
```diff
@@ -294,7 +294,7 @@ theorem honestTx_valid
         + ((issuance ledger.length : ℤ) + (honestTx spends sighash bindingSig).vBalance))
     (hbind : P.bindingVerify ((honestTx spends sighash bindingSig).bvk P) sighash
         bindingSig)
-    (hvb : (honestTx spends sighash bindingSig).vBalance.natAbs < P.valueBound)
+    (hvb : (honestTx spends sighash bindingSig).vBalance.natAbs ≤ P.vBalanceBound)
```
dannywillems, line 297: Reviewer note: vBalanceBound is set to 2^63. Inequality non strict is actually better now, as we have an integer in [-2^63; 2^63 - 1] (abs value is therefore correct).
`P.valueBound` is set as `valueBound := 2 ^ 64`. Therefore, before the hypothesis was larger.

Zcash/Security/Ledger/ExtractionArm.lean
```diff
@@ -79,7 +99,7 @@ def ValueShape.premissOrBreakFallible [DecidableEq G]
     (S : ValueShape P) (B : BindingSigShape P S)
     (hval : ValidLedger P kv issuance maxActions ledger)
     -- + 1 is for the valueBalance
```
dannywillems: Outdated comment?

Zcash/Security/Ledger/ExtractionArm.lean
```diff
@@ -79,7 +99,7 @@ def ValueShape.premissOrBreakFallible [DecidableEq G]
...
-    (hr : (maxActions + 1) * P.valueBound ≤ r)
```
dannywillems, line 82: Reviewer note: new condition is weaker/more constrained - ok.

dannywillems, line 82: Reviewer note:

- `(2^16 - 1 + 1) * 2^64 = 1208925819614629174706176`
- `(2^16 - 1) * (2^64 - 1) + 2^63 = 1208916596242592319864833`

book/src/formal-verification/security-models.md
```diff
@@ -22,9 +22,74 @@ So each definition sits on a three-layer stack:
...
+  and combined by union $\cup$.
+- The *probability measure*, $\mu$, of an event set is **monotone** and
+  **finitely subadditive**. Monotone means that if $A \subseteq B$ then
+  $\mu(A) \le \mu(B)$ (a subset of events has no greater probability than the
+  original set). Finitely subadditive means that
+  $\mu(A \cup B) \le \mu(A) + \mu(B)$ (the probability of a union is at most
+  the sum of the probabilities of its parts). That's all we need: no independence
+  and no inclusion–exclusion principle.
+- The *lift that carries a per-parameter event into the sample space* is a
+  **monotone join-homomorphism** — it preserves both $\subseteq$ and $\cup$.
```
dannywillems, line 55: Reviewer note/reading: https://en.wikipedia.org/wiki/Semilattice.

book/src/formal-verification/security-models.md
```diff
@@ -22,9 +22,74 @@ So each definition sits on a three-layer stack:
...
+own bound. The composition happens at the **reduction layer** (Layer B), not the
+probability layer (Layer C).
+
+We use that approach because probability statements over *different* sample spaces
+don't combine straightforwardly — conditioning on a sub-event reweights the measure,
+and product measures and marginals get in the way. Computable reductions cause no
+such friction. A reduction is a total function from the adversary's output to a
+break, and so "run the adversary, then run the reduction" is just another machine
+(algebraic if both its components are) at an unchanged query count. Reductions
+compose by ordinary function composition —the path of least resistance that Lean's
```
dannywillems, line 40: > —the path of least resistance that Lean's proof tactics handle well—

Why is this added in this paragraph? It is okay, but I don't understand what kind of information it adds to it for the reader.

book/src/formal-verification/security-models.md
```diff
@@ -22,9 +22,74 @@ So each definition sits on a three-layer stack:
...
+  The lift is "there is a valid run, at the sampled parameters, whose output
+  lands in the event"; it preserves unions because $\exists$ distributes over
+  $\lor$.
+
+So a composed bound is proved in two moves. First, a distribution-independent
+**set-level containment**: the bad event is contained in a union of per-arm
+events. This involves no probability over multiple sample spaces, so it is
+easily reusable. Second, lift to a probability, and sum:
+
+* monotonicity carries the containment into the sample space;
```
dannywillems, line 65: Reviewer note: that's why the whole paragraph before the mathematical definitions aims to say. It is a prose for these mathematical transformations. Good composition diagram.

book/src/formal-verification/security-models.md
```diff
@@ -22,9 +22,74 @@ So each definition sits on a three-layer stack:
...
+and product measures and marginals get in the way. Computable reductions cause no
+such friction. A reduction is a total function from the adversary's output to a
+break, and so "run the adversary, then run the reduction" is just another machine
+(algebraic if both its components are) at an unchanged query count. Reductions
+compose by ordinary function composition —the path of least resistance that Lean's
+proof tactics handle well— and probability is taken *once*, at the end.
+
+Three pieces of structure make that last step essentially mechanical:
+
+- The *event sets* form a **Boolean algebra**, ordered by inclusion $\subseteq$
```
dannywillems, line 45: Reviewer notes (maybe wrong): I was wondering if we could be stricter and say that the sets are measure algebras. It is not wrong to say Boolean algebras though. When reading first, I had in mind the measure theory notions in my head.
Or maybe it is because we want to define the "lift" on boolean algebras, and this is transported in the measure? I should draw on paper the category/transformation diagram.

book/src/formal-verification/security-models.md
```diff
@@ -22,9 +22,74 @@ So each definition sits on a three-layer stack:
...
+### The Balance integrity argument as a worked example
+
+Balance integrity has exactly this shape. Its set of violation events —the
+shielded pool going negative, or the pools failing to sum to the minted
+issuance— is contained in the union of the three Balance-subset break arms
+(Merkle, note-commitment, key-binding) and the Balance conservation violation.
+That containment is `balanceIntegrityViolationBefore_subset_conservation`; it
+mentions no probabilities and holds at every prefix at once.
+
+Lifting it to the sample space through the join-homomorphism `sampledLedgerEvent`
```
dannywillems, line 78: Fascinating! And Claude can give a nice diagram when prompted to talk in terms of category.

dannywillems (approved):
utACK 22d44f3a6057d70f21d3f4de82ee72a7fc8c1bf7, no important comments.
I went through each theorem and definition, and written down on a paper, including ensuring that the diagram transformations make sense - one with the help of Claude for the reduction with the measures - awesome.

book/src/formal-verification/security-models.md
daira: It supports the point that it's a lot simpler than composing statements about probability.

book/src/formal-verification/security-models.md
daira: That's a good point. We would need to define what a measure algebra is, but that's what we're doing anyway here. I will make that change.

Zcash/Security/Ledger/ExtractionArm.lean
daira: Yes.

book/src/formal-verification/security-models.md
daira: A measure algebra would require $\mu(A) > 0$ when $A \neq \emptyset$ (Jech, https://arxiv.org/pdf/1705.01000 or definition (22.1) of https://fa.ewi.tudelft.nl/~hart/onderwijs/set_theory/Jech/22-saturated_ideals.pdf ), which we *don't* require here, because an event could have probability zero. I think in practice none of the events we consider have probability zero, but with the given definition, we don't rely on that or anything from the theory of measure algebras that depends on it.

daira: Here is the categorical picture, as requested.

**The categories.** Each event lattice is a *thin* (posetal) category: objects are events, with a (unique) morphism $A \to B$ iff $A \subseteq B$. Union is the finite coproduct. On the numeric side, $([0,\infty], \le)$ is likewise a thin category, with $+$ as a monoidal product.

**The two functors.** The composition apparatus is a composite of two monoidal functors of different strengths:

```
 (∏_P Ev(P), ⊆, ∪)     per-parameter events, pointwise order
        │
        │  L = sampledLedgerEvent
        │  (strong/strict monoidal functor)
        ▼
 (Ev(Ω), ⊆, ∪)         events over the sample space
        │
        │  μ = toOuterMeasure
        │  (oplax monoidal)
        ▼
 ([0,∞], ≤, +)
```

* $L$'s domain is the family category — per-parameter events with pointwise order and unions, since the lift takes a *family* $P \mapsto \mathrm{Event}(P)$. Functoriality of $L$ **is** `sampledLedgerEvent_mono`; monoidality **is** `sampledLedgerEvent_union`, and it is *strong* (preserves $\cup$ on the nose) because $\exists$ distributes over $\lor$.
* $\mu$'s functoriality **is** monotonicity, and subadditivity **is** precisely the *oplax* structure map $\mu(A \cup B) \to \mu(A) + \mu(B)$, a morphism in $([0,\infty], \le)$. It is oplax rather than strong: the map does not invert, and nothing requires it to.

**A composed bound is one diagram chase.** It starts from the set-level containment, a single morphism $c : V \to B_1 \cup B_2$ in the source category, and pushes it through the composite $\mu \circ L$, which is oplax monoidal (strong ∘ oplax = oplax):

```
 μ(L V)
    │   μ(L c)    (functoriality: monotonicity)
    ▼
 μ(L(B₁ ∪ B₂))
    ║             (join preservation: L(B₁ ∪ B₂) = L B₁ ∪ L B₂)
 μ(L B₁ ∪ L B₂)
    │             (oplax structure: subadditivity)
    ▼
 μ(L B₁) + μ(L B₂)
```

reading off $\mu(L\,V) \le \mu(L\,B_1) + \mu(L\,B_2)$. So the intuition above — the lift is defined on the Boolean algebras and transported in the measure — is exactly right: $L$ lives wholly in the event world, $\mu$ enters once at the end, and the two-move proof is "apply one oplax monoidal functor to one morphism".

**Where the measure algebra sits.** It is a third panel, orthogonal to the composition: quotient $\mathrm{Ev}(\Omega)$ by the null ideal $\mathcal{N}$, and $\mu$ descends to a strictly positive $\bar\mu$ — Jech's object:

```
 Ev(Ω) ────q────▶ Ev(Ω)/𝒩        (measure algebra, Jech)
      ╲               │
     μ ╲              │ μ̄  (strictly positive)
        ╲             ▼
          ▶  ([0,∞], ≤, +)          μ = μ̄ ∘ q
```

The development works pre-quotient — null events are allowed, per the strict-positivity point above — so the measure algebra is a connection one can note, not the working object.

(A convention check, since this is where drawings tend to differ: some would call $\mu$ *lax* rather than oplax — that is just the order-convention flip, taking morphisms $x \to y$ to mean $x \ge y$. With $A \to B$ meaning $A \subseteq B$ and $x \to y$ meaning $x \le y$, subadditivity comes out oplax as above.)

Zcash/Security/Ledger/OrchardIntegrityExperiment.lean
```diff
@@ -0,0 +1,406 @@
...
+Orchard ledger is valid and violates balance integrity at some prefix `i < k` is at most
+`ε_sinsemilladlr + (ε_dl + (qH+2)/#F)`. This is the abstract integrity experiment for an
+idealized adversary with the non-negativity side discharged at the Orchard instantiation:
+the three Balance-subset arms collapse onto the single advantage `ε_sinsemilladlr` of computing a
+nontrivial Sinsemilla discrete-log relation (`sampledBalanceSubsetArms_subset_orchardRelation`),
+replacing the abstract `ε_nonneg` with one term and no factor of three. The conservation side is
+the abstract experiment's combined coin-consuming finder, so the whole bound carries one
+discrete-log advantage per side: `ε_sinsemilladlr` among the fixed Sinsemilla bases, `ε_dl` at the
+sampled value-commitment bases. -/
+theorem orchardBalanceIntegrityBefore_measure_le_experiment_idealizedks (p : PMF ι)
```
TalDerei, line 155: possible census escape hatch? doubt our regex predicates cover `_idealized...`

Zcash/Security/BindingSignature/Orchard.lean
```diff
@@ -46,13 +46,33 @@ theorem orchard_natAbs_lt {r : ℕ} (vs : List ℤ) (vBalance : ℤ)
 theorem orchardVSumBound_lt_pallasScalarOrder : orchardVSumBound < (pallasScalarOrder : ℤ) := by
   norm_num [orchardVSumBound, vSumBound, pallasScalarOrder]
 
+/-- The ledger model's no-overflow bound at the Orchard limits: under the dedicated consensus
+action caps —`nActionsOrchard` and `nActionsIronwood` are each less than `2^16` (spec §7.1.2,
+<https://zips.z.cash/protocol/protocol.pdf#txnconsensus>), so the bound applies to either
+Orchard-protocol pool— the ledger layer's `vSum`-shaped per-transaction bound fits the
+Pallas scalar order. Derived from `orchardVSumBound_lt_pallasScalarOrder` by monotonicity
+of `vSumBound` in the action count: no numeric fact beyond the one anchor. -/
+theorem orchard_ledger_no_overflow {n : ℕ} (hn : n < 2^16) :
```
TalDerei, line 55: optionally link to [zip 229](https://github.com/zcash/zips/blob/ad30e59e63d7cb0f209567c6ae23ea1004ce739d/zips/zip-0229.md)

Zcash/Security/Ledger/OrchardIntegrityExperiment.lean
```diff
@@ -0,0 +1,406 @@
...
+    (hdl : ∀ j : A.ι, TextbookDLWithCoinsAdvantageLE pallasGen (A.conservationFinder k j)
+      ε_dl) :
+    A.experiment.toOuterMeasure (A.violationEvent issuance maxActions
+        (fun P => balanceIntegrityViolationBefore (P := P) (kv := keyBinding)
+          (issuance := issuance) (maxActions := maxActions) k))
+      ≤ ε_sinsemilladlr + (ε_dl + ((A.qH + 2 : ℕ) : ℝ≥0∞) / Fintype.card Fq) :=
+  orchardBalanceIntegrityBefore_measure_le_experiment_idealizedks spendAuthVerify
+    (redPallasBindingVerify H_bind) issuance maxActions 2 pallasGen 0 1 orchardQueryOf id
+    A.LA A.coins (by decide) A.queryBound A.algebraic (orchard_ledger_no_overflow hmax) k
+    hsin hdl
```
TalDerei, line 359: ACK `Pr[balance violation] ≤ ε_SinsemillaDLR + ε_DL + (qH + 2)/|Fq|` (implies either a sinsemilla relation break or a discrete-log break, with accounting for bad challenge set).

TalDerei (approved):
reviewed this more slowly than usual; the main balance violation reduction and joint probability accounting both look wired up correctly.

Zcash/Security/BindingSignature/Orchard.lean
daira: The v6 transaction format will be in the protocol spec (it's backlog from NU6.3) so no need to link to the ZIP. The consensus rule for Ironwood is already in the spec.

Zcash/Security/Ledger/OrchardIntegrityExperiment.lean
daira: Right — and the hatch was wider than the example: the `_experiment`-qualified endpoint names escaped both patterns too. All of them happened to be pinned voluntarily, so nothing was unpinned today, but the check would not have demanded a pin for a future endpoint in either family.

Addressed by 36ddcec4: both pattern sides (`ENDPOINT_RE` and its Lean port `isEndpointBaseName`) now allow the qualifier trailers `_for`, `_experiment`, `_idealizedks`, and `_experiment_idealizedks` after a semantic suffix, raising the demanded endpoint count from 154 to 161, all pinned.

daira (commented):
Self-reACK

PR #177, opened 2026-08-06
book: Explain the adversary models and split out the ledger games

daira:
This splits the former Security Definitions page and reorganizes the definitions:
* **Security Models** — methodology shared throughout the development;
* **Ledger Security Games** — the capstones picture and its reading guide;
* **Definitions** —the former Glossary— now collects every definition card, the ledger-side groups included, and sits before the Source Map in the contents.

Page filenames match their titles (`security-models.md`, `definitions.md`); all references, including the one Lean doc comment that cites the Definitions page, are updated. Old `security-definitions` and `glossary` URLs are not preserved. The contents order becomes Proof Journey, Proof Map, Security Models, Ledger Security Games, Definitions, Source Map.

The Security Models page then gains the content this split was for. It says what a reduction in an idealized model means: an advantage transfer (in our case, to an exhibited discrete-log finder, generic in the caller-supplied bound), not security conditional on a hardness premiss. The latter would be difficult to define due to the *existence* of attacks with massive precomputation; strictly following the breaks-as-computed-data approach sidesteps that issue. The page explains that the capstones' resource numbers are coverage parameters rather than attack-cost estimates: the work target, its three-bit finder envelope, and a statistical remainder proved at the full covered budget. It states the algebraic-adversary restriction, notes that it lives in the adversary's *type* rather than in named hypotheses, and gives the supporting claims the heuristic rests on. And it integrates the fixed-bases/reference-string section with the other half of the hash-to-curve story: the property that makes hash-to-curve outputs good bases also implies that a realistic adversary with direct access to `GroupHash` is not algebraic over any fixed finite basis. The games do not currently model this gap; that and the per-game enumeration of bases are stated as known limitations. (Spendability and Spend Authority need the hash-to-curve oracle in the adversary's interface; tracked as #188.)

In the ledger picture, reductions stated for algebraic adversaries are drawn as heavy purple edges —the model scopes both endpoint games of such a reduction, so it is an edge property, not a dependency node— and the random-oracle node remains a terminal because some error terms genuinely bottom out there as counting arguments. A new definitions card defines **generator-RO** (the programmable reference-string oracle and its naming convention), linked from the Security Models page.

The RedDSA discharge edge now names the straight-line AGM+ROM extraction (Fuchsbauer–Plouviez–Seurin 2019/877, Theorem 1) —the same extraction that discharges the binding-signature extractability arm— composed with the [FKMSSS2016] re-randomization reduction, replacing the forking route. Nothing in the re-randomization layer depends on the extraction mechanism; this was checked at full-text depth against FKMSSS 2015/395, ZEXE 2018/962, Das–Faust–Loss 2019/698, and Re-Randomized FROST 2024/436. The discharge edge and the extractor-justification bridge join the purple class, with the key extended to "a reduction (or intended reduction)" and a separate dashed-purple key entry for the bridge; the purple edges are styled by named edge IDs (mermaid ≥ 11.6 `classDef`/`class`) instead of `linkStyle` indices, so diagram edits no longer require recounting edge positions.

The Definitions page gets a polish: every anchor identifier links to its Lean source file (matching the diagram's convention; the pass caught one stale theorem name), the card math renders with KaTeX, the cards fit the page width with right-justified anchor lists that break between dotted names, a copy-editing pass tightens the prose, and new cards define **algebraic** and the **Uniform Reference String**. The key's "algebraic" links to the model section, the book prose is normalized to premiss/premisses (the book side of #187), and the convention blockquote on the Security Models page now quotes its source verbatim.

Citations on the new page are full-text-verified (Rogaway 2006/281; Corrigan-Gibbs–Kogan 2017/1113; Bernstein–Lange 2012/318; Koblitz–Menezes 2012/359; Shoup's Eurocrypt 1997 lower bounds; Fuchsbauer–Kiltz–Loss 2017/620; Canetti–Goldreich–Halevi 1998/011; Jaeger–Tessaro 2020/1213; Bernstein's *Understanding brute force*). The Canetti–Goldreich–Halevi citation previously read 2003/077, which is Bellare–Boldyreva–Palacio's uninstantiable-scheme paper, a successor result.

Planned follow-ups, coordinated separately: the capstone-card rework in the proof-map and journey data (so every card states its model and coverage), the bit-security/bibliography page for the Pollard-rho references behind the $2^{126}$ figure, and the hash-to-curve modelling issue above.

Rebased over the merged #178–#182 train. The definition cards and diagram edges follow #181's module moves. The proof-map embed and journey data are restored to main's content, with the premiss normalization re-applied; the journey's pinned historical paths stay as they are. The coverage-parameters section and the affected definition cards adopt the staged-certified endpoint story: $\mathrm{Adv}_{\mathrm{DLOG}}(2^{124}, 2^{126}) + 2^{-83}$ at the $2^{123}$ and $2^{125}$ budgets. A new **shallow cost language** card defines the accounting language's boundary, and the source map links it. The captured-endpoint card keeps this branch's corrected identifier — the pre-rename glossary cites a name that no longer exists.

It also carries the book changes from #185, ported across the renames, so that PR's rebase can drop its book hunks. The Formal Verification page gains the typed Rust↔Lean boundary paragraph, the pinned-not-regenerated circuit-fixtures caveat, and a Modelling Boundaries section on the Fiat–Shamir idealization. The Definitions cards for the straight-line extractor, the $x$-squeeze budget (its causal half now discharged via `pinnedX`), the capstones, κ, and the birthday bound are updated. The Security Models page notes that no Lean theorem instantiates the soundness endpoints at the deployed bases.

The Proof Map page gets an interaction and readability overhaul, driven by testing during review. The detail rail is click-to-pin instead of hover-driven: nothing flickers while the pointer roams, and the card can be read and scrolled at leisure. Selection emphasis is additive, and dimmed content stays readable. The legend is a compact full-width bar below a map-beside-card layout. Long node labels wrap to fit their boxes. The page uses the viewport's width and height instead of a fixed narrow frame. Book-wide, the chapter-nav strips narrow to a comfortable click target, inset symmetrically. The Proof Journey page's analogous issues are tracked separately (#190).

Full build green; the book renders with the new pages and anchors. The typos CI lint is removed — it kept flagging identifiers — with its configuration kept for local runs.

🤖 Claude Fable 5


book/src/formal-verification/proof-map-embed.html
```diff
@@ -503,7 +520,17 @@ <h2>Detail</h2>
...
 nodes.forEach(n => {
   n.cx = COL(n.c) + (n.dx || 0); n.cy = ROW(n.r);
-  const ch = n.label.length;
+  // Wrap long single-line labels onto two lines at the space nearest the
+  // middle, so the box (whose width is capped below) always fits its text.
+  if (!n.line1 && !n.w && !n.tag && n.label.length > 18 && n.label.includes(' ')) {
+    const mid = n.label.length / 2;
+    let best = -1;
+    for (let i = n.label.indexOf(' '); i !== -1; i = n.label.indexOf(' ', i + 1))
+      if (best === -1 || Math.abs(i - mid) < Math.abs(best - mid)) best = i;
```
daira:
```suggestion
    for (let i = n.label.indexOf(' '); i !== -1; i = n.label.indexOf(' ', i + 1)) {
      if (best === -1 || Math.abs(i - mid) < Math.abs(best - mid)) { best = i; }
    }
```

daira (commented):
Self-ACK

TalDerei (approved):
ACK

PR #178, opened 2026-08-06
feat(fv): split the Lean CI cache into layered dependency and build caches

### LINKED ISSUE #153: Improve CI performance
github's caching the entire .lake dir + mathlib (refs https://github.com/zcash/ironwood/blob/main/.github/workflows/lean.yml#L80-L100), but fresh builds still take 30+ min and warm builds still choke on cache thrashing / misses, so we should investigate import graph over-coupling and expensive kernel elaborations that slow it down further.

TalDerei:
Closes https://github.com/zcash/ironwood/issues/153. Cold starts are the same as before, but cold starts happen less often then before with more clever caching. One prudent way to reduce CI times is to bump our runners up from 4 cores (there are diminishing returns at some point, but going from 4 to 8 alone would more than halve CI times).

----------

**1. `lean.yml` caching rework** (`019a0b19`)

- Disabled lean-action's built-in cache (one 2.65 GB `.lake` blob per commit — only ~3 fit in the 10 GB quota, so concurrent PRs constantly evicted each other → ~35 min cold rebuilds).
- Split it into two explicit layers:
  - **Deps** (`.lake/packages`, ~2 GB): keyed on toolchain + manifest hashes only; one stable entry reused by all PRs between dependency bumps; saved only after a green build so a partial build can never become the shared entry.
  - **Project build** (`.lake/build`, ~0.7 GB): per-commit key with prefix restore of the newest entry; saved with `always()`, so failed and cancelled runs now warm the next attempt (previously saves were skipped on both — `actions/cache` saves via a post hook gated on `post-if: success()`, and `cancel-in-progress: true` cancels every superseded push).
  - Mathlib oleans stay out of GitHub's cache on bumps — `lake exe cache get` refills them in ~2 min.

**2. Fixture lanes decoupled from the soundness stack** (`51464e2e`)

The byte-locked captures must `import Zcash.Snark`, and the umbrella re-exported all of `Soundness/` — 24.9 min of the ~36 min cold build was this one serial chain. The umbrella stops re-exporting `Soundness/`; the captures' one real dependency (`VestaG` + its `Fp`-module structure) is hoisted into the new `Core/Vesta.lean` with all names unchanged. Coverage is pinned twice: an explicit `Zcash.Snark.Soundness.+` lakefile glob, and the new `scripts/check_build_coverage.sh` in CI asserting every module file is reachable from some default target. Verified locally with a full `lake build --wfail` (3,835 jobs); generated fixtures byte-untouched. Critical path: **24.9 → 15.5 min**.

**3. Rational-match tier trimmed from the umbrella** (`d5b73ced`)

The quantified-random-match tier (`SampleSpace`, `Rational/`, `Epsilon`) also leaves the umbrella: the captures reference none of it, its consumers already import it directly, and its ~4 min `Rational/Capstone` chain sat on every capture lane's build path. Fixture spine: **~15.5 → ~11.5 min**, leaving the Soundness spine (~14.9 min) as the binding critical path.

**4. `Mathlib.Tactic` umbrella swept** (`28e75190`)

All 83 bare `import Mathlib.Tactic` uses (measured at ~+1.6 s import-load and +1.3 GB RSS per Lean process vs. narrow imports) replaced with `Mathlib.Tactic.{Ring, NormNum, FieldSimp, Linarith}` plus modules for tactics each file demonstrably uses. `check_no_umbrella_imports.sh` now bans the bare `Mathlib.Tactic` alongside `import Mathlib` so it cannot creep back.

Commits 3 and 4 are guard-verified (build coverage, umbrella, csimp and endpoint censuses, zizmor) but await this PR's CI build for full elaboration — narrow-import fallout, if any, surfaces as unknown identifiers fixable by adding the named import.

### DIFF
--- a/.github/workflows/lean.yml
+++ b/.github/workflows/lean.yml
@@ -114,7 +114,10 @@ jobs:
       #    and manifest pins alone — not the commit. Between dependency bumps
       #    every PR and push reuses ONE stable entry, which stays LRU-fresh
       #    and is never re-uploaded (the save step is skipped on an exact
-      #    key hit).
+      #    key hit). Caveat: entries saved from a pull_request run are scoped
+      #    to that PR's merge ref, so after a bump each open PR cold-builds
+      #    its own entry until the first post-merge push to main publishes
+      #    the shared main-scoped one — merge bumps promptly.
       #  * Project build (`.lake/build`, ~0.7 GB compressed): keyed per
       #    commit and restored by key prefix from the most recent entry.
       #    Entries are small enough that the quota holds ~10 of them next to
@@ -188,7 +191,12 @@ jobs:
       #    cleanup, so partial progress from an interrupted build warms the
       #    next attempt. Cache entries are immutable, but this layer's keys
       #    are per-commit, so a partial entry is superseded as soon as the
-      #    next push builds further.
+      #    next push builds further. The restore step must itself have
+      #    COMPLETED, though: a cancellation landing mid-untar would
+      #    otherwise save the truncated tree, and since restore-keys resolve
+      #    to the most recently created prefix match, every subsequent run
+      #    would warm-start from the gutted entry and re-publish it under
+      #    its own key.
       #  * Dependencies: only if the build step actually succeeded. This
       #    layer's key is shared and stable until the next dependency bump,
       #    and the exact-key-hit guard (needed to avoid futile duplicate-key
@@ -199,7 +207,7 @@ jobs:
       #    the previous entry (restore-keys above) until its first green run
       #    publishes a complete one.
       - name: Save project build cache
-        if: always() && steps.restore-build.outputs.cache-hit != 'true'
+        if: always() && steps.restore-build.outcome == 'success' && steps.restore-build.outputs.cache-hit != 'true'
         uses: actions/cache/save@caa296126883cff596d87d8935842f9db880ef25 # v5.1.0
         with:
           path: .lake/build
--- a/.github/workflows/lean.yml
+++ b/.github/workflows/lean.yml
@@ -187,16 +187,31 @@ jobs:
       #
       # The two layers save under different conditions, deliberately:
       #
-      #  * Project build: `always()`, which also runs during cancellation
-      #    cleanup, so partial progress from an interrupted build warms the
-      #    next attempt. Cache entries are immutable, but this layer's keys
-      #    are per-commit, so a partial entry is superseded as soon as the
-      #    next push builds further. The restore step must itself have
-      #    COMPLETED, though: a cancellation landing mid-untar would
-      #    otherwise save the truncated tree, and since restore-keys resolve
-      #    to the most recently created prefix match, every subsequent run
-      #    would warm-start from the gutted entry and re-publish it under
-      #    its own key.
+      #  * Project build: saved whenever the build step FINISHED, green or
+      #    red. A red build's `lake` process exited on its own, so its writes
+      #    are whole, and letting a failed run warm the next attempt is the
+      #    big win for a PR under active iteration. Cache entries are
+      #    immutable, but this layer's keys are per-commit, so a partial
+      #    entry is superseded as soon as the next push builds further.
+      #
+      #    Cancelled runs deliberately do NOT save. A cancellation can kill
+      #    `lake` between an artifact write and its trace bookkeeping, and
+      #    whether the resulting tree is safe to restore then rests on lake's
+      #    write ordering: if an intact, input-matching trace can outlive a
+      #    truncated olean, lake skips rebuilding the corrupt file. Since
+      #    restore-keys resolve to the most recently created prefix match and
+      #    every run re-saves under its own key, one torn artifact would
+      #    propagate into every descendant entry until something loaded it
+      #    and failed — curable only by manually evicting the caches. The
+      #    partial progress given up is largely redundant with the prefix
+      #    restore of the last complete entry anyway. `always()` still leads
+      #    the expression because an `if` with no status-check function gets
+      #    an implicit `success()` prepended, which would stop failed runs
+      #    from saving at all.
+      #
+      #    The restore step must also have COMPLETED: a cancellation landing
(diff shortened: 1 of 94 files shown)


lakefile.toml
```diff
@@ -24,8 +24,18 @@ name = "mathlib"
...
+# captures must `import Zcash.Snark`, and re-exporting the soundness argument there put the
+# whole subtree on every capture lane's build path), the Soundness modules reach the root
+# closure only through the census imports in `Zcash/TrustBoundary.lean` — real but fragile
+# coverage, gone the moment a census entry is refactored. The explicit glob pins the whole
+# subtree into the default build regardless of what the import graph does;
+# `scripts/check_build_coverage.sh` (CI) asserts the same end-to-end property for every
+# module in the package.
 [[lean_lib]]
 name = "Zcash"
+globs = ["Zcash", "Zcash.Snark.Soundness.+"]
```
mitschabaude: why don't we just import those `Soundness` modules in `Zcash` now? otherwise the `Zcash` import file seems a bit meaningless to me

mitschabaude (approved):
ACK with the one comment above

daira (commented):
This is well designed and unusually well documented where it matters: the caching rationale lives in the workflow next to the steps it explains, the umbrella trims are explained in `Zcash/Snark.lean` itself, and the new coverage check states which Lake semantics it mirrors and why header-only import parsing is load-bearing. The layering directly addresses the measured eviction thrash (one ~2.6 GB per-commit blob → one stable ~2 GB dependency entry plus ~0.7 GB per-commit build entries), and the PR-scoped caveat on dependency entries saved from pull_request runs is stated rather than glossed.

The restructuring is properly guarded: coverage stopped being an import-reachability property when `Zcash/Snark.lean` stopped re-exporting `Soundness/`, and `scripts/check_build_coverage.sh` asserts the end-to-end property (every module reachable from some default target) against the lakefile's actual `defaultTargets` and globs, mirroring pinned Lake glob semantics. The phantom-edge rationale for stopping import scanning at the first real command is exactly right. The `check_no_umbrella_imports.sh` extension anchors `Mathlib(\.Tactic)?` on following whitespace/EOL, so `Mathlib.Tactic.Ring` cannot false-positive.

The body's caveat that commits 3 and 4 "await this PR's CI build for full elaboration" is now discharged: Lean CI, Fixtures CI, and the book CI are all green on the head.

One suggestion on the save conditions (line comment), and two notes that cannot anchor to changed lines:

* The fixture regenerate-and-diff lane was skipped on this PR, although the capture import surface changed (`Zcash/Snark.lean` umbrella, `Zcash/Snark/Core/Vesta.lean`, the lakefile globs). The local byte-diff covered it this time; consider adding those paths to the fixture-relevant filter in `.github/workflows/fixtures.yml` so the lane self-triggers on future restructures of the capture import path.

* `book/src/formal-verification.md` says CI "builds `Zcash` and `FixtureCheck` as default targets", which is now two targets short of the actual list. Out of this PR's scope; we can fold the one-line correction into #177 (which touches that page) unless you would rather take it here.

🤖 Claude Fable 5

.github/workflows/lean.yml
```diff
@@ -110,8 +171,53 @@ jobs:
...
+      #    layer's key is shared and stable until the next dependency bump,
+      #    and the exact-key-hit guard (needed to avoid futile duplicate-key
+      #    saves in the steady state) means whatever gets saved first under a
+      #    new key is FINAL for that key. An `always()` save here could pin
+      #    an interrupted, mostly-empty dependency build as the shared entry
+      #    for weeks; success-gating instead lets a bump PR warm-start from
+      #    the previous entry (restore-keys above) until its first green run
+      #    publishes a complete one.
+      - name: Save project build cache
+        if: always() && steps.restore-build.outcome == 'success' && steps.restore-build.outputs.cache-hit != 'true'
```
daira: On `if: always() && steps.restore-build.outcome == 'success' && ...`:

The restore-completed guard closes the mid-untar case, but `always()` still saves a tree from a lake process killed mid-write: a cancellation can land between an artifact write and its trace bookkeeping. Whether that tree is safe to restore then rests on lake's artifact/trace write ordering — if an intact, input-matching trace can coexist with a truncated olean, lake will skip rebuilding the corrupt file, and because restore-keys resolve to the newest prefix match and each run re-saves under its own key, the torn file propagates through every descendant entry until something loads it and fails, curable only by manual eviction. The comment's "stale artifacts ... are trace-checked and inert" holds for stale complete artifacts; torn ones are the open case.

Suggestion: add an outcome gate so cancelled runs do not save while failed runs still do. `always()` stays in the expression — an `if` with no status-check function gets an implicit `success()` prepended, which would stop failure runs from saving at all:

```suggestion
        if: always() && (steps.lake.outcome == 'success' || steps.lake.outcome == 'failure') && steps.restore-build.outcome == 'success' && steps.restore-build.outputs.cache-hit != 'true'
```

A failed build's lake process exited, so its writes are whole, and red builds warming the next attempt is the big iteration win; a cancelled run's partial progress is largely redundant with the prefix restore of the previous complete entry. Even if lake's write order turned out to handle torn artifacts correctly, the gate costs nothing and means we do not need to rely on that.

🤖 Claude Fable 5

lakefile.toml
TalDerei: well before, `Soundness` was only built because `Zcash/Snark.lean` re-exported it. but the fixture also imports `Zcash.Snark` (emitted by halo2), so the captures were blocked on the entire soundness stack building, which was like half the build time. removing the re-exports from `Zcash/Snark.lean` improves build times, but required stuffing them into the glob instead. it can technically be removed from the glob since other things like `TrustBoundary.lean` census imports build those modules, but it doesn't hurt and it's extra assurance. not sure why `Zcash` imports are "meaningless" because it still imports snark (albeit less modules without any soundness paths), common, circuits, security, trust boundary, etc.

.github/workflows/lean.yml
TalDerei: adding this CI guard.

PR #179, opened 2026-08-06
chore(fv): refresh proof map

TalDerei:
Updates proof map towards the current paradigm on main that only does straight-line AGM extraction and adaptive soundness rather than special soundness + non-adaptive soundness.


daira:
## Review of #179 — refresh proof map

**Overview.** This retires the special-soundness/forking narrative from all three navigation artifacts at once: the proof-journey data (now version 2, pinned to a single `main` SHA instead of six PR heads, with a nine-stage rewrite), the proof-map graph (forking/rewind/tree/recursive-extractor nodes replaced by the straight-line, oracle-model, and adaptive-statement architecture), and the validator (the retired 36-Vesta-capstone ledger check replaced by a bidirectional endpoint census plus a new edge-anchor existence check). A handful of Lean docstring corrections ride along, all docstring-only.

**Strengths:**

- **The new validator ratchets are well designed.** `validate_endpoints` is bidirectional — a retired endpoint left in the journey *and* a new endpoint the journey never mentions both fail — and each declared endpoint must additionally be anchored by some stage. `validate_map_edge_anchors` checks every formal edge's via-lemma against the *checked-out* Lean tree, while stage symbols check against the *pinned* ref; that asymmetry is what will make the #191 renames fail loudly rather than silently (see finding 1), and the docstring is explicit that it's a syntactic guard, not a direction-and-hypotheses review.
- **The content corrections are real claim-precision improvements**, consistent with the direction #194 set. The definitions entry now says the compressed constraint identity yields row-level semantics only with the separately priced exclusions and routing hypotheses. `SnarkRelation` is consistently re-described as a retained constraint-side payload rather than the Action capstones' consumer boundary — the map hint, node card, ConstraintRelations/KnowledgeSoundness docstrings, and the census comment all agree. And the old `hlrel`/`vkfaith` floors are replaced by `ledgerComposition`/`deploymentProvenance` cards that correctly credit `actionSpec_to_ledger` and the Lean-derived VK while naming exactly what remains.
- Collapsing six PR-head sources to one pinned `main` SHA removes the "the map describes a stack that no longer exists" failure mode the old version had, and the identifier-boundary regex got strictly better (`.'!?` excluded on both sides).
- **The new cards' specifics check out against current main.** The `slCapstone` card's `Adv_DLOG(2^126, 2^126) + 2^-84` matches `orchard_deployed_straightline_consensus_2pow123_generatorRO_finite_security`'s statement exactly, including its docstring's own three-bit-loss reading, and the `peel` card's census-tier note is precise: `separateOrRelationWitness` is pinned `+choice`, and the classify theorems carry `+native(CompElliptic.Curves.Pasta.Vesta.p_nsmul_Gpt)`.

**Findings:**

1. **The rebase after #191 is a four-surface sweep, and the pin must move with it.** The `endpoints` list, the stage theorem symbols, and the map node/edge anchors all carry the *old* capstone names that #191 renames; the pinned ref `c4b5976` predates both #191 and the already-merged #194. Once #179 rebases onto post-#191 main, `validate_map_edge_anchors` (working-tree scope) will fail on every old-name anchor — the guard works — but making it green requires, in one pass: bump the pinned ref, rename the endpoint list and stage symbols, update the map anchors, and add #194's sixth endpoint (the Python `ENDPOINT_RE` matches it, so `validate_endpoints` will demand it appear in the journey with a stage anchor). While in there, the oracle stage and `roModel` card still describe `PMFEventBiasLE` as "a stated premiss, not an equality" — true at the pinned SHA, but an underclaim on today's main, where #194's `Challenge255` layer prices the conversion exactly. Same class of semantic reconciliation as #191's; it should be one deliberate pass, not a conflict-marker sweep.
2. **The endpoint-pattern is now maintained in three languages** — the shell `ENDPOINT_RE`, #194's Lean `isEndpointBaseName`, and this PR's Python `ENDPOINT_RE` (deliberately narrower). Each pair is sync-by-comment. Same observation as on #194, now ×3: a shared fixture that all three must match would make the sync mechanical.

Overall: the refresh does exactly what the audit needs the navigation layer to do — describe the tree as it stands, with validators that make future drift loud — and the claim-precision corrections materially improve what an auditor will take away. The merge-order dependency on #191 is real: this PR's content is stale-at-birth unless its rebase includes the four-surface rename sweep and pin bump above.

TalDerei:
let's merge this since it's a faithful representation of the current construction. i'll add a review skill policy that flags map drift that needs correcting. we can also separately tackle sizing issues in https://github.com/zcash/ironwood/issues/190.

PR #180, opened 2026-08-06
chore(fv): reduce the capstone surface to knowledge soundness only

### LINKED ISSUE #174: Reduce the Action capstones to knowledge soundness only
Follow-up to https://github.com/zcash/ironwood/pull/165#discussion_r3710143413.

The Action surface carried four top-level capstones: two soundness and two knowledge soundness. The plain-soundness ones were only a simpler stepping stone while building up to extraction, and knowledge soundness tightly implies them, so they should go.

#165 already drops the plain-soundness Action endpoints, leaving `orchard_action_adaptive_bundle_knowledge_error_bound` and `orchard_action_adaptive_bundle_knowledge_finite_security`. Remaining work: land on exactly those two — one definitional (the compositional error formula), one concrete finite-security argument — and confirm nothing downstream still consumes a standalone soundness statement.

TalDerei:
Closes https://github.com/zcash/ironwood/issues/174. We had extraneous soundness capstones which are unnecessary because knowledge soundness provides strictly stronger guarantees and subsumes them. We deleted statements of the form "accept a bundle in which some Action is not satisfiable" in favor of strictly "accept a bundle from which we cannot produce the satisfying witnesses".


Zcash/Snark/Soundness/Action/AdaptiveStatementKnowledge.lean
```diff
@@ -215,6 +215,29 @@ def adaptiveStatementKnowledgeFailureEvent {pp : ProofParams}
...
 
+/-- Accepting a false bundle statement is a knowledge failure.  A returned witness entails
+`BundleStatement` by `ActionTerminal.ActionBundleWitness.statement`, so on a false statement the
+executable projection must have returned `none`.
+
+This is the step by which the advertised knowledge endpoints imply ordinary Action soundness at
+the same error: the accepting-false-statement set is contained in the event they already bound,
+so that weaker property is not stated separately.  It is a containment, not an endpoint -- no
+probability is claimed here. -/
+theorem acceptFalseStatement_subset_knowledgeFailure {pp : ProofParams}
```
daira, line 251: Since these two containments are the justification for deleting the soundness endpoints —the source-map now points at them as "the step by which the advertised knowledge endpoints imply ordinary Action soundness"— consider giving both an `assert_axioms` pin. They are supporting lemmas rather than endpoints, so the census families do not demand it, but the deletion's justification seems worth census-guarding at the standard tier so a regression cannot silently weaken it.

Nit, same docstrings: two ASCII `--` where the surrounding prose uses an em-dash.

🤖 Claude Fable 5

daira (commented):
Excellent, love to see this much red in a PR. The subsumption is proved, as it should be, rather than asserted (the two `acceptFalseStatement_subset_knowledgeFailure` containments for adaptive-statement and straight-line variants), with no hypotheses beyond the shared `hchar`. Their docstrings scope them correctly ("a containment, not an endpoint — no probability is claimed here").

One minor suggestion in a line comment: add those two containments to the trust census.

Deletion hygiene checks out: both dropped census pins correspond to deleted declarations, all `actionTerminalRelationFinder` occurrences in the diff are within the deletion hunks, no book page references the deleted names, the census script's guard-family comment is updated accurately (`*workFactor*` now names the knowledge capstone's `2^123` instantiation), and the source-map paragraph is rewritten.

Merge-order note, as agreed: #186 lands first, and this PR's rebase then deletes `AdaptiveStatementCapstone.lean` together with the compat plumbing #186 adds there.

🤖 Claude Fable 5

daira (approved):
ACK, but it is currently conflicted.

daira:
Re-checked at 4f37ae52 after the rebase over #186: the deletion takes the post-#186 file (the `terminalRelationFinderWithAcceptanceV` compat plumbing goes with it), the module-doc and source-map endpoint enumerations now match the staged-certified endpoint set, and none of the deleted declarations is referenced anywhere on the head. Both review items are addressed — `assert_axioms` pins on the two containment lemmas, and the em-dashes. Still good from our side.

🤖 Claude Fable 5

PR #182, opened 2026-08-06
chore(fv): census the compiled surface, not just the axiom footprint

TalDerei:
Pure hardening; makes the census check what the compiler actually runs, so nobody can slip in a fixture whose compiled value differs from the one the kernel proves things about.

### DIFF
--- a/Zcash/Arithmetic/FastMsm.lean
+++ b/Zcash/Arithmetic/FastMsm.lean
@@ -20,7 +20,9 @@ statement surface and the kernel-level meaning of `evalNat` are untouched. This 
 the proven-equality counterpart of `implemented_by`, which is forbidden here because it
 is unchecked — a ban the trusted-base census enforces rather than leaves to convention:
 `Zcash.Meta.assert_axioms` and `Zcash.Meta.assert_computable` fail the build on any
-`@[implemented_by]` or `@[extern]` outside the toolchain's own ambient surface.
+`@[implemented_by]` or `@[extern]` outside the toolchain's own ambient surface, anywhere in
+the import closure of the module the census entry sits in. `Zcash.TrustBoundary` records
+which modules that leaves uncovered.
 -/
 
 namespace Zcash.Arithmetic.Msm
--- a/Zcash/Arithmetic/FastMsm.lean
+++ b/Zcash/Arithmetic/FastMsm.lean
@@ -17,8 +17,10 @@ registered with `@[csimp]`:
 the compiler substitutes the fast implementation at every subsequently compiled call
 site — in particular inside the fixtures' `native_decide` auxiliaries — while the
 statement surface and the kernel-level meaning of `evalNat` are untouched. This is
-the proven-equality counterpart of `implemented_by` (which is forbidden here because
-it is unchecked).
+the proven-equality counterpart of `implemented_by`, which is forbidden here because it
+is unchecked — a ban the trusted-base census enforces rather than leaves to convention:
+`Zcash.Meta.assert_axioms` and `Zcash.Meta.assert_computable` fail the build on any
+`@[implemented_by]` or `@[extern]` outside the toolchain's own ambient surface.
 -/
 
 namespace Zcash.Arithmetic.Msm
--- a/Zcash/Arithmetic/FastMsm.lean
+++ b/Zcash/Arithmetic/FastMsm.lean
@@ -20,9 +20,10 @@ statement surface and the kernel-level meaning of `evalNat` are untouched. This 
 the proven-equality counterpart of `implemented_by`, which is forbidden here because it
 is unchecked — a ban the trusted-base census enforces rather than leaves to convention:
 `Zcash.Meta.assert_axioms` and `Zcash.Meta.assert_computable` fail the build on any
-`@[implemented_by]` or `@[extern]` outside the toolchain's own ambient surface, anywhere in
-the import closure of the module the census entry sits in. `Zcash.TrustBoundary` records
-which modules that leaves uncovered.
+`@[implemented_by]` or `@[extern]` outside the ambient roots
+(`Zcash.Meta.ambientPackageRoots` — the toolchain *and* the pinned dependency stack, not
+the toolchain alone), anywhere in the import closure of the module the census entry sits
+in. `Zcash.TrustBoundary` records which modules that leaves uncovered.
 -/
 
 namespace Zcash.Arithmetic.Msm
--- a/Zcash/Meta/AxiomCheck.lean
+++ b/Zcash/Meta/AxiomCheck.lean
@@ -1,6 +1,9 @@
 import Lean.Util.CollectAxioms
 import Lean.Elab.Command
 import Lean.DeclarationRange
+import Lean.Compiler.ImplementedByAttr
+import Lean.Compiler.ExternAttr
+import Lean.Compiler.CSimpAttr
 import Mathlib.Util.AssertNoSorry
 
 /-!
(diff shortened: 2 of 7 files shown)


Zcash/Meta/AxiomCheck.lean
```diff
@@ -44,6 +53,118 @@ namespace Zcash.Meta
...
+once; that is the intent — an override anywhere is undisclosed compiler trust that the census, as
+the artifact's statement of its trusted base, must not certify around. -/
+def checkCompiledBodyDisclosure (n : Ident) : CommandElabM Unit := do
+  let overrides := undisclosedCompiledBodyOverrides (← getEnv)
+  unless overrides.isEmpty do
+    throwError "{n} cannot be censused: {overridesText overrides} \
+      substitute(s) a compiled body Lean never checks against the kernel body, so the value the \
+      compiler runs is unconstrained by anything proved about it. Use the proven-equality \
+      `@[csimp]` instead, or — if the substitution really belongs in the trusted base — disclose it \
+      in `Zcash.TrustBoundary` and add it to `Zcash.Meta.allowedCompiledBodyOverrides`."
```
daira:
```suggestion
    throwError "{n} cannot be censused: {overridesText overrides} \
      substitute(s) a compiled body Lean never checks against the kernel body, so the value the \
      compiler runs is unconstrained by anything proved about it. If you must, use `@[csimp]` \
      instead, but be aware that `@[csimp]` still substantially increases the trust surface."
```
(and similarly for the pinned error message in `Zcash/Meta/Tests/CompiledOverride.lean`).

Zcash/Meta/AxiomCheck.lean
```diff
@@ -44,6 +53,118 @@ namespace Zcash.Meta
...
+  let mods := env.header.moduleNames
+  let mut found : Array (Name × Name) := #[]
+  for i in [0:mods.size] do
+    let idx : ModuleIdx := i
+    let impls := Lean.Compiler.implementedByAttr.ext.getModuleEntries env idx
+    let externs := Lean.externAttr.ext.getModuleEntries env idx
+    unless impls.isEmpty && externs.isEmpty do
+      unless ambientPackageRoots.contains mods[i]!.getRoot do
+        found := impls.foldl (fun acc (d, _) => acc.push (d, `implemented_by)) found
+        found := externs.foldl (fun acc (d, _) => acc.push (d, `extern)) found
```
daira, line 130: The skill file added in #183 says not to use `@[implemented_by]` or `@[extern]` at all; no exceptions. I believe that's the right policy, so I suggest dropping `allowedCompiledBodyOverrides` entirely: the def, the early return in `checkNoCompiledBodyOverride`, the filter here, and the "disclose it in `Zcash.TrustBoundary`" alternative in the two error messages.

(The ambient-root exemption is separate and stays: the toolchain and pinned dependency stack carry 1339 of these attributes in this closure, and we can't remove them.)

🤖 This comment was edited by Claude Fable 5 at Daira-Emma's request; the original version had an incorrect suggestion.

TalDerei: trying to just restrict all these compiler directives / attribute i can think of.

TalDerei: this isn't the right way to implement it, opting to just delete this allowlist.

Zcash/Meta/AxiomCheck.lean
```diff
@@ -160,6 +281,93 @@ def checkNativeAllowance (n : Ident) (axs : Array Name) (allowed : Option (Array
...
+`assert_computable` rejects a `partial` target already (it is not a `def`), but the entry that
+launders one is the `assert_axioms` on the certificate *consuming* it, whose own declaration is
+clean. So this is checked where it bites: inside what a `native_decide` certificate actually
+evaluated. That surface is exact and cheap, because a `native_decide` auxiliary axiom is
+`@Eq Bool e Bool.true` with `e` the closed term the compiler ran (`Lean.Meta.nativeEqTrue`) —
+walking `e` walks the data, not the proof.
+
+`e` is the term as the *kernel* records it, though, and the compiler does not run it verbatim: a
+`@[csimp]` lemma replaces a constant wholesale in compiled code, so where `e` says `evalNat` the
+compiled certificate ran `evalNatFast`. The walk therefore follows those replacements. Without that
```
daira, line 302: This is fine now, but let's remember to simplify it when we prohibit `@[csimp]`.

daira (approved):
ACK

Zcash/Arithmetic/FastMsm.lean
```diff
@@ -17,8 +17,10 @@ registered with `@[csimp]`:
 the compiler substitutes the fast implementation at every subsequently compiled call
 site — in particular inside the fixtures' `native_decide` auxiliaries — while the
 statement surface and the kernel-level meaning of `evalNat` are untouched. This is
-the proven-equality counterpart of `implemented_by` (which is forbidden here because
-it is unchecked).
+the proven-equality counterpart of `implemented_by`, which is forbidden here because it
+is unchecked — a ban the trusted-base census enforces rather than leaves to convention:
+`Zcash.Meta.assert_axioms` and `Zcash.Meta.assert_computable` fail the build on any
+`@[implemented_by]` or `@[extern]` outside the toolchain's own ambient surface.
```
daira: A precision point: the ambient roots include the pinned dependency stack (Mathlib and friends), not only the toolchain — perhaps "outside the ambient roots (`Zcash.Meta.ambientPackageRoots`)". (Not naming `allowedCompiledBodyOverrides` here becomes exactly right once the allowlist is dropped, per the comment at `undisclosedCompiledBodyOverrides`.)

🤖 Claude Fable 5

daira (commented):
[Claude's review, in addition to my comments.]

This closes the one census bypass that reaches a value, and the quarantined demonstration makes the threat concrete (`fingerprintMatch` by `native_decide` against `refutesAtKernel` by `decide`, both accepted). What I verified:

- Soundness of the closure-wide sweep: both attributes are `ParametricAttribute`s, whose `add` refuses imported declarations, so every override a commit can introduce lives in a module outside the ambient roots — and the scan reads the same per-module entry arrays the compiler reads, plus stage 2 for the current module. An override the compiler acts on cannot be invisible to it.
- The coverage story composes: the `partial` walk follows `@[csimp]` replacements exactly because csimp is the one allowed substitution; the disallowed ones are globally absent whenever the disclosure check passes; safe terms cannot reach `unsafe` directly; and a body-substituted `opaque` falls to the attribute sweep. The sealed-subtype barrier idiom stays green, and the test files pin both sides of each rejection.
- `CompElliptic` (with its vendored CompPoly) is outside `ambientPackageRoots`, so the dependency with the native lane sits inside the guarded region — consistent with its own csimp census and opt-in guards — and a new dependency fails the build until its root is classified, which makes adding one visible as the trusted-base change it is.
- The fail-every-entry semantics of the disclosure check is right for the stated reason: cone-scoping would let the commit that adds an override keep every unrelated entry green, and the error names the offender.
- The `check_csimp_census.sh` generalization is needed for exactly the case the comment gives (a test-only `@[csimp]` cannot be pinned from a file production must not import), and the `^assert_axioms` anchor keeps commented-out lines from counting.
- CI is green. This also turns the ban #183's review skill states (no `@[implemented_by]`/`@[extern]`) from convention into a build failure.

One doc-sync item, no hunk to anchor to: the book's census paragraph (`book/src/formal-verification.md:75`) still says csimp entries live in `Zcash/TrustBoundary.lean`; after the script change they may sit in any census file. While touching it, a sentence there on the new compiled-surface bound would complete the public story that the `Zcash.TrustBoundary` module doc now tells.

🤖 Claude Fable 5

Zcash/Meta/AxiomCheck.lean
TalDerei: added a comment to that effect.

daira (commented):
Post-hoc ACK

PR #183, opened 2026-08-06
feat(fv): review skill

TalDerei:
A pre-PR audit checklist mined from the review record of PRs #4-#173: census inclusion and reachability, `native_decide` necessity, `noncomputable`-on-reduction discipline, break-event fabricability, overclaiming prose, and merge-artifact churn.

### DIFF
--- a/.claude/skills/pr-preflight/SKILL.md
+++ b/.claude/skills/pr-preflight/SKILL.md
@@ -30,7 +30,7 @@ Check all six failure modes; each has occurred in a merged or nearly-merged PR.
    entries write the full path.
 4. **Endpoint names fit the census regex.** `scripts/check_endpoint_census.sh` only sees the
    listed families and suffixes. A new endpoint is renamed to fit the pattern — the regex is
-   not widened ad hoc.
+   not widened ad hoc. Refine the regex only when genuinely needed.
 5. **Rebases drop entries silently.** After any rebase or merge, read
    `git diff main...HEAD -- Zcash/TrustBoundary.lean` and account for every *deleted* line by
    name. A consolidation that loses another PR's pins passes CI and loses the guarantee.
@@ -64,8 +64,15 @@ justification in the PR. Native *execution* (evaluation-based checks in the lane
 is an explicit documented opt-in — never ambient.
 
 The same tier logic bans bespoke axioms and `implemented_by` outright: when the fact is proven
-upstream (CompElliptic, Mathlib), remove the axiom rather than pin it. `@[csimp]` with a
-kernel-checked equality is the permitted counterpart, and every `@[csimp]` is censused.
+upstream (CompElliptic, Mathlib), remove the axiom rather than pin it.
+
+There should be no uses of `@[extern]`, `@[implemented_by]`, or `unsafe`, and no
+new uses of `@[csimp]`. Despite the restrictions that this development puts on use of
+`@[csimp]` (checked in CI), it pulls in a
+[large trust surface](https://github.com/daira/CompElliptic/blob/main/design/lean-native-trust-research.md#appendix-a-extern-vs-implemented_by-vs-csimp),
+which is not worth the potential performance improvement. The equivalence proof it
+requires only a limited class of mistakes; it does not address this expansion of the
+trusted base.
 
 ## 3. `noncomputable` — props for specs, defs for reductions
 
@@ -83,7 +90,8 @@ For each `noncomputable` the diff adds, and for each new reduction:
 - Don't forget computed witnesses to `∃` or `Nonempty`: a reduction that computes an opening
   and then existentially closes it cannot be consumed as data by the next reduction, and
   retrofitting extraction-friendly statements is expensive. If a downstream layer consumes the
-  witness, return the data.
+  witness, return the data. `Nonempty` is basically equivalent to `∃` in introducing
+  non-extractability.
 - Each surviving exception is documented in the book, and un-marking one later obliges the
   census-tier upgrade and the book-paragraph deletion in the same PR.
 
@@ -129,7 +137,13 @@ For every touched docstring, module doc, book sentence, and the PR body:
 - No dev-history narration, no stale identifiers, no references to closed issues (open-issue
   references for tracked gaps stay — removing them is blocking). Every theorem, including
   trivial private lemmas, carries a doc comment. Constants cite the spec section and the
-  upstream Rust identifier; papers get author, year, and link.
+  upstream Rust identifier.
+- Cite papers that claims rely on, or that provide important context. Check that
+  it's the right paper and covers what is claimed. Ask the user to download it if
+  you can't. Use this citation format:
+  `(Author(s), linked title[, section/theorem][, venue year])`.
+  For example:
+  
 
 ## 6. Diff hygiene — the diff contains only its own changes
 
--- a/.claude/skills/pr-preflight/SKILL.md
+++ b/.claude/skills/pr-preflight/SKILL.md
@@ -1,0 +1,175 @@
+---
+name: pr-preflight
+description: Pre-PR audit of an ironwood branch against the violations reviewers flag most — census inclusion and reachability, native_decide necessity, noncomputable on reductions, break-event fabricability, overclaiming prose, and merge-artifact churn. Use before opening or updating a PR, when self-reviewing a branch, after a rebase or merge, when adding endpoints/native_decide/noncomputable/break structures, or when asked to "preflight", "pre-PR check", or "audit the branch".
+---
+
+# PR preflight (compiled from PR #4–#173 review feedback)
+
+Every rule here is a violation reviewers have flagged more than once. The skill is an audit
+procedure, not a style essay: run the checks against `git diff main...HEAD`, answer each
+question, and fix or justify before handing over. Prose register and docstring structure are
+owned by the `proof-comment-style` skill — this skill owns what the prose *claims* and what the
+Lean *trusts*.
+
+## 1. Census — every new trust-bearing declaration gets a direct pin that actually runs
+
+A pin that exists but never elaborates is the worst outcome: CI stays green over a `sorry`.
+Check all six failure modes; each has occurred in a merged or nearly-merged PR.
+
+1. **Direct pin exists.** Every new endpoint (the capstone naming families, or the semantic
+   suffixes `_error_bound`, `_finite_security`, `_measure_le`, `_probability_bound`,
+   `_capstone`) has its own `assert_axioms` / `assert_computable` entry in
+   `Zcash/TrustBoundary.lean` or a fixture-local trust-boundary module. Transitive coverage
+   through a dependent's pin is not coverage — it vanishes when the dependent is refactored.
+2. **The pinned module is reachable from a build root.** A census entry in a file no target
+   imports never runs. For each new `.lean` file, confirm it is in the import cone of
+   `Zcash.lean` or `FixtureCheck`; a file with no consumer and no target membership is checked
+   by nothing, however many assertions it contains.
+3. **Fully qualified names.** `open Zcash.Snark` does not bring `Deployed.foo` into scope as
+   bare `foo`; an unqualified census entry can silently resolve to the wrong homonym. Census
+   entries write the full path.
+4. **Endpoint names fit the census regex.** `scripts/check_endpoint_census.sh` only sees the
+   listed families and suffixes. A new endpoint is renamed to fit the pattern — the regex is
+   not widened ad hoc.
+5. **Rebases drop entries silently.** After any rebase or merge, read
+   `git diff main...HEAD -- Zcash/TrustBoundary.lean` and account for every *deleted* line by
+   name. A consolidation that loses another PR's pins passes CI and loses the guarantee.
+6. **Tightest true tier, minimal flags.** Try the census entry without `+choice` / `+native`;
+   keep flags only if the build demands them. Never `assert_no_sorry` — both census commands
+   imply it and we always want an axiom assertion. `#guard_msgs`-pinned `#print axioms` is
+   weaker than `assert_axioms +native(...)`, not stricter; don't describe or use it as the
+   stronger check outside the fixture censuses where the exact axiom set is the claim.
+
+Run `scripts/check_endpoint_census.sh` and `scripts/check_csimp_census.sh` before handing over.
+
+## 2. `native_decide` — data anchoring only, minimal, and trending to zero
+
+The repo's stated direction is removal ("the performance gain is not worth the extension of the
+trusted base"). For each `native_decide` the diff adds, in order:
+
+1. Does `decide` or `norm_num` close it? Then use that. (A reviewer caught `(-1 : Fp) ≠ 1`
+   proved by `native_decide`.)
+2. Does a compositional proof from existing lemmas exist? Then prove it. Prefer certifying one
+   general fact (an element's order, a generic congruence) over native-deciding an
+   instance-specific table.
+3. Is the statement minimal — one fact, no fused conjunctions, no re-running `assemble` for a
+   second claim that is cheap or already checked elsewhere?
+4. Is it anchoring *captured data* (a fixture capture check, `fingerprint_matches` tier — facts
+   with no other source)? That is the only category where a new native axiom is legitimate.
+   Correctness properties of objects the repo derives itself are proved by construction or
+   generically, not natively.
+
+Any survivor needs its census pin with the `+native(decl)` origin and a one-sentence
+justification in the PR. Native *execution* (evaluation-based checks in the lane import cone)
+is an explicit documented opt-in — never ambient.
+
+The same tier logic bans bespoke axioms and `implemented_by` outright: when the fact is proven
+upstream (CompElliptic, Mathlib), remove the axiom rather than pin it. `@[csimp]` with a
+kernel-checked equality is the permitted counterpart, and every `@[csimp]` is censused.
+
+## 3. `noncomputable` — props for specs, defs for reductions
+
+The discipline applies to *reduction producers*: anything that computes break data or feeds a
+`def`. Proof-side machinery (measure-valued, polynomial-valued, choice-selectors that appear
+only inside theorems) may be `noncomputable` freely.
+
+For each `noncomputable` the diff adds, and for each new reduction:
+
+- A reduction producer is a plain `def` pinned `assert_computable`. `Classical.choice` entering
+  only through erased `Prop` certificate fields is the `+choice` flag, not a reason to mark the
+  def noncomputable.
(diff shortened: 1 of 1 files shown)


.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+  zombies (a file restored into no build target, never elaborated). Self-review trust-boundary
+  files in particular.
+- New public declarations with zero consumers are wired, deleted, or explicitly kept as a named
+  result with a comment saying so.
+
+## 7. Interface and proof hygiene (quick checklist)
+
+- Interfaces keep abstract spellings (`actionCircuit.domainExponent`, not `11`); concretization
+  happens inside proofs or at the fixture boundary. No raised `maxHeartbeats` — seal the
+  concrete def (`irreducible` / opaque) instead.
```
mitschabaude: not really agreeing with this one, or at least there's a nuance:
* most code should be made generic over the circuit, so it can use `circuit.domainExponent`
* final capstone theorems, that aren't generic, should use 11, not `actionCircuit.domainExponent`. reason: 11 is the actual deployed zcash number that is trusted, while `actionCircuit.domainExponent` could be anything _a priori_.

mitschabaude (approved):
ACK with a minor comment, this is very useful

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+Check all six failure modes; each has occurred in a merged or nearly-merged PR.
+
+1. **Direct pin exists.** Every new endpoint (the capstone naming families, or the semantic
+   suffixes `_error_bound`, `_finite_security`, `_measure_le`, `_probability_bound`,
+   `_capstone`) has its own `assert_axioms` / `assert_computable` entry in
+   `Zcash/TrustBoundary.lean` or a fixture-local trust-boundary module. Transitive coverage
+   through a dependent's pin is not coverage — it vanishes when the dependent is refactored.
+2. **The pinned module is reachable from a build root.** A census entry in a file no target
+   imports never runs. For each new `.lean` file, confirm it is in the import cone of
+   `Zcash.lean` or `FixtureCheck`; a file with no consumer and no target membership is checked
```
daira: Once #178 lands (first in the agreed merge order), this item's manual audit is exactly what `scripts/check_build_coverage.sh` automates, and the coverage criterion changes shape: a module can be covered by a lakefile glob (`Zcash.Snark.Soundness.+`) without being in any import cone, and the root set is all five default targets, not `Zcash.lean`-or-`FixtureCheck`.

Suggest rewording this item to "reachable from some default target (import cone or lakefile glob) — run `scripts/check_build_coverage.sh`", and adding that script to the final sweep list.

🤖 Claude Fable 5

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+  goes through `scripts/regenerate-fingerprint-fixtures.sh` and the SHA-256 pins.
+- Dependencies pin full SHAs on canonical repos, never personal forks or mutable refs.
+- The PR description is part of the reviewed artifact: refresh it after force-pushes, and
+  "Closes #N" only with per-task accounting. Follow-up work gets a tracking issue, not only a
+  docstring. Confirm CI ran on the exact head that merges.
+
+## Final sweep
+
+Before handing over: `scripts/check_endpoint_census.sh`, `scripts/check_csimp_census.sh`,
+`scripts/check_no_umbrella_imports.sh`, `typos`, `lake build Zcash` — then one pass over the
```
daira: `lake build Zcash` elaborates less than the coverage property in §1 item 2 demands (FixtureCheck, CircuitCheck, MetaCheck, and SecurityCheck cones are not touched). If the narrowing is deliberate (for example because full fixture builds are slow), say so here; otherwise name the default-target build (`lake build --wfail`) as the sweep step.

🤖 Claude Fable 5

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+- Read `git diff main...HEAD` hunk by hunk. Unrelated comment rewording, blank-line churn, and
+  spelling regressions get reverted — prefer main on any comment the PR isn't about. An
+  incidental formatting change needs a stated reason.
+- After a rebase, check for resurrections (a declaration deleted on main reappearing) and merge
+  zombies (a file restored into no build target, never elaborated). Self-review trust-boundary
+  files in particular.
+- New public declarations with zero consumers are wired, deleted, or explicitly kept as a named
+  result with a comment saying so.
+
+## 7. Interface and proof hygiene (quick checklist)
```
daira, line 195: One repo rule that should be enforced for new/edited code is missing from this list: do not suppress the unused-section-variable lint file-wide (`set_option linter.unusedSectionVars false`); `omit` the unused instances instead. The reason is census-relevant, so it belongs in this skill rather than the style one: an unused instance argument is a subterm axiom collection traverses, so a concrete consumer passing an axiom-carrying instance propagates that axiom into its census footprint. `omit` instead removes the argument. There are currently four files with `set_option linter.unusedSectionVars false`, all under `Zcash/Security/Ledger/`; we will convert those in #171 (the just-stated argument was discovered after they were added).

🤖 Claude Fable 5

daira (commented):
This mines the review record accurately, and the selection is well-chosen: the fabrication test with its four real instances, the terminal-vs- intermediate break distinction from PR #50, the census failure modes (including rebase-dropped pins), the native_decide ladder, and the "prose may not outrun the proof term" section all match how those reviews actually went. The division of labour with `proof-comment-style` (register there, claims and trust here) is clean, and the trigger phrases in the description are the ones people actually use. The final sweep that runs `typos` locally remains valid with the CI lint removed in #177; it may now pick up typos from previously merged PRs, but any false-positives can be easily dismissed.

Three line comments below; all are small.

🤖 Claude Fable 5

.claude/skills/pr-preflight/SKILL.md
daira: Agreed. Abstract in the middle, concrete at both ends: interfaces and lemmas should be generic over the circuit, and final capstone statements should state the deployed literals. This makes the capstones easier to review without chasing another definition. They should be tied together by one explicit censused bridge equation such as `actionCircuit.domainExponent = 11`, so the literal cannot drift from the model and the correspondence has a single place to check.

🤖 Assisted by Claude Fable 5

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+2. **The pinned module is reachable from a build root.** A census entry in a file no target
+   imports never runs. For each new `.lean` file, confirm it is in the import cone of
+   `Zcash.lean` or `FixtureCheck`; a file with no consumer and no target membership is checked
+   by nothing, however many assertions it contains.
+3. **Fully qualified names.** `open Zcash.Snark` does not bring `Deployed.foo` into scope as
+   bare `foo`; an unqualified census entry can silently resolve to the wrong homonym. Census
+   entries write the full path.
+4. **Endpoint names fit the census regex.** `scripts/check_endpoint_census.sh` only sees the
+   listed families and suffixes. A new endpoint is renamed to fit the pattern — the regex is
+   not widened ad hoc.
```
daira:
```suggestion
   not widened ad hoc. Refine the regex only when genuinely needed.
```

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+5. **Rebases drop entries silently.** After any rebase or merge, read
+   `git diff main...HEAD -- Zcash/TrustBoundary.lean` and account for every *deleted* line by
+   name. A consolidation that loses another PR's pins passes CI and loses the guarantee.
+6. **Tightest true tier, minimal flags.** Try the census entry without `+choice` / `+native`;
+   keep flags only if the build demands them. Never `assert_no_sorry` — both census commands
```
daira: It isn't actually possible to make this mistake; redundant `+choice` or `+native` entries are always flagged as an error.

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+   imply it and we always want an axiom assertion. `#guard_msgs`-pinned `#print axioms` is
+   weaker than `assert_axioms +native(...)`, not stricter; don't describe or use it as the
+   stronger check outside the fixture censuses where the exact axiom set is the claim.
+
+Run `scripts/check_endpoint_census.sh` and `scripts/check_csimp_census.sh` before handing over.
+
+## 2. `native_decide` — data anchoring only, minimal, and trending to zero
```
daira: Removing `@[csimp]` is more important than removing `native_decide`. The former extends the trust surface much more. As we discussed on the call just now, it's not really necessary (or practical) for uses of `native_decide` to trend to zero &mdash; but uses of `@[csimp]` should.

I would frame this more generally in terms of avoiding unnecessary trust surface. There is a "[Trust Discipline](https://github.com/daira/CompElliptic#trust-discipline)" section in CompElliptic's README that we should copy to the Ironwood book and summarize here. We should also reference https://github.com/daira/CompElliptic/blob/main/design/lean-native-trust-research.md which has all of the detail about what `native_decide` and `@[csimp]` actually depend on.

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+   Correctness properties of objects the repo derives itself are proved by construction or
+   generically, not natively.
+
+Any survivor needs its census pin with the `+native(decl)` origin and a one-sentence
+justification in the PR. Native *execution* (evaluation-based checks in the lane import cone)
+is an explicit documented opt-in — never ambient.
+
+The same tier logic bans bespoke axioms and `implemented_by` outright: when the fact is proven
+upstream (CompElliptic, Mathlib), remove the axiom rather than pin it. `@[csimp]` with a
+kernel-checked equality is the permitted counterpart, and every `@[csimp]` is censused.
```
daira: I really think we should just eliminate `@[csimp]`; I'll file a PR to do that if @mitschabaude does not beat me to it.

```suggestion
upstream (CompElliptic, Mathlib), remove the axiom rather than pin it.

There should be no uses of `@[extern]`, `@[implemented_by]`, or `unsafe`, and no
new uses of `@[csimp]`. Despite the restrictions that this development puts on use of
`@[csimp]` (checked in CI), it pulls in a
[large trust surface](https://github.com/daira/CompElliptic/blob/main/design/lean-native-trust-research.md#appendix-a-extern-vs-implemented_by-vs-csimp),
which is not worth the potential performance improvement. The equivalence proof it
requires only a limited class of mistakes; it does not address this expansion of the
trusted base.
```

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+
+- A reduction producer is a plain `def` pinned `assert_computable`. `Classical.choice` entering
+  only through erased `Prop` certificate fields is the `+choice` flag, not a reason to mark the
+  def noncomputable.
+- A `noncomputable` marker is acceptable only when inert — the declaration appears inside
+  theorems and never feeds a def. If in doubt, drop the marker and let the compiler object.
+- Don't forget computed witnesses to `∃` or `Nonempty`: a reduction that computes an opening
+  and then existentially closes it cannot be consumed as data by the next reduction, and
+  retrofitting extraction-friendly statements is expensive. If a downstream layer consumes the
+  witness, return the data.
```
daira:
```suggestion
  witness, return the data. `Nonempty` is basically equivalent to `∃` in introducing
  non-extractability.
```

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -0,0 +1,175 @@
...
+  of X" claims a wiring; if the wiring is planned, write "to be discharged by X". State a
+  discharge as intended rather than done; qualify with the hypotheses and modelling gaps.
+- Security figures distinguish design target from achieved bound — never present the
+  accounting scale as a proven end-to-end bound.
+- Proof-map edges: node colour carries done/not-done; an edge label states what the reduction
+  *uses*. Don't draw composition that isn't formalized.
+- No dev-history narration, no stale identifiers, no references to closed issues (open-issue
+  references for tracked gaps stay — removing them is blocking). Every theorem, including
+  trivial private lemmas, carries a doc comment. Constants cite the spec section and the
+  upstream Rust identifier; papers get author, year, and link.
```
daira:
````suggestion
  upstream Rust identifier.
- Cite papers that claims rely on, or that provide important context. Check that
  it's the right paper and covers what is claimed. Ask the user to download it if
  you can't. Use this citation format:
  `(Author(s), linked title[, section/theorem][, venue year])`.
  For example:
  ```
  (Jaeger–Tessaro, <a href="https://eprint.iacr.org/2020/1213">Expected-Time Cryptography: Generic Techniques and Applications to Concrete Soundness</a>, Lemma 3)
  ```
  Link to full text if at all possible. Include the venue and year only if it is
  not an eprint and the version referenced is the one we want readers to look at;
  don't link to one version and then give the venue and year for another
  substantially different conference or journal version. Publication precedent
  doesn't matter for our purposes, pointing readers to a full, preferably
  open-access copy with all corrections does.
````

**Review by daira — APPROVED** [review]
Did an additional human review pass and added some suggestions.

File: .claude/skills/pr-preflight/SKILL.md [hunk 11]
```diff
@@ -0,0 +1,175 @@
...
+  result with a comment saying so.
+
+## 7. Interface and proof hygiene (quick checklist)
+
+- Interfaces keep abstract spellings (`actionCircuit.domainExponent`, not `11`); concretization
+  happens inside proofs or at the fixture boundary. No raised `maxHeartbeats` — seal the
+  concrete def (`irreducible` / opaque) instead.
+- `@[simp]` only where the RHS is a genuine normal form wanted everywhere; otherwise consumers
+  invoke the lemma explicitly.
+- No unearned wrappers or generality; prune a route the moment it is made redundant; check
```
daira: One sharpening from live experience: "legacy" as a name is how redundant routes survive their pruning deadline. I suggest extending this bullet: a superseded route is removed in the same PR, or kept only with a stated reason and a tracking issue for its removal. Also durable comments should not describe the new route by comparison against the superseded one ("unlike the legacy X ..."), which inverts the present-state rule in §5 and goes stale exactly when the pruning happens.

🤖 Claude Fable 5

TalDerei:
something else i'd like to append here is that every theorem should have a description.

.claude/skills/pr-preflight/SKILL.md
TalDerei: this "reachable from" wording is certainly more accurate.

.claude/skills/pr-preflight/SKILL.md
TalDerei: that's true this was overly conservative, removing.

.claude/skills/pr-preflight/SKILL.md
TalDerei: thought this entry also included csimp, appending.

TalDerei:
appending two more rules: every theorem should have a description, and no glob mathlib imports.

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -64,8 +64,15 @@ justification in the PR. Native *execution* (evaluation-based checks in the lane
...
-kernel-checked equality is the permitted counterpart, and every `@[csimp]` is censused.
+upstream (CompElliptic, Mathlib), remove the axiom rather than pin it.
+
+There should be no uses of `@[extern]`, `@[implemented_by]`, or `unsafe`, and no
+new uses of `@[csimp]`. Despite the restrictions that this development puts on use of
+`@[csimp]` (checked in CI), it pulls in a
+[large trust surface](https://github.com/daira/CompElliptic/blob/main/design/lean-native-trust-research.md#appendix-a-extern-vs-implemented_by-vs-csimp),
+which is not worth the potential performance improvement. The equivalence proof it
+requires only a limited class of mistakes; it does not address this expansion of the
+trusted base.
```
daira: Missing word in my applied suggestion.
```suggestion
requires prevents only a limited class of mistakes; it does not address this expansion
of the trusted base.
```

.claude/skills/pr-preflight/SKILL.md
```diff
@@ -129,7 +137,13 @@ For every touched docstring, module doc, book sentence, and the PR body:
 - No dev-history narration, no stale identifiers, no references to closed issues (open-issue
   references for tracked gaps stay — removing them is blocking). Every theorem, including
   trivial private lemmas, carries a doc comment. Constants cite the spec section and the
-  upstream Rust identifier; papers get author, year, and link.
+  upstream Rust identifier.
+- Cite papers that claims rely on, or that provide important context. Check that
+  it's the right paper and covers what is claimed. Ask the user to download it if
+  you can't. Use this citation format:
+  `(Author(s), linked title[, section/theorem][, venue year])`.
+  For example:
```
daira, line 173: The example and following paragraph are missing. I think GitHub must have misapplied the embedded triple-backticks (annoying because I got the Markdown quoting correct, and GitHub *displayed* it how I intended it to be applied).

Here's the intended text:

````markdown
  For example:
  ```
  (Jaeger–Tessaro, <a href="https://eprint.iacr.org/2020/1213">Expected-Time Cryptography: Generic Techniques and Applications to Concrete Soundness</a>, Lemma 3)
  ```
  Link to full text if at all possible. Include the venue and year only if it is
  not an eprint and the version referenced is the one we want readers to look at;
  don't link to one version and then give the venue and year for another
  substantially different conference or journal version. Publication precedent
  doesn't matter for our purposes, pointing readers to a full, preferably
  open-access copy with all corrections does.
````

File: .claude/skills/pr-preflight/SKILL.md [hunk 14]
```diff
@@ -0,0 +1,256 @@
...
+increasing the trust base.** CompElliptic's
+[Trust Discipline](https://github.com/daira/CompElliptic#trust-discipline) is the standard this
+development follows, and
+[lean-native-trust-research.md](https://github.com/daira/CompElliptic/blob/main/design/lean-native-trust-research.md)
+records what `native_decide` and `@[csimp]` actually depend on.
+
+Both extend the trusted base and neither is free. `@[csimp]` extends it further — its
+equivalence proof rules out only a limited class of mistakes, not the expansion itself — so
+removing an existing `@[csimp]` outranks removing a `native_decide`, and no *new* `@[csimp]` is
+added at all. But `native_decide` is held to the same direction of travel, not a lower bar:
```
daira, line 63:
```suggestion
to be added at all. But `native_decide` is held to the same direction of travel, not a lower bar:
```

daira (approved):
ACK, all of my comments are addressed.

PR #185, opened 2026-08-06
chore(fv): review fixes

TalDerei:
review fixes from local auditing passes covering specification parity, vacuous proofs, and breaks-as-computed-data violations.


TalDerei:
rebasing and resolving merge conflicts.

book/src/formal-verification/glossary.md
daira, line 1: This should be rebased on top of #177 (it may be sufficient to just drop these changes if #177 already includes them).

book/src/formal-verification.md
```diff
@@ -149,7 +149,36 @@ What the fixture captures actually *check* is the statement of record in each fa
...
+
+**Fiat–Shamir is idealized.** The deployed verifier derives each challenge by hashing the
+transcript so far with Blake2b and reducing 64 bytes to a field element. The development models
+that as an abstract `squeeze` (`Verifier/FiatShamir.lean`) and, in the security layer, as a
+uniform random oracle — the assumption that carries interactive soundness to the deployed
+non-interactive check. Identifying the deployed hash with that oracle is external, over and
+above the byte-level boundary noted above: the fixtures check challenge schedules against typed
+captures, not against transcript bytes.
 
 Coined terms and shorthand for the development, including the two conventions above, are
```
daira:
```suggestion
Coined terms and shorthand for the development are
```
This sentence is too far away from the conventions, and the reference to "the two conventions above" was always likely to go stale.

.github/workflows/lean.yml
```diff
@@ -152,28 +152,33 @@ jobs:
...
       # `configure`/`synthesize` against the CS and layout fixtures dumped from the Rust
       # circuit, so a divergence between the Lean circuit model and the deployed VK fails here.
+      #
+      # `MetaCheck` runs the census machinery's own negative regression suite, and
+      # `SecurityCheck` the ledger-bridge regression checks.
       - uses: leanprover/lean-action@38fbc41a8c28c4cbaec22d7f7de508ec2e7c0dd9 # v1.5.0
         id: lake
         with:
-          build-args: "--wfail"
+          build-args: "--wfail Zcash FixtureCheck CircuitCheck MetaCheck SecurityCheck"
```
daira, line 182: Pinning the list here closes one drift direction — the lakefile shrinking under CI — but opens the mirror one: a target added to `defaultTargets` that never lands here. A mechanical equality check (for example in `check_build_coverage.sh`, comparing this `build-args` list against `lakefile.toml`'s `defaultTargets`) would close both directions and let the keep-in-sync comment retire.

🤖 Claude Fable 5

daira (commented):
A substantive audit round; the two non-vacuity constructions are its core. What I verified:

- `capture_deployedAccepts` (both honest families) makes the satisfiability of `DeployedAccepts` — the hypothesis every soundness endpoint consumes — a build fact rather than an assumption, with exact census pins.
- The zero adaptive-statement family inhabits the capstones' adversary interface for every `ProofParams`, with the representation cap discharged concretely ($2 + 29 + 15 = 46$ entries against the $2^{89}$ limit, the 29 anchored by the new `fixedColumnCount_eq`). `adaptiveStatementInterface_nonempty` turns "the quantifier is not vacuous" into a censused theorem.
- The trust surface gets more honest rather than smaller: the fixture premiss list grows from seven to eleven (naming sampled-point distribution, ε-class membership, positional-frame stability, and container-merge exclusion), and the book gains the pinned-not-regenerated circuit-fixtures caveat — including `actionBaseLayout.json`'s no-anchor status — and the Fiat–Shamir modelling-boundary section. Both referenced PROVENANCE files exist, and the five→four capture count fixes stale text (there are four families).
- Census: the helper-leaf pins close exactly the transitive-coverage gap their comment describes, and the finder/extractor wrappers gain direct `assert_computable` pins. The endpoint-regex extension is right — the anchored suffix group keeps the conditional `_prob_le_of_*` spellings out, and `(_for)?` restores the obligation that generalizing over `numProofs` had silently retired.
- CI hardening: the capture-set check runs in both directions (canonical names must equal the pinned list; any exporter-header file must be in it) on every event via the cheap `--check-set` path (11 s in this run), and the weekly canary attributes pin rot to the pin rather than to the next unlucky PR.
- The x-squeeze schedule card's upgrade from named premiss to discharged is grounded — `pinnedX` is where the card says. And `Main.lean`'s rewritten semantic-reach comment matches the tree: the chain ends at `ActionBundleWitness` with `ActionSpec` satisfaction proofs, with the remaining floors stated. We will update the games page's bridge annotation to match, on our side.

## Merge planning

#177 now carries this PR's book changes, ported across its renames: the Formal Verification additions (with the heading respelled "Modelling"), the five card updates on the Definitions page, the deployed-bases scope sentence on Security Models, and the source-map wording, with premiss spellings normalized. So the rebase over #177 is trivial: drop the `glossary.md`, `security-definitions.md`, `formal-verification.md`, and `source-map.md` hunks, and the Lean side replays as is.

🤖 Claude Fable 5

book/src/formal-verification/glossary.md
TalDerei: rebased

PR #186, opened 2026-08-06
feat(fv): accounting for adaptive adversary work bounds

### LINKED ISSUE #172: Mechanically certify adaptive adversary work bounds
## Motivation

The valid claim remains:

> Any adversary genuinely limited to 2^123 work has success bounded by the 2^126 Vesta-DLOG advantage plus the statistical error.

What Lean does not currently verify is whether an arbitrary supplied adversary genuinely satisfies that work limit. That is an external modeling premise, standard in cryptographic reductions.

So it does not weaken the underlying security claim, provided the claim is stated conditionally. It only prevents saying that the entire concrete work bound is mechanically certified end to end.

PR #167 certifies random-oracle support and locality against the executable code, charges all three direct-decode executions, and derives the named assembly/basis group-work sub-budget. `proverGroupWork` and `reductionGroupWork` remain declared bounded resources. The latter covers complete finder postprocessing until every group-valued branch is connected to an operational counter.

In the current profile, the conditional claim therefore requires both the supplied adversary and the complete reduction postprocessing to satisfy their declared `2^123` bounds. A costed interface would establish those facts mechanically while preserving the existing conditional theorem.

## Resource-accounting presentation

After #167 lands, I would also like to do a pass presenting our resource accounting more cleanly and get feedback on both the presentation and the concrete parameters chosen:

- adversary computational-work limits;
- random-oracle query budgets;
- reduction overhead; and
- statistical soundness bounds.

There are separate random-oracle query and computational-work budgets. For instance, statistical soundness loss is essentially linear in the adversary's random-oracle query budget, so, when obtained by lowering the evaluated query cap, gaining `X` statistical bits costs about `X` query-budget bits. These are levers we can play with, and tradeoffs like this are made in the modeling.

One specific follow-up is the reduction overhead. #167 currently proves a conservative `8×` envelope, mapping the `2^123` adversary-work profile to a `2^126` DLOG-solver profile. The current accounting charges repeated traversals and then rounds up. Investigate whether caching and reusing those executions can soundly reduce the overhead to at most `2×`. This is a follow-up target, not a result currently proved by #167.

If that reduction is proved, a separate `2^125` adversary computational-work budget could remain within the `2^126` Vesta-DLOG group-work ceiling because `2 × 2^125 = 2^126`. The random-oracle query budget and its statistical-soundness consequence should remain independently stated rather than being raised automatically with the work budget.

I would like to tighten all of this up so the accounting and the doc comments explaining it clearly expose the separate budgets, parameter choices, tradeoffs, and conditional assumptions.

## Checklist

- [ ] Define a costed adversary/reduction DSL or monad.
- [ ] Reify Vesta group operations and their cost rules.
- [ ] Provide erasure/equivalence proofs to the current `LabeledOracleComp` adversary.
- [ ] Instrument the complete relation-finder and extractor postprocessing paths.
- [ ] Prove structural work bounds and connect them to both `proverGroupWork` and `reductionGroupWork`.
- [ ] Check that the costed language does not unintentionally restrict the adversary class.
- [ ] Investigate and, if sound, formalize a cached/reused-execution reduction with at most `2×` overhead.
- [ ] If the `2×` work reduction is proved, add a `2^125` adversary-work to `2^126` DLOG-work endpoint while keeping the RO-query budget separate.
- [ ] Re-state the `2^123` adversary to `2^126` DLOG endpoint using the certified bounds (closing #175).
- [ ] Update trust-boundary documentation and census pins.

- [ ] Rework the proof-map and journey cards for the adaptive-statement capstones, per the [#167 review](https://github.com/zcash/ironwood/pull/167#pullrequestreview-4870445975): each card should state what is guaranteed, for which adversaries, and at which coverage, in prose that does not require reading the proof first. Concretely — name the model (algebraic adversaries, challenge derivation modelled as a random oracle), which lives in the adversary's type and so is invisible to a summary written from the theorem statement alone; say a capstone proves a *bound*, not a "ceiling", with the quoted numbers named as coverage parameters; drop the internal anatomy ("six-way pre-IPA, k-way IPA, semantic, and source checks") in favour of meaning and placement in the development; retire undefined jargon such as the journey card's "explicit `PMFEventBiasLE` transport conjunct"; and stop implying the theorem takes DLOG hardness as a premise, since it is a reduction in the AGM. The register is book-wide rather than new to #167, so coordinate this with the resource-accounting presentation pass above.

TalDerei:
Closes #172.

## Summary

Adds executable group-work accounting for the adaptive adversary and its reduction.

- reifies Vesta group operations in a costed oracle DSL whose erasure is the original adversary;
- specializes one selected adversary path while retaining its annotations and group nodes, builds a proof-carrying cache from that result, and passes that same runtime cache into reified finder and extractor postprocessing;
- proves the composed counter is adversary work plus reduction work and proves the cached extractor has at most `2×` overhead;
- preserves the `2^123` adversary-work endpoint and proves a separate `2^125` adversary-work endpoint fits the `2^126` DLOG group-work ceiling;
- keeps the random-oracle budget independent at `Q ≤ 2^123`; and
- updates executable census pins, capstones, tests, and the proof map.

## Trust boundary and residual obligations

The syntax, value flow, group-operation counters, structural reduction bounds, and endpoint arithmetic are checked by Lean. The following premises remain explicit:

- the supplied adversary program must faithfully stage its Vesta group work;
- because the cost language is shallow, each complete execution must also faithfully stage Vesta work performed by generic host callbacks such as key construction, hashing, and fixed-representation production; and
- the three-decode bound relies on the family obligation `fixedRepresentations_length_le`; the generic theorem does not construct a concrete deployed-family representation table.

Equality checks, field operations, list scans, direct-coordinate decoding, and random-oracle queries are separate from the Vesta group-operation measure. DLOG advantage is the reduction conclusion, not a hardness proposition assumed by Lean.


book/src/formal-verification/source-map.md
```diff
@@ -219,8 +224,21 @@ root, IPA, and semantic surfaces under the single `(Q + 1)` query factor; `Prove
...
-its failure bound; and `Profile` prices the whole reduction at the `2^123` computational work
-target. The shared `AdaptiveSurfaces` and `AdaptiveTerminal` supply the per-commitment
+its failure bound; `Cached` proves that one retained execution gives the same finder and extraction
+event; and `Cost` constructs the programmed basis through reified Vesta operations, feeds those
+computed points into the adversary, threads reified canonical commitments into verifier assembly,
+and derives the conditional staged `2×` group-operation bound. Its private composition carries
+one closed program that constructs the charged basis, specializes the exact adversary path with
+its annotations and group nodes, constructs a proof-carrying cache from that result, and consumes
+that same cache in reified postprocessing. Lean proves the resulting counter equals adversary work
+plus reduction work. Since the cost language is shallow, fidelity of the supplied adversary and
```
daira: "Shallow" (or "shallow language") is not defined. It should be added to the glossary.

Zcash/Snark/Capstones/ActionBudgets.lean
```diff
@@ -1276,6 +1277,28 @@ theorem orchard_adaptiveActionStatementSurface_measure_le_for
...
+  simp only [assembleGroupOpsBudget, queryBudget, Zcash.Snark.Fixture.shape]
+  have hn' : numProofs ≤ 2 ^ 16 - 1 := by
+    simpa only [orchardConsensusMaxProofs] using hn
+  calc
+    _ = 30830 + 2050 * numProofs + 27 * (50 * numProofs + 46) ^ 2 := by
+      ring
+    _ ≤
+        30830 + 2050 * (2 ^ 16 - 1) + 27 * (50 * (2 ^ 16 - 1) + 46) ^ 2 := by
+      gcongr
+    _ ≤ 2 ^ 123 := by norm_num
```
daira, line 1306:
```suggestion
    _ ≤ 289909719069212 := by norm_num
    _ ≤ 2 ^ 123 := by norm_num
```
To save curious people from having to calculate it.

Zcash/Snark/Soundness/Action/AdaptiveStatementCached.lean
```diff
@@ -0,0 +1,539 @@
...
+  family.cachedRunOfExecution basis O ((family.adversary basis).runWithAnnotations O)
+
+/-- The non-provenance view of a shared execution. -/
+def CachedRun.toRunView {pp : ProofParams}
+    {family : ComputedAdaptiveActionStatementFSFamily pp}
+    {basis : AugmentedIndex (2 ^ (AdaptiveActionStatementShape pp).k) → VestaG}
+    (cache : CachedRun pp family basis) : RunView pp family basis :=
+  { output := cache.output, pre := cache.pre, rounds := cache.rounds }
+
+@[simp] theorem cachedRun_output_eq {pp : ProofParams}
```
daira, line 69: Add brief descriptions.

TalDerei: https://github.com/zcash/ironwood/pull/183#issuecomment-5220487675

daira (commented):
Fantastic improvement. The accounting is not a parallel model one has to trust: per-node costs are derived from syntax (an MSM's cost is its term-list length, not a caller-selected tag), and cost composition is proved. The staged route is anchored to the pre-existing reduction by pointwise equalities (the cached finder equals the original finder; the extractor succeeds on exactly the same oracle tables), with the certified variants carrying their defining equations so the counted operations are the ones producing the values consumed downstream. The trust surface is confined to the two staging-fidelity premises, both stated by name, and the endpoints re-export the certificate's obligations (erasure equality, `StagedGroupWorkFaithful`, the coverage properties) as conclusion conjuncts. That makes the model conditions visible in the statement itself rather than buried in a structure — exactly the visibility discipline the Security Models page argues for.

The endpoint arithmetic checks out against the book updates: the bound is `profile.advantage (2^124) (2^126) + 2^-83` with the oracle parameter separate at `Q ≤ 2^123`; reduction group work is bounded by `2^123`, so both adversary budgets fit the ceiling (`2^123 + 2^123 = 2^124` and `2^125 + 2^123 = 5·2^123 ≤ 2^126`), and the cached extractor's `≤ 2×` overhead is a stated conjunct. The diff adds zero `native_decide`s; the three costed producers get `assert_computable +choice +native` in the fixture-local census and the three certified endpoints `assert_axioms +native` — the right tiers. The book, journey, and embed updates match the Lean statements, and their framing (coverage parameters; the DLOG advantage as the reduction's conclusion, not a premiss) is exactly right.

`Meta/Tests/CostedOracle.lean` deserves a callout: pinning the *absence* of the forgery APIs (`ofLabeled`, public `group`, the core `mk`) as `#guard_msgs` unknown-constant checks turns the module's negative API surface into compile-failure regressions. One suggestion below states the convention behind it outright in the module doc.

Merge-order note, as agreed: this lands before #180, whose deletion of `AdaptiveStatementCapstone.lean` then removes the compat plumbing this PR adds there.

🤖 Claude Fable 5

Zcash/Snark/Soundness/AGM/CostedOracle.lean
```diff
@@ -0,0 +1,578 @@
...
+small, erasable accounting language for the other finite resource used by the concrete AGM
+endpoints: Vesta group work.  A group node is an explicit audit event; erasure removes those
+events and leaves exactly the original labeled oracle tree.
+
+The cost rules count additions, negations, scalar multiplications, and MSM terms.  This matches the
+convention used by `AssembleGroupCost`: a primitive MSM costs one modeled group-work unit per term,
+and its fixed-generator sweep and accumulated terms are charged explicitly.  Equality tests,
+field operations, and list traversal are kept outside the DLOG group-operation budget.  In
+particular, an adversary-sized scan must not be disguised as a shape-only group-work charge.
+
```
daira, line 16: The `Meta/Tests` tombstones catch re-introduction of the canonical forgery APIs by name. I'm a bit skeptical that this is useful: an *accidental* promotion under another name, or a lift from a different `OracleComp` flavour, would escape them. For a potentially more robust approach, see the comment on `StagedGroupWorkFaithful` below, but that's not a blocker for this PR and I don't object to the lightweight check. In any case I think it would be useful to add a sentence here stating the convention outright:

```suggestion

Any new producer of `CostedLabeledOracleComp` / `CostedVestaComp` must route
through the smart constructors. Adding a lift from any uncosted computation
type would allow bypassing the intended costing and would be a correctness bug.

```

🤖 Claude Fable 5

Zcash/Snark/Capstones/Action.lean
```diff
@@ -102,6 +102,74 @@ theorem orchard_action_knowledgeFailure_prob_le_adaptiveStatement_for
           (adaptiveStatement_pairCount_lt numProofs family) B epsilon
             profile.finderAdvantageLE hsurface)
 
+/-- Consensus-generic adaptive-statement knowledge soundness with conditional staged work
+accounting. The adversary program erases to the original game. Its staged group work and the
+cached reduction's group-law work are derived from syntax and the captured verifier shape.
+External implementation-to-program fidelity remains explicit; “certified” here therefore
+describes the checked group-operation accounting, not an assumption-free closed theorem. -/
```
daira: Claude and I both initially parsed "checked group-operation accounting" as automatic cost extraction from the reduction terms. I suggest saying which half is checked: node costs derived from syntax, proved cost composition, and pointwise equality of the staged route to the original finder. Expressing the computation through the reified nodes is manual, with its fidelity as the named premiss.

🤖 Claude Fable 5

Zcash/Snark/Soundness/Action/AdaptiveStatementCost.lean
```diff
@@ -0,0 +1,2195 @@
...
+        hchar certificate basis O).program.StagedGroupWorkFaithful) ∧
+    ∀ (C : VestaG)
+      (x y : AugmentedIndex (2 ^ (AdaptiveActionStatementShape pp).k) → Fp) O,
+      (family.costedProgrammedCachedRelationFinder
+        hchar certificate B C x y O).program.StagedGroupWorkFaithful ∧
+      (family.costedProgrammedCachedKnowledgeExtractor
+        hchar certificate B C x y O).program.StagedGroupWorkFaithful
+
+/-- DLOG advantage interface for the conditionally certified one-execution finder. Unlike the
+legacy profile, there are no caller-selected prover or reduction group-work numbers: those come
```
daira: Two things here. Describing the new route by comparison against a superseded one goes stale exactly when the pruning happens — prefer stating the present behaviour in its own terms (the caller-selected numbers point can be made without the comparison). And the "legacy profile" itself: if it is redundant once this lands, its pruning belongs in #180's sweep or in a tracking issue; "legacy" as a name is how redundant routes outlive their deadline. (Same suggestion made on #183 for the skill's prune bullet.)

🤖 Claude Fable 5

Zcash/Snark/Soundness/AGM/CostedOracle.lean
```diff
@@ -0,0 +1,578 @@
...
+def GroupWorkBound (A : CostedLabeledOracleComp T F Label α) (bound : Nat) : Prop :=
+  ∀ O, A.groupWork O ≤ bound
+
+/-- External staging judgment for the shallow cost language.  It means that construction of
+`pure` payloads, query arguments, and host-language continuations performs no Vesta group-law
+work except through the reified nodes of `A`.  Lean has no operational cost semantics for an
+arbitrary host term, so this judgment is intentionally abstract: a concrete implementation or a
+deeper embedded program must discharge it.  Keeping it in the certificate prevents a zero-cost
+`pure` tree from being silently presented as an end-to-end mechanical work proof. -/
+opaque StagedGroupWorkFaithful (A : CostedLabeledOracleComp T F Label α) : Prop
```
daira, line 129: A design question: the opaque judgment is honest, but it makes even the in-repo, combinator-built reduction programs conditional on an in-Lean-undischargeable premiss. Would a parametric variant of the staged language —programs written against an abstract group interface, so host code cannot perform group law outside the reified nodes *by parametricity*— let combinator-built programs carry staging fidelity as a theorem, shrinking the opaque judgment to the boundary where concrete `VestaG` enters (the supplied adversary program, host callbacks)? Possibly expensive to retrofit; if it is, a tracking issue for the idea would still be good.

🤖 Claude Fable 5

Zcash/Snark/Soundness/Action/AdaptiveStatementCost.lean
```diff
@@ -0,0 +1,2195 @@
+import Zcash.Snark.Soundness.AGM.CostedOracle
+import Zcash.Snark.Soundness.Action.AdaptiveStatementCached
+import Zcash.Snark.Soundness.Action.AdaptiveStatementReads
+import Zcash.Snark.Soundness.Composition.AssembleGroupCost
+
+/-!
+# Certified adaptive-statement work accounting
```
daira, line 7: Reader routing: this module (and `AdaptiveStatementCached`, `ActionBudgets`, the capstone section) should point at `AGM/CostedOracle.lean` as the prerequisite — the cost language's semantics, erasure, and the staging judgment live there, and a reader starting from the endpoints currently has no pointer back to it.

🤖 Claude Fable 5

Zcash/Snark/Soundness/AGM/CostedOracle.lean
TalDerei: don't think this actually is necessary (defining over concrete `VestaG` seems fine, rather than parameterizing over an abstract G). our program definitionally isn't polymorphic, so this is a moot point.

daira:
The bare `#check`s in `Meta/Tests/CostedOracle.lean` (the positive existence assertions) print their full signatures as `info` diagnostics into every build log. `example := @name` asserts existence just as hard, silently — and a `#guard_msgs` signature pin would be fragile against pretty-printer drift, so the silent form seems right.

🤖 Claude Fable 5

PR #191, opened 2026-08-08
feat(fv): contract interface

### LINKED ISSUE #148: Expose the knowledge-soundness contract as an auditable interface
The knowledge-soundness *claim* is now centralized; the *contract* behind it is not. A human auditor can find the theorem, but reconstructing what it actually promises still means reading across the implementation layers that prove it.

## What changed since this was filed

#165 built `Zcash/Snark/Capstones/` and cut the advertised Action surface from twelve endpoints to five, all pinned in `Fixtures/MultiAction/Honest/TrustBoundary.lean`:

- `orchard_action_knowledgeFailure_prob_le_adaptiveStatement_for` — the consensus-generic compositional error formula
- `orchard_action_knowledgeFailure_prob_le_adaptiveStatement_certified_for` — the same with staged work accounting
- `orchard_action_knowledgeFailure_adaptiveStatement_2pow123_workFactor_generatorRO_for`
- `orchard_action_knowledgeFailure_adaptiveStatement_certified_2pow123_work_generatorRO_for`
- `orchard_action_knowledgeFailure_adaptiveStatement_certified_2pow125_work_generatorRO_for`

That closes the original "prune redundant or unreachable APIs" and "mark secondary endpoints as internal" bullets — the sequential and fixed-statement endpoints were deleted rather than demoted. `source-map.md` documents the layout.

What it did not do is split along the axis this issue asked for. `Capstones/` factors by build order — `ActionEvents → ActionChecks → ActionBudgets → Action` — not by the auditor's reading order. `ActionEvents.lean` is a misnomer worth fixing on its own: it defines no events, and says so, deferring them to `Soundness/Action/`.

## The remaining gap

Reading the contract behind the primary endpoint means opening eight modules across four trees:

| Contract element | Where it lives today |
| --- | --- |
| Claim | `Snark/Capstones/Action.lean` |
| Adversary model | `Snark/Soundness/Action/AdaptiveStatementModel.lean` — `ComputedAdaptiveActionStatementFSFamily` |
| Failure event | `Snark/Soundness/Action/AdaptiveStatementKnowledge.lean` — `adaptiveStatementKnowledgeFailureEvent` |
| Extractor | same module — `adaptiveStatementKnowledgeExtractor` |
| Witness | `Snark/Soundness/Action/StraightLineTerminal.lean` → `Circuits/Integration/TopLevelCorrectness.lean` |
| Assumptions | `Snark/Soundness/Action/AdaptiveStatementProfile.lean` — `AdaptiveStatementDlogProfile`; plus bare theorem hypotheses |
| Bound | inline literals in the endpoint + `Snark/Soundness/Composition/AlgebraicRootBudget.lean` |
| Ledger bridge | `Security/Ledger/Bridge.lean` — `actionSpec_to_ledger` |

`Soundness/Action/` is nineteen modules to navigate around those definitions.

## Design decision to settle first

The original proposal put the interface at `Zcash/Snark/Soundness/Action/`, which is circuit-specific by construction. That drew the objection that an auditor-facing layer should not be tied to a single circuit, and I think that is right on the evidence: `Security/Ledger/OrchardCapstone.lean` already runs a second capstone family with the same shape — a break event, a `_subset_relation` containment proving the event covers the property, and a `_measure_le` bound on it. `acceptFalseStatement_subset_knowledgeFailure` is the Action-side instance of exactly that pattern.

Proposed resolution: define the interface generically — a record of claim, event, witness, extractor, assumptions, and bound — and instantiate it for Action first. Action is the only circuit carrying an advertised capstone today; `CommitIvk`, `NoteCommit`, `Ecc`, `Sinsemilla`, and `Poseidon` are components composed into its spec rather than independent surfaces. So Action is the only instance in the near term, but the abstraction is what keeps the second one from being another bespoke tree. No such abstraction exists in the repo yet.

Naming the modules for the contract rather than the circuit follows from the same decision.

## Scope

1. **Interface.** Define the generic contract record and instantiate it at Action, re-exporting the existing witness, event, extractor, assumptions, and primary capstone. No proof rewrites, no bulk file moves. Changes to existing modules limited to imports/exports, the source map, and census entries.
2. **Prose.** The corresponding book chapter, stating the Action contract as claim → event → witness → assumptions → bound. This is the half that decides whether the issue is actually done, and it overlaps #47 — the two cannot be closed independently.
3. **Cleanup, later.** Once the interface and its documentation are stable: simplify imports, retire what the interface makes redundant, rename `ActionEvents.lean`, and move internals only where that clarifies the boundary.

## Non-goals

- Changing the security model.
- Starting with mass file moves.
- A re-export layer of the shape #165 deliberately removed. That PR deleted three compatibility shims nothing imported and the `Soundness/FiatShamir.lean` umbrella; nothing mechanically blocks re-adding that shape, since `check_no_umbrella_imports.sh` only guards Mathlib, but it should not be re-added by accident. The interface must be a stated contract, not a forwarding header.

## Done when

A reader can state, from the interface modules and the book chapter alone: what witness the extractor returns, what the failure event is, every assumption the bound is conditional on, the concrete bound, and where the circuit-to-ledger continuation picks up.

## Related

- #47 — the human-readable guide to the theorems. Owns the prose half above.
- #147 — the circuit-to-ledger Action witness bridge, which the contract's last element points at. Its description is written against `Bridge.specPost_to_ledger`, a name no longer in the tree; the current spelling is `Bridge.actionSpec_to_ledger`.
- #165 — landed the `Capstones/` layer and the endpoint reduction described above.

The original text also asked that #48 audit the resulting contract. That issue is closed, so the adversarial pass needs either a new ticket or folding into this one's review.

TalDerei:
Closes https://github.com/zcash/ironwood/issues/148.

This makes the Action knowledge-soundness contract readable in one place, instead of reconstructible only by traversing the five modules that prove it. Specifically, it adds an auditor-facing contract interface, instantiated as `actionKnowledgeContract` for the deployed Action circuit, that gathers the extractor, witness guarantee, failure event, assumptions, and concrete error bound in one reviewable place that a reader can consume. Previously a reader had to dig through the capstone call stack to find these individually.


Zcash/Snark/Contract/Action.lean
```diff
@@ -0,0 +1,92 @@
...
+open Zcash.Snark.ActionTerminal
+open Zcash.Snark.Capstone
+open Zcash.Snark.Keygen (actionProofParamsFor)
+open Zcash.Circuits Zcash.Circuits.Action
+
+/-- The deployed Action circuit's knowledge-soundness contract at one consensus-valid bundle size.
+
+Every field is the object the proving layer already defines; the record only puts them beside each
+other. `knowledge_sound` is the advertised endpoint, applied unchanged. -/
+noncomputable def actionKnowledgeContract (numProofs : ℕ) {T : Type*} [DecidableEq T]
```
TalDerei, line 48: `noncomputable` def because of the internal fields (PMF and error).
- **inputs** (setting, assumptions, adversary, hardness)
- one **output** (`KnowledgeContract`) whose components are the real objects: acceptance, extractor, witness type, event, bound. the last field `knowledge_sound` is the endpoint theorem itself slotted in, and it typechecks only if every other field is consistent with it.

book/src/formal-verification/knowledge-contract.md
```diff
@@ -0,0 +1,109 @@
...
+
+Accepted, yet extraction returned nothing. Both conjuncts matter: a rejected run is not a
+failure, and neither is an accepted run that yielded a witness.
+
+### 6. What is the error?
+
+The endpoint's compositional formula: the adversary's discrete-log advantage at its query and
+group-work counts, plus $1/|\mathbb{F}|$, plus a per-query term collecting the Schwartz–Zippel
+budgets of each challenge surface. At the $2^{123}$ work-factor target it lands on
+$\mathrm{Adv}_{\mathrm{DLOG}}(2^{126}, 2^{126}) + 2^{-83}$.
```
daira: Define the arguments of $\mathrm{Adv_{DLOG}}$ somewhere and link from here to there (similarly for other uses of $\mathrm{Adv_{DLOG}}$).

book/src/formal-verification/knowledge-contract.md
```diff
@@ -0,0 +1,109 @@
...
+
+## What the contract does not say
+
+**Completeness.** That some run accepts, or that an honest prover's proof extracts, is a
+separate property; a contract whose acceptance predicate holds nowhere satisfies every field.
```
daira: This can be misread with the opposite sense if you don't notice it is supposed to be read in the context of the heading.

book/src/formal-verification/knowledge-contract.md
```diff
@@ -0,0 +1,109 @@
...
+Read the contract as a bound on the adversary, never as evidence that the circuit works.
+
+**Ordinary soundness — because it is free.** On a false statement the extractor must have
+returned `none`, since a returned witness would have entailed it. So `acceptFalseStatement_le`
+gives the soundness bound at the same error for every contract, and no separate endpoint is
+advertised for it.
+
+**The ledger.** The contract ends at `ActionBundleWitness`. The formal continuation begins, per
```
daira: "What the contract does not say: the ledger." is nonsensical. It should refer to the ledger security capstones.

book/src/formal-verification/knowledge-contract.md
```diff
@@ -0,0 +1,109 @@
...
+Action, at `Zcash.Security.Ledger.Bridge.actionSpec_to_ledger`, which consumes the public input,
+private witness, and `ActionSpec` proof — and returns an `ActionBreak` or an existentially
+witnessed ledger statement in `Prop`, not an executable ledger witness. The circuit-level result
+is therefore knowledge soundness while the ledger-level consequence remains ordinary soundness.
```
daira: The last sentence does not make sense. The ledger-level proofs need knowledge soundness.

daira:
## Review of #191 — contract interface

**Overview.** Three parts: (1) the `Capstones/` reorganization — `ActionEvents`/`ActionChecks`/`ActionBudgets` become `Action/{Base,Checks,Budgets}` with `Action.lean` as the only endpoint file, plus the endpoint renames onto the `orchard_action_<setting>_<qualifiers>_<property>_<form>` scheme; (2) the new `Contract/` layer — the generic `KnowledgeContract` record and its `actionKnowledgeContract` instance; (3) the book page and navigation updates. The contract design is genuinely good; the load-bearing review item is the rebase against #194.

**The contract design (checked, and strong):**

- `witness_statement` is exactly the right substantive field — without it the record is satisfiable by a junk extractor, and both the module doc and book page say so explicitly. The derived `acceptFalseStatement_le` correctly recovers ordinary soundness, and the census comment's decision *not* to pin it (so as not to re-advertise a property the Action surface deliberately stopped advertising) is careful census hygiene.
- The layer provably adds nothing, by construction: `knowledgeError` transcribes the endpoint's formula and `knowledge_sound` applies the endpoint unchanged, so any drift between the transcription and the real formula fails to typecheck — the module docs' "applied unchanged" claim is enforced by elaboration, not convention.
- `failure_eq := rfl` means the instance's named event is *definitionally* the acceptance-with-failed-extraction set — the strongest possible form of that identification.
- The book page is accurate against the code as far as I can check, and the statements in "what the contract does not say" are framing that an auditor needs. Note my manually posted comments, though: some of the wording in that section is unclear.

**Findings:**

1. **The #194 rebase carries semantic conflicts, not just textual ones — a mechanically resolved rebase can build green and still leave the endpoint surface internally inconsistent.** #194 added a *sixth* endpoint to `Capstones/Action.lean` — `orchard_action_deployedKnowledgeFailure_2pow123_workFactor_prob_le` — under the *old* naming convention, plus its census entry and the model-boundary docstring. Reconciling the two requires deliberate decisions, not merge-tool output: the new endpoint renamed into the scheme (or deliberately exempted, with the naming paragraph saying why); its census entry and the `CensusCheck`/`EndpointCensus` surfaces kept consistent under whichever name; the "Five endpoints" ↔ "Six endpoints" module-doc reconciliation (both PRs rewrote that paragraph); the overlapping `source-map.md` edits merged; and #194's `DeploymentRecord` import landing in the renamed `Action.lean`. Since this PR should merge before the commit for the external audit is pinned, the rebase deserves a careful pass of its own rather than a conflict-marker sweep.
2. **A comment becomes false on rebase:** this PR's updated `check_endpoint_census.sh` comment claims `*workFactor*` "currently match[es] nothing" — post-#194, the deployed endpoint's name contains `workFactor` and is matched by exactly that pattern. If the deployed endpoint is renamed into the scheme, the comment becomes true again; either way it needs a deliberate look, since it documents which guard patterns are live.
3. **Minor:** the ~30-line `+native(…)` witness list is now copied verbatim for a sixth time (and #194 added two more copies). If the census macro can take a shared named list, that's worth a follow-up; if not, this is the cost of explicitness and fine.

Overall: the contract layer is exactly the auditor-facing artifact the PR body promises, and the naming rationalization improves the endpoint surface. The one thing I'd insist on before merging is doing the #194 rebase attentively — two PRs each rewrote the meaning-carrying front matter of the same endpoint file, and the reconciliation is a design decision the merge tool cannot make.

TalDerei:
currently rebasing

PR #192, opened 2026-08-08
feat(fv): book entry for the advantage function

TalDerei:
Adds an Interpreting Security page (`book/src/formal-verification/action-advantage-function.md`): what the Action knowledge-soundness bound means, with an expert section stating the advantage function precisely, a plain-language section, and an illustrative log-log work chart. It also removes the three todo-stub entries (Preliminaries, User Documentation, Migration) from the sidebar.

Since review began, the branch was rebased onto main (the old guide and contract ancestry dropped when #78 and #191 merged) and gained two commits:

- symbols normalized to the book's conventions: lowercase q for query budgets, and the advantage function written `Adv_DLOG(q, g)` as the definitions page has it (the argument-order question is #205);
- claims stated at their scope: the subject is knowledge soundness of the deployed Action verifier —not "Action"— with the statistical error hedged in the bottom line, the witness computed by the extractor from the prover's declared representations rather than from the proof alone, and the covered scope naming the sampled-bases idealization. The numbers are unchanged.

🤖 Claude Fable 5


daira (approved):
ACK (with my changes).

PR #194, opened 2026-08-09
chore(fv): address audit finding and clarify trust boundaries

TalDerei:
Hodgepodge of feedback from Codex's cyber model and Zellic's v12, largely confirming boundaries the development already discloses; each one is now either remediated or documented in place with a doc comment at the relevant theorem stating the accepted floor and its known extension.

There will be merge conflicts with #191, but this preserves the organizational changes and decisions from that PR and applies them here.

## AI Summary

- Discloses the accepted model floors — adversary class, work accounting, URS basis, challenge uniformity, DLOG hardness, key digest, acceptance — as doc comments at the endpoints they qualify.
- Prices the deployed `Challenge255` conversion's exact reduction bias rather than only naming the idealization, and pins it in the trust-boundary census.
- Hardens the review guards: a second, environment-level endpoint census (`CensusCheck` target) that surface-syntax evasions cannot slip past; a source census pinning the staged group-work host callbacks; the csimp census now checks every target of a multi-target `attribute [csimp]`.
- Fixes audit-triage nits: symlink guard in the fixture-regeneration script, `<`-escape in the book's todo preprocessor, and contract docstrings for `ipaFold` lengths, the Nat-kernel MSM zero-window degeneracy, and ambient-root classification.


daira (approved):
ACK

daira:
I've read the full diff (2,168 lines, 37 files). Here's the post-hoc review.

## Overview

The PR does four distinct things, all responsive to the audit findings: (1) a genuinely new proof layer pricing the deployed `Challenge255` conversion's reduction bias, composed adaptively; (2) a machine-readable `ActionDeploymentInstantiation` record plus a deployed capstone endpoint consuming it; (3) three hardened review guards (environment-level endpoint census, staged group-work source census, fixture provenance manifest); and (4) a sweep of floor-disclosure docstrings plus small triage fixes. The build-checked pieces are sound as far as I can verify from the diff; my findings are mostly robustness nits in the shell guards.

## The mathematics (checked)

- **Challenge255 bias is exactly right.** With `2^512 = q·p + r`, the overshoot of any event is `(p·h − s·r)/(p·2^512)` for `s` elements of which `h` are heavy, maximized at `s = h = r`, giving `r(p−r)/(p·2^512)` — attained, so the constant is exact as claimed. `challenge255_key_le` and its weighted ENNReal twin encode precisely this. The weighted form (`PMFWeightedBiasLE`) is the right strengthening for adaptive composition, and the `Q·ε` accumulation in `runFreshPMF_eventBiasLE` is a clean induction over the query bound.
- **Deduplication is handled correctly.** `runFreshPMF` gives fresh answers per node, which over-samples repeated RO queries; the new `run_dedup` lemma (dedup preserves lookup-table execution) plus the record's `failureObserverQueryBound` on the dedup'd observer close that soundness gap explicitly. The docstrings even warn the joint hybrid must be applied post-dedup.
- **`ZeroBasisAcceptance`** is what it says: computation-free structural steps, with the two remainders (assembly success at a guard-passing oracle; the base-provenance walk) named rather than glossed.

## Design observations

- **The `ActionDeploymentInstantiation` record is the strongest thing in the PR** — the "push through to what the deployed protocol actually uses" move, done here for the SNARK capstones, with each model floor as a typed identification field instead of prose. Two details I particularly like: `dlogAdvantageAgrees` closing the advertise-a-different-advantage loophole, and the honest placement of the lazy-vs-eager table-sampling equivalence inside `idealFailureMeasure_eq` — that field is where the real refinement seam hides, and its docstring says so.
- **The environment-level census fails closed in the right places**: `isCensusKind` is an exhaustive match so a new declaration kind breaks the build until classified, and the forged-declaration tests pin the exact expected discovery set. The `Zcash`-module-root scoping and the disclosed "a committer who can edit the lakefile crosses the same review boundary" floor in `ambientPackageRoots` are honest about what a ratchet can and cannot do.
- **The Ledger docstring rewrites** are pure improvements: "at no extra cost in `k`" could be read as a derived absolute bound; "from one bound on the shared break event… the containment adds no factor of `k`" states what is actually proven.

## Findings (all minor; none affect proved statements)

1. **`check_costed_group_work_census.sh` dies undiagnosed if the costed types vanish.** Under `set -euo pipefail`, `actual_modules=$(find … | xargs -0 grep -lE … | sort)` exits the script via `grep`'s status-1 on zero matches, so the "census changed" diff diagnostic never prints in exactly the case it should. Same emptiness-needs-a-failure-distinguisher class as the manifest script handles correctly.
2. **The awk `-` counter counts every hyphen** — including `->` arrows and inline `--` comments inside a def body (a comment dash counts twice). Since expected counts are pinned exactly it's only ratchet noise, not unsoundness, but the header's "direct scalar-multiplication, addition/negation … footprint" slightly overstates what the token counters measure.
3. **`ENDPOINT_RE`/`isEndpointBaseName` sync is convention-only.** Both sides carry keep-in-sync comments, which is probably proportionate, but nothing fails if they drift — a drifted pair silently narrows whichever check has the weaker pattern. A tiny fixture matched by both (like the forged-endpoint tests, but exercised by the shell script too) would make the sync check mechanical.

The `<`-escape in the book preprocessor is the right fix (only `<` can terminate a raw-text `<script>` element, and `JSON.parse` round-trips the escape), and the symlink guard in the regeneration script correctly protects the `git reset --hard`/`clean` target.

Overall: a strong PR — the audit responses that could be theorems became theorems, and the ones that couldn't became either typed record fields or disclosed floors with named extensions.

PR #195, opened 2026-08-11
chore(book): fix pinned journey

TalDerei:
fixes failing book CI on main.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -48,8 +48,14 @@ jobs:
     permissions:
       contents: read
     steps:
+      # Full history: the proof journey pins its sources to commits, and a pin
+      # that is an ancestor of `main` rather than its tip is absent from the
+      # default depth-1 checkout. Fetching a branch into a shallow clone does
+      # not deepen it, so the pinned commit stays unreadable and the validator
+      # reports every source file as missing.
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
+          fetch-depth: 0
           persist-credentials: false
 
       - name: Install mdbook and plugins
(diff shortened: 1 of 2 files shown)


daira (commented):
## What it fixes

The **book CI** on main. `validate-proof-journey.py` checks each theorem anchor's symbol against its *pinned source commit* via `git show <ref>:<path>`. The failure mode:

- The journey pins `defaultSource: main` to a **specific commit**. Once main advances, that pin becomes an **ancestor** of main's tip rather than the tip itself.
- CI checked out **depth-1 (shallow)**, and fetching the `main` *branch* into a shallow clone does not deliver an ancestor commit's objects.
- The old readability probe was `git cat-file -e <ref>^{commit}` — but a shallow clone can hold the **commit object with its tree absent**. So the probe passed, then every `git show <ref>:<path>` failed → *every* source file reported "unavailable" → cascade.

## Assessment — correct and well-reasoned

- **`^{commit}` → `^{tree}`** (`source_is_readable`) is the key correctness fix: the tree is exactly what `git show` needs, and it's what goes missing in a shallow clone.
- **`fetch-depth: 0`** makes ancestor pins present after checkout (the common case), so no fetch is needed there; removing the hardcoded `git fetch … main` step (letting the validator fetch from the *recorded* pin) removes a ref that could drift out of sync with the data.
- **SHA-first fetch fallback** (`dict.fromkeys([source_ref, source["fetch"]])`, re-checking readability after each) makes cross-repo pins (a PR head) reachable by requesting the object directly, with the branch as fallback. The ordered-dedup idiom is fine.
- **`validate_endpoints` `complete` flag** — only flags "retired endpoint" when all declaring files were readable, so an unreadable file no longer buries the real error under spurious retirements.
- Comments explain the *why* (especially the `^{tree}` rationale), which is what a future reader needs here.

## Risks / nits (minor)

- A genuine file rename and an unreadable tree both surface as `git show` rc≠0 → the same "source file unavailable" text; they're distinguishable only by the separate fetch-failure line on stderr. Acceptable, and arguably out of scope.
- No test for the validator change — reasonable for a CI script; main CI going green is the proof.

**Verdict:** correct and complete for the stated failure — good to merge. Since it unblocks main CI, merging it also clears the red currently on the merge checks of other PRs based on main (e.g. #199, #200), which inherit the broken base.

PR #196, opened 2026-08-11
chore(fv): pin the instance-construction and generic-verifier audit floors

TalDerei:
Appending some low-hanging hardenings from this [audit list](https://gist.github.com/ebfull/b36e1c85af75f744fcc6b0e12212d3ae) — nothing exploitable, just pinning down boundaries we’d already flagged. Not all of the listed findings are valid: several are invalid, and several others describe documented idealizations of the model. This PR pins only the genuine boundaries.

- the nine/ten-row instance collision is now a named theorem, and exact-column / positive-count is pinned as `ActionDeploymentInstantiation` fields.
- documented the generic-verifier degradation sites and the certified-key facts that discharge them (docs only — nothing needed fixing).
- structural pre-NU6.3 separation — the old circuit is definitionally `synthesizeBase`, and `SpecPost = SpecBase ∧ cross-clause`.
- both honest capture point tables are checked to cover every `g`, `w`, and `u` index used by the augmented URS, preventing `getD` from silently substituting the identity.

The last two commits are naming and census follow-ups (the first is a broad terminology sweep, ~35 files):

- **Rename `CircuitPreIronwood` to `CircuitPreNU63` and normalize NU6.3 naming.** The Action circuit is selected by consensus branch, not by pool: the post-NU6.3 circuit serves both the Orchard and Ironwood pools. Renames the module and every reference; normalizes `NU 6.x` to `NU6.x` (keeping the verbatim `"post-NU 6.3 cross-address checks"` region label); replaces `ironwood`/`pre-ironwood` circuit-version labels with `post-`/`pre-NU6.3`, and `Ironwood` (for the halo2-native circuit development) with `Orchard-protocol`, keeping genuine references (the orchard `ironwood` git branch, `zcash/ironwood`, the project name); and records the post-NU6.3 circuit as released in orchard 0.15.0, correcting the fixture provenance in `PROVENANCE.md`. The auto-generated fixed-base and generator tables keep their `ironwood-branch` provenance headers, to be fixed in the unpublished generator.
- **Census the pre-/post-NU6.3 separation facts.** Moves `Separation.lean`'s inline `assert_no_sorry` checks to `assert_axioms` pins in `Zcash/TrustBoundary.lean`, so the facts join the library census.

### DIFF
--- a/Zcash/Circuits.lean
+++ b/Zcash/Circuits.lean
@@ -88,8 +88,8 @@ import Zcash.Circuits.Action.SelectorCoherence
 import Zcash.Circuits.Integration
 
 /-!
-# Ironwood circuits
+# Orchard-protocol circuits
 
-Aggregator for the Ironwood (halo2-native) circuit modules. Ironwood is kept separate
-from the main `Clean` library export and can be built with `lake build Zcash`.
+Aggregator for the Orchard-protocol (halo2-native) circuit modules. These are kept
+separate from the main `Clean` library export and can be built with `lake build Zcash`.
 -/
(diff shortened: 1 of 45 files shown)


daira (approved):
ACK

Issue #198, opened 2026-08-13
Indifferentiability of the Pasta group hashes from a random oracle (game side)


daira:
The Pasta group hashes are modelled as random oracles into the curve groups in our security arguments. daira/CompElliptic#19 formalized the deployed construction (`mapHashOutputsToCurve`, the two-term sum after `hash_to_field`) and its uniformity analysis (`TwoTermUniformity`, with the Weil bound as the named hypothesis `WeilBounded`). This issue tracks the game-theoretic side of justifying the random-oracle modelling: indifferentiability of the construction from a random oracle into the group, with `hash_to_field(·, 2)` modelled as a random oracle into `F × F`, following Theorem 1 of Brier, Coron, Icart, Madore, Randriam, and Tibouchi (<https://eprint.iacr.org/2009/340>).

The proof is split across the two repos so that CompElliptic contains no probability and ironwood contains no new counting. CompElliptic supplies exact counting facts (daira/CompElliptic#25, re-scoped): a per-point fibre bound for the single-term `mapToCurve`, the L¹/L² deviation exports from `TwoTermUniformity`, and the deployed-vs-zero-repaired pair transport. This side builds the probabilistic statement on the existing oracle machinery in `Zcash/Snark/Soundness/Oracle/`:

- the fibre sampler: the `PMF` on `F × F` that samples `u₀` uniformly, computes the at-most-`d` fibre of `Q - f(u₀)`, and rejection-corrects the weighting;
- the single-query lemma: `PMFWeightedBiasLE` between the two one-squeeze laws — uniform on `F × F` (real) and `(uniform G).bind sampler` (ideal) — with bias the regularity distance;
- the collapse: in both worlds every answer the distinguisher sees is a deterministic function of the per-query pair (the simulator's pair satisfies `F(pair) = R(m)` exactly), so the two-oracle game data-processes to a one-oracle game;
- `IndiffFromRO` over genuine `OracleComp` query trees, obtaining the multi-query bound from the single-query bias by the existing adaptive hybrid `runFreshPMF_eventBiasLE` — which charges a one-squeeze bias once per query node — after `dedup`;
- the costed simulator: exactly one `R`-query per fresh query, and the rejection-round cost as a distribution over operation counts (per-`k` tails, geometric expectation bound), on the costed oracle layer;
- a book page explaining what indifferentiability is and what the theorem does and does not give, written for readers who have not seen an indifferentiability proof before.

[...]

Issue #188, opened 2026-08-07
Model hash-to-curve as an adversary-queryable oracle


daira:
The AGM games fix, per game, an enumerated basis of the generators their honest algorithms use, and quantify over adversaries algebraic over that basis. A realistic adversary can evaluate `GroupHash` directly, and every output it obtains is a group element held with no representation over a fixed basis — so an adversary with that access is not algebraic over any fixed finite basis. The faithful modelling makes hash-to-curve an oracle of the game: the adversary may query it, each fresh output joins the AGM basis as a new independent element, and the reduction may embed its challenge in programmed outputs.

Two consequences, stated as known limitations on the Security Models page (#177):

* Spendability and Spend Authority's honest key generation calls `DiversifyHash`, so the modelled adversary cannot express strategies a realistic adversary performs routinely. Those games need the oracle in the adversary's interface before their capstones carry their intended weight.
* Enumerating, per game, the generators judged relevant leaves out fixed bases from other protocol components that a deployed adversary can obtain. Nothing known suggests they help, but "nothing known suggests" is itself a heuristic judgement, and it should be visible rather than implicit.

🤖 Filed with [Claude Code](https://claude.com/claude-code)

daira:
A concrete instance of the second bullet, from the review of #219: the Balance experiments' algebraic hypothesis for the binding signature, `AlgebraicAtBindingPointsAt` in `Zcash/Security/Ledger/ExtractionKnowledgeError.lean`, requires the adversary to announce representations of its binding keys over the two value bases 𝒱 and ℛ alone. A deployed algebraic adversary knows every public Orchard point and may build a binding key with a nonzero coefficient on, for example, the note-commitment domain point; such an adversary announces no two-slot representation and is outside the class the theorems quantify over. The narrowing cannot be removed by translation, since a representation of another public point over 𝒱 and ℛ is itself a nontrivial relation.

The combined deployed basis introduced in #219 (`orchardPoints`, the Sinsemilla table and the eight deployed points) is the natural target for the adapter:

* widen `AlgebraicAtBindingPointsAt` so representations range over the combined basis;
* take the pivot over every slot other than ℛ, so the knowledge-error argument's bad-challenge count covers each slot;
* keep extraction as reading the ℛ coefficient, and route the relation arm into the combined term, where a nonzero coefficient off the value slots already lands.

Posted with the assistance of Claude Fable 5.1

PR #29, opened 2026-07-03
feat(fv): bind witness columns to verifier openings

TalDerei:
**Human Summary:**

Closes https://github.com/zcash/ironwood/issues/18.

Binds the multiopen witness to decoded columns: the constraint-side capstones now
check the circuit gates on column polynomials recovered from the extracted IPA
witness's batch openings, instead of quantifying over arbitrary decode functions.

----------

**AI Summary:**

This PR closes the witness-to-columns half of the issue-#18 constraint-side
bridge. Mental model: the IPA extractor returns *one* coefficient vector for the
batched multiopen commitment, but the gate check speaks about individual column
polynomials — and the old `hquot`/`hgood` hypotheses quantified over every
mathematical opening, which is unsatisfiable at a prime-order curve for any
decode that genuinely reads the witness. The fix is a decode layer: rewound
openings of the same batched commitment at distinct batching challenges
Vandermonde-invert into the real columns, the extracted witness is pinned as
their batch-challenge combination, and the gate check is stated over exactly
that canonical decode.

## What Is Proven

- **Decode layer** (`MultiopenDecode.lean`): `BatchOpeningsForWitness` models
  rewound batched openings at pairwise-distinct challenges;
  `decodedColumnFamily_of_batch_openings` recovers the columns via the explicit
  Vandermonde inverse (no `Classical.choice`), and the `reconstruct` field pins
  every rewound witness — hence the extracted one (`currentWitness_eq`) — as the
  challenge-power combination of the decoded coefficients: witness-level
  binding, not merely same-commitment openings.
- **Canonical-decode hypotheses**: all decoded capstones state `hquot`/`hgood`
  for `decodedCols` of the *supplied* batch — the family their proofs construct.
  Quantifying over every decoded family instead is jointly unsatisfiable (only
  the commitment and one evaluation pin a column; a kernel tweak vanishing at
  the opened point and `x` defeats the pair) — the argument is recorded in the
  module doc's scope section.
- **Witness tie**: the transcript-tied capstones extract the transcript's own
  witness (`ipaRelation_of_acceptV`) and identify it with the batch's — a
  mismatch is a `commit` collision, returned as the relation branch
  (`hasNontrivialRelation_of_two_openings`) — so `hencodes` only ever receives
  the extracted witness.
- **Three bridge granularities**, one conclusion shape: transcript-tied
  (`hbatch : ∀ t, IpaAcceptV … t → MultiopenRewindBatch … t`), bundled
[...]

PR #199, opened 2026-08-15
Group-hash indifferentiability: security games, rejection-sampling simulator, and concrete endpoints

daira:
The ironwood side of the indifferentiability arc (#198): the one-oracle game, the collapse of the full two-oracle game onto it, the rejection-sampling simulator, and concrete endpoints at the deployed mappings. The counting side is daira/CompElliptic#27, and the Weil-bound reduction above it is daira/CompElliptic#29, both merged; the pin references CompElliptic main at the #29 merge.

**The headline.** `pallas_`/`vesta_indiffFromRO` (`Indiff.lean`): every `q`-query adaptive distinguisher between the real per-query law and the ideal one is moved by at most `q/2^120`, in both directions and against every acceptance event. The only inputs are the two `CharSumBounded` bounds at `100·#F` — Weil's theorem at the branch covers. The parametric forms keep the `_of_weilBounded` suffix.

**The route** (each module's doc has the detail):

- `Sampler.lean` — the single-query core: the one-squeeze bias in both worlds is bounded by the regularity distance.
- `Pasta.lean` — the deployed instantiation under the named `WeilBounded` hypothesis, with the zero-repair transport at `(4·#F − 2)/(#F)²`, and the simulator's counting input `deployedFibreBound = 11`.
- `MultiOracleComp.lean` / `TwoOracle.lean` — indexed query trees with per-index budgets; the two-oracle game collapses computably onto the one-oracle core at the folded budget.
- `Model.lean` — the `PMFEventBiasLE` combinators the arc composes with.
- `Simulator.lean` — the capped rejection sampler behind the ideal law, with exact costs: round-count tail `(1 − p)^k`, geometric round law, two-sided output bias `(1 − p)^K`, exactness on empty fibres, and convergence to the fibre sampler as `K → ∞`.
- `Indiff.lean` — the compositions: `indiffFromRO_of_regularity` (a regularity-distance budget of `β` gives indifferentiability at `q·β`), the capped variant charging `simLawBias` per query, and the concrete endpoints (`β := 1/2^120 − 4/#F` makes the transport term telescope; the rest is exact arithmetic at the deployed sizes).

**The book page** teaches the argument to a general mathematical audience and states the what-is-proved/what-is-modelled boundary; its Conclusion gives the advantage as the single term `q/2^120`.

**Census**: the arc's top-level leaves are pinned, with their `native_decide` owners named — the Pasta base-field structures, and the curve-group order certificates at the concrete endpoints and the simulator pins; the census module doc's CompElliptic owner tally ends at six.

**What stays cited**: Weil's theorem at the two branch covers — stating it needs function-field vocabulary tracked at daira/CompElliptic#30. The consuming games' oracle access is #188.

The first commit adopts CompElliptic's absorbed `CompEllipticExtras` (daira/CompElliptic#12) in place of ironwood's copies.

🤖 Claude Fable 5, Claude Opus 4.8 (1M context)


Zcash/Security/GroupHash/Sampler.lean
```diff
@@ -0,0 +1,449 @@
...
+    (hne : (fibre f Q).Nonempty) (hp : p ∉ fibre f Q) : fibreSampler f Q p = 0 := by
+  rw [fibreSampler, dif_pos hne, PMF.uniformOfFinset_apply_of_notMem _ hp]
+
+omit [DecidableEq F] [Fintype G] in
+/-- On the empty fibre of `Q`, the sampler falls back to the uniform pair weight. -/
+theorem fibreSampler_apply_empty [Nonempty F] (f : F → G) {Q : G} (p : F × F)
+    (hempty : ¬ (fibre f Q).Nonempty) :
+    fibreSampler f Q p = (Fintype.card (F × F) : ℝ≥0∞)⁻¹ := by
+  rw [fibreSampler, dif_neg hempty, PMF.uniformOfFintype_apply]
+omit [DecidableEq F] [Fintype G] in
```
daira, line 151:
```suggestion

omit [DecidableEq F] [Fintype G] in
```

Zcash/Security/GroupHash/Simulator.lean
```diff
@@ -0,0 +1,609 @@
+import Zcash.Security.GroupHash.Sampler
+
+/-!
+# The costed simulator
```
daira:
```suggestion
# The costed rejection-sampling simulator
```

Zcash/TrustBoundary.lean
```diff
@@ -675,9 +754,10 @@ key-binding and ledger sections above, expresses them through the `Zcash.Meta.Ax
   inherit CompElliptic's `native_decide` curve point-count axiom (`+native`).
 
   Note that CompElliptic's witnesses are *not* the bulk of what `+native` covers library-wide.
-  Exactly three owners across the whole census are CompElliptic's: the two curve point counts
-  (`Pallas.q_nsmul_Gpt`, `Vesta.p_nsmul_Gpt`) and the Tonelli–Shanks root-of-unity data
-  (`Fields.Pasta.pallasBase`, whose certificate sits in a structure-field auto-param). Every other
+  Exactly four owners across the whole census are CompElliptic's: the two curve point counts
+  (`Pallas.q_nsmul_Gpt`, `Vesta.p_nsmul_Gpt`) and the two Tonelli–Shanks root-of-unity data
+  (`Fields.Pasta.pallasBase` and `Fields.Pasta.vestaBase`, whose certificates sit in
+  structure-field auto-params). Every other
```
daira: Nit: rewrap.

daira (commented):
Self-ACK, only minor nits.

daira (commented):
Self-reACK

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,592 @@
...
+is essentially what it did — pick one of the coordinates, and then the equation
+in the other is quadratic. It's easy to construct a hash into the field with
+low bias using a conventional hash function with a large enough output size, by
+taking its output as an integer modulo the field size. Then the mapping above
+is bijective, so its output points will be approximately evenly distributed,
+although only among *half* of the curve points — the half chosen by the
+deterministic $\possqrt{\cdot}$.
+
+The problem is that then not all $x$ coordinates, and therefore not all inputs
+to the hash, map to a point. For each $x$-coordinate, we have 0, 1, or 2
```
daira: Nit: inconsistent hyphenation of $x$-coordinate[s].

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+low bias using a conventional hash function with a large enough output size, by
+taking its output as an integer modulo the field size. Then the mapping above
+is bijective, so its output points will be approximately evenly distributed,
+although only among *half* of the curve points — the half chosen by the
+deterministic $\possqrt{\cdot}$.
+
+The problem is that then not all $x$-coordinates, and therefore not all inputs
+to the hash, map to a point. For each $x$-coordinate, we have 0, 1, or 2
+solutions for $y$ depending on the number of square roots of $x^3 + a \cdot x + b$.
+Heuristically, roughly half of the $x$-coordinates should have no solutions
```
dannywillems, line 55: Reviewer notes: this is because of the number of (non-)quadratic residues ((p+1)/2 - (p-1)/2) when the modulo is prime.

dannywillems: I was initially confused by the wording "Heuristically [...]".

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+$1/\sqrt{\FieldSize}$ of $\fraction{1}{2}$.
+
+For fixed generators, having a group hash that is not a total function is not
+so much of a problem: we can extend it to a total function by repeated hashing with
+an index. Since there are only a fixed set of generators and they are found off-line,
+non-constant timing due to the variable number of iterations is not an issue.
+But Sapling had introduced *diversified addresses*, which require on-line use
+of the group hash in order to derive an address from a diversifier. We avoided
+timing attacks in Sapling by not doing repeated hashing, and accepting that
+only half of all diversifiers would be valid. But we had encountered complications
```
dannywillems, line 77: Comment: I like these historical notes.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+in the application protocol ([ZIP 32](https://zips.z.cash/zip-0032) and its usage)
+due to this abstraction leak from the underlying cryptography. We wanted to avoid
+that when designing Orchard.
+
+Fortunately, an Informational RFC for deterministic, constant-time Hashing to
```
dannywillems, line 82: > [...] constant-time Hashing to Elliptic Curves [...]

Reviewer note: https://www.rfc-editor.org/rfc/rfc9380.html#name-constant-time-requirements.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+available for each input $u$.
+
+It also fixes the sign at the end: the output's $y$-coordinate is negated if
+necessary so that its sign matches the sign of the input, in the convention
+that RFC 9380 calls `sgn0`. This makes $f$ *odd*, that is, $f(-u) = -f(u)$
+for $u \neq 0$. Oddness carries weight below: it is what splits an input pair
+$\{u, -u\}$ across a point and its negation, and the character-sum analysis
+relies on it too.
+
+The images of $g$ and $h$ are not disjoint; for Simplified SWU they in fact
```
dannywillems: Maybe the reader is supposed to be reading the link redirecting the the construction "Simplified SWU" given above, but if a reader jumps to this section directly by reading linearly this document, they might be missing some contexts. At least, it is the case for me.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+by the deterministic square root, using the other half of the $x$-coordinates.
+But there is no known way of doing so (in fact, if there were then it would
+indicate undesired structure and potential cryptographic weaknesses in the curve).
+
+The basic idea of having two different cases depending on whether a given input
+yields solutions for a square root, however, is exactly what RFC 9380's
+"[Simplified SWU](https://www.rfc-editor.org/rfc/rfc9380.html#name-simplified-shallue-van-de-w)"
+construction does. For now we will ignore a complication that arises for short
+Weierstrass curves with $a = 0$, like Pallas and Vesta; we'll get to that in
+[its own section](#the-detour-through-an-isogenous-curve). Then, ignoring
```
dannywillems, line 102: Nit/Reviewer note: until now, in this document, we have discussed about mapping $u 
\leftarrow (u, f(u))$. However, the definition of the construction called Simplified SWU is not about mapping identically `u` for the `x` coordinate; there is a transformation. A consequence is that we don't only talk about computation the "square root of a value". I think that it would be nice to mention it somehow - even though it is surfaced in the first paragraph of this section.
Personally, I had to go back to the RFC9380 to refresh my memory.

book/src/formal-verification/group-hash-indifferentiability.md
dannywillems: Related to this comment: https://github.com/zcash/ironwood/pull/199/changes#r3806392071.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+reaching $P$ via $h$ and vice versa. Hence $P$ is reached via $g$ iff it is reached
+via $h$.
+
+This coexists with the exact halves above because those partition the *inputs*,
+not the outputs. The correspondence $u \mapsto \pm 1/(Z \cdot u)$ carries the
+$g$-half of the inputs into the $h$-half and back, preserving the point reached.
+About $\fraction{3}{8}$ of the output space is reached
+—with 2 or 4 preimages per reached point excluding exceptional cases— and
+the remaining $\fraction{5}{8}$ by neither map. (These proportions are heuristic;
+we confirmed them by exact computation on small curves, and they can be proven
```
dannywillems: Reviewer note: "[...] we confirmed them by exact computation on small curves [...]" - I think a link would be useful for the reader, to confirm the statement.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+with error $O(1/\sqrt{\FieldSize})$ by counting points on the branch varieties —
+Lang–Weil, "Number of Points of Varieties in Finite Fields",
+*Amer. J. Math.* 76(4), 1954, [doi:10.2307/2372655](https://doi.org/10.2307/2372655).
```
dannywillems, line 153: Reviewer note: it would be nice to point to the actual proof.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+via $h$.
+
+This coexists with the exact halves above because those partition the *inputs*,
+not the outputs. The correspondence $u \mapsto \pm 1/(Z \cdot u)$ carries the
+$g$-half of the inputs into the $h$-half and back, preserving the point reached.
+About $\fraction{3}{8}$ of the output space is reached
+—with 2 or 4 preimages per reached point excluding exceptional cases— and
+the remaining $\fraction{5}{8}$ by neither map. (These proportions are heuristic;
+we confirmed them by exact computation on small curves, and they can be proven
+with error $O(1/\sqrt{\FieldSize})$ by counting points on the branch varieties —
```
dannywillems, line 151: Also, is there a proof of it written in Lean/Rocq/others? Maybe it is in `CompElliptic` as part of this work, but I haven't had a look yet.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+Lang–Weil, "Number of Points of Varieties in Finite Fields",
+*Amer. J. Math.* 76(4), 1954, [doi:10.2307/2372655](https://doi.org/10.2307/2372655).
+A modern exposition of that paper is Tao,
+[The Lang-Weil bound](https://terrytao.wordpress.com/2012/08/31/the-lang-weil-bound/),
```
dannywillems, line 155: Reviewer note: I didn't read the proof, nor verify if the hypothesis of the theorem satisfies our cases. Here is the Lang-Weil bound theorem stated in the Terry Tao's article:

Image.

I went quickly through the blog article, and I didn't see any Lean4 formalization.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+
+This coexists with the exact halves above because those partition the *inputs*,
+not the outputs. The correspondence $u \mapsto \pm 1/(Z \cdot u)$ carries the
+$g$-half of the inputs into the $h$-half and back, preserving the point reached.
+About $\fraction{3}{8}$ of the output space is reached
+—with 2 or 4 preimages per reached point excluding exceptional cases— and
+the remaining $\fraction{5}{8}$ by neither map. (These proportions are heuristic;
+we confirmed them by exact computation on small curves, and they can be proven
+with error $O(1/\sqrt{\FieldSize})$ by counting points on the branch varieties —
+Lang–Weil, "Number of Points of Varieties in Finite Fields",
```
dannywillems, line 152: I did _not_ read in detail the article linked by the JSTOR article. However, the content seems aligned with the algebraic structures and the statements made here.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+*Amer. J. Math.* 76(4), 1954, [doi:10.2307/2372655](https://doi.org/10.2307/2372655).
+A modern exposition of that paper is Tao,
+[The Lang-Weil bound](https://terrytao.wordpress.com/2012/08/31/the-lang-weil-bound/),
+2012.)
+
+```admonish info title="Where the ⅜ comes from"
```
dannywillems, line 158: At the time of this comment, I skipped a formal read for most of the above, and until `[...] Particular application protocols [...]`, as, I believe, it requires the context of the Simplified SWU construction. I might be coming back to this while (resp. after) I'm going (resp. I went) through the Lean4 specification to see what is actually proven.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+and one to its negation. So the point is reached from $2$ preimages (one per
+branch) with probability $\fraction{1}{2} \cdot \fraction{1}{2} = \fraction{1}{4}$,
+and from $4$ preimages (two per branch) with probability
+$\fraction{1}{2} \cdot \fraction{1}{2} \cdot \fraction{1}{2} = \fraction{1}{8}$;
+otherwise it is unreached. The reach probability is
+$\fraction{1}{4} + \fraction{1}{8} = \fraction{3}{8}$, and reached points have
+$\fraction{8}{3}$ preimages on average.
+```
+
+It turns out, for the Pasta curves, that we cannot do much better than this
```
dannywillems, line 184: `we cannot do much better than this`: is it formal statement? Or is it "at the time of writing, we don't know any method to achieve better than this?"

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+$\fraction{3}{8}$ coverage by mapping directly from a single field element (or
+at least, trying to do so would not lead to a less complicated scheme overall,
+given other constraints like the desire for a constant-time group hash).
+
+Particular application protocols might actually be perfectly fine with this
+kind of non-uniform mapping. However, it can easily be distinguished from a
+uniform one, and so we would have to take account of that non-uniformity in
```
dannywillems: `[...] so we would have to take account of that non-uniformity in all of our security arguments [...]` - I keep a note while reviewing. It should probably be mentioned in the Lean4 codebase too, as a comment/reminder - it might be already existing, I haven't read it yet at the time of writing this comment.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+The short Weierstrass form with $a = 0$ corresponds to curves with
+$j$-invariant $0$, that is, with Complex Multiplication by $\mathbb{Z}[\zeta_3]$
+and an automorphism group of order $6$: there are exactly six invertible mappings
+from the curve to itself that preserve the group structure, namely
+$(x, y) \mapsto (\zeta_3^k \cdot x, \pm y)$. (These stay on the curve because
+$x$ appears only cubed, and $(\zeta_3^k)^3 = 1$.)
+
+Daira-Emma Hopwood's ZK Study Club talk
+"Optimizing Halo and Constructing Graphs of Elliptic Curves"
+([part 1](https://www.youtube.com/watch?v=q7bAYgxkHUE),
```
dannywillems, line 218: Reviewer comment: good old times 💯.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+both have the curve equation $y^2 = x^3 + 5$, i.e. $a = 0$. This is not by
+coincidence; the same Complex Multiplication structure that allows us to
+find a cycle of curves is what blocks Simplified SWU from working.
+
+```admonish info title="One symmetry, two effects"
+The short Weierstrass form with $a = 0$ corresponds to curves with
+$j$-invariant $0$, that is, with Complex Multiplication by $\mathbb{Z}[\zeta_3]$
+and an automorphism group of order $6$: there are exactly six invertible mappings
+from the curve to itself that preserve the group structure, namely
+$(x, y) \mapsto (\zeta_3^k \cdot x, \pm y)$. (These stay on the curve because
```
dannywillems: Reviewer note: checking that this is proven later in `CompElliptic`.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+
+Daira-Emma Hopwood's ZK Study Club talk
+"Optimizing Halo and Constructing Graphs of Elliptic Curves"
+([part 1](https://www.youtube.com/watch?v=q7bAYgxkHUE),
+[bonus session](https://www.youtube.com/watch?v=IQVGIqcdxL4),
+[slides](https://raw.githubusercontent.com/daira/halographs/master/halographs.pdf))
+explains why the Pasta curves have this form: the two curves of a 2-cycle
+necessarily share their CM discriminant, and with known methods a cycle can only
+feasibly be found when that discriminant is tiny (the Pasta search fixed the
+smallest, $D = -3$, which is exactly the case $j = 0$). Slides 8 and 9 give a
```
dannywillems, line 224: Reviewer comment: I would love having an explanation of the visual form of the argument; I'm struggling understanding it.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,644 @@
...
+$x$ appears only cubed, and $(\zeta_3^k)^3 = 1$.)
+
+Daira-Emma Hopwood's ZK Study Club talk
+"Optimizing Halo and Constructing Graphs of Elliptic Curves"
+([part 1](https://www.youtube.com/watch?v=q7bAYgxkHUE),
+[bonus session](https://www.youtube.com/watch?v=IQVGIqcdxL4),
+[slides](https://raw.githubusercontent.com/daira/halographs/master/halographs.pdf))
+explains why the Pasta curves have this form: the two curves of a 2-cycle
+necessarily share their CM discriminant, and with known methods a cycle can only
+feasibly be found when that discriminant is tiny (the Pasta search fixed the
```
dannywillems, line 223: `[...] that discriminant is tiny [...]`

Is it possible to quantify the meaning of "tiny". Tiny relative to?

dannywillems (commented):
Mostly comments. First batch. Include https://github.com/daira/CompElliptic/pull/31

book/src/formal-verification/group-hash-indifferentiability.md
daira: It's heuristic because it depends on the number of quadratic residues corresponding to $x^3 + ax + b$, not $x$. So it will be roughly half if $x^3 + ax + b$ behaves randomly. In fact the Hasse bound tells you that the heuristic is correct to within $`\pm 2\sqrt{\#\kern-0.1em F}`$, but that's precisely what the text below says.

book/src/formal-verification/group-hash-indifferentiability.md
daira: I was trying to avoid getting into the nitty gritty details that I think are not important for a high-level understanding. All the detail you could possibly want is in `SimplifiedSWU.lean` and `BranchCovers.lean` in CompElliptic.

book/src/formal-verification/group-hash-indifferentiability.md
daira: I think that script is not actually in the CompElliptic repo, I'll add it.

book/src/formal-verification/group-hash-indifferentiability.md
daira: I think there isn't, daira/CompElliptic#29 doesn't have it directly. We don't actually need it; the security bound does not depend on it. (We prove a lot of things — these PRs are huge, so I don't want to prove things just because we can.)

book/src/formal-verification/group-hash-indifferentiability.md
daira: The statement of what we need (essentially a concrete version of this) is in daira/CompElliptic#27, `WellDistributed.lean`.

book/src/formal-verification/group-hash-indifferentiability.md
daira: No it's not a formal statement. We can get to 5/8 for some curves (Icart's method, https://eprint.iacr.org/2009/226.pdf). But 5/8 is not that much better than 3/8.

book/src/formal-verification/group-hash-indifferentiability.md
daira: Well, strictly speaking we still do, but the point of this formalization is that the adversary's advantage after $q$ queries is bounded by $q/2^{120}$ so it's not going to break any security arguments. I will tighten up this wording.

book/src/formal-verification/group-hash-indifferentiability.md
daira: This is just background. We don't need it for indifferentiability. We have to focus on what we actually *need* to prove, because just that is already very complicated.

book/src/formal-verification/group-hash-indifferentiability.md
daira: https://arxiv.org/pdf/0912.1831 explains why we cycles of CM curves are much more common than cycles of non-CM curves. Up to isomorphism there are only 13 CM curves over $\mathbb{Q}$, and they all have discriminant $\leq 163$ in absolute value. This doesn't exclude there being (many!) curves over finite fields that don't have small discriminant in absolute value, but it's part of why they are hard to find.

book/src/formal-verification/group-hash-indifferentiability.md
daira: I suggest watching the talk!

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,658 @@
...
+A *distinguisher* $\Dist$ talks to two oracles and tries to tell which of two
+worlds it is in:
+
+- the **real world** $(\mathsf{hash\_to\_field}, H)$ — the genuine intermediate
+  oracle and the genuine construction built on top of it;
+- the **ideal world** $(\Sim^R, R)$ — the ideal random oracle $R$ into the
+  group, and the simulator faking the intermediate hash consistently with it.
+
+The construction is $(q, \eps)$-**indifferentiable** if some simulator makes
+every distinguisher's advantage at most $\eps$ after $q$ queries. The point
```
daira, line 313: We use $\varepsilon$ here differently to everywhere else (but the usage here is the more conventional one). Elsewhere we use it for the single-query bias that contributes the term $q \cdot \varepsilon$ to the advantage. We should use a different name for the latter.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,658 @@
...
+explains why the Pasta curves have this form: the two curves of a 2-cycle
+necessarily share their CM discriminant, and with known methods a cycle can only
+feasibly be found when that discriminant is tiny (the Pasta search fixed the
+smallest, $D = -3$, which is exactly the case $j = 0$). Slides 8 and 9 give a
+nice visual form of the argument.
+
+Simplified SWU, for its part, obtains its branch pair by *solving for the
+$x$-coordinate* at which the scaling defect
+
+$$g(\lambda x) - \lambda^3 g(x) = a \lambda (1 - \lambda^2) \cdot x + b (1 - \lambda^3)$$
```
daira: This is not the same $g$ as before. It's the RHS of the curve equation as in RFC 9380, i.e. $g(x) = x^3 + ax + b$. We need to ~~expand the alphabet~~ find different names (the RFC uses $x_1$ and $x_2$ for the *outputs* of these functions; it does not name the functions themselves).

book/src/formal-verification/group-hash-indifferentiability.md
daira: No change needed since we link the RFC a few lines below, and the importance of constant-timeness is called out prominently in its Introduction.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,686 @@
...
+the coordinates, that preserves the identity point. In general it need not
+be invertible; over the algebraic closure, a degree-$3$ isogeny is
+$3$-to-$1$. For Pallas and Vesta, the auxiliary curves are the ones that
+the protocol specification and the `pasta_curves` crate call iso-Pallas
+and iso-Vesta respectively. Having used Simplified SWU to obtain a point
+on the auxiliary curve, we apply the isogeny (here of degree 3; the Pasta
+curves were chosen to make the degree as low as possible), in order to land
+on the intended curve.
+
+For the analysis on this page the detour is short, at least conceptually.
```
dannywillems, line 251: Reviewer note: this is mostly prose to say that these isogenies are isomorphisms (kernel identity = monomorphism).

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,686 @@
...
+
+Because the isogeny is a homomorphism, there are two equivalent ways to
+compute <span style="white-space: nowrap">$f(u_0) + f(u_1)$:</span>
+either by adding the two Simplified SWU outputs on the auxiliary curve and
+applying the isogeny once, or by mapping each point across the isogeny and
+then adding. The former method is used by RFC 9380 and `hashtocurve.sage`;
+the latter by `pasta_curves`. The two orders agree exactly
+(`mapHashOutputsToCurve_eq`), so nothing depends on the choice.
+
+Although a correctly constructed isogeny is *always* a homomorphism
```
dannywillems, line 270: Reviewer note: oh, I forgot that it was not _always_ defined as an homomorphism. [Wikipedia](https://en.wikipedia.org/wiki/Isogeny) does use it in its definition while Silverman actually shows it as a corollary of the definition - I checked quickly in the book. I confirm the theorem number.

book/src/formal-verification/group-hash-indifferentiability.md
```diff
@@ -0,0 +1,686 @@
...
+  (CompElliptic's [`Hashing/WeilSupport.lean`](https://github.com/daira/CompElliptic/blob/main/CompElliptic/Hashing/WeilSupport.lean)) — the delivered scope of
+  [CompElliptic#28](https://github.com/daira/CompElliptic/issues/28).
+  The input's own statement needs function-field vocabulary (genus,
+  places, covers) that Mathlib does not yet have, tracked at
+  [CompElliptic#30](https://github.com/daira/CompElliptic/issues/30).
+* The security games that want to use this result need the group hash
+  added to their adversary's interface first
+  ([#188](https://github.com/zcash/ironwood/issues/188)). The composition
+  requirement for multi-stage games (the [† note above](#dagger-note))
+  applies at each consumption site.
```
dannywillems, line 687: Reviewer note: LGTM for this section - a lot of interesting information!

dannywillems (commented):
Next batch. LGTM up to now. Remaining review are:
- the new files in Zcash/Security/GroupHash
- `MultiOracleComp.lean`

book/src/formal-verification/group-hash-indifferentiability.md
daira: It's defined as a rational map on coordinates that preserves the identity; the proof that this implies it's a homomorphism in general is nontrivial (to say the least). In any case, what we *implement* is the rational map, so we can't avoid the issue of showing that is a homomorphism by defining it away. That is, if an isogeny were defined to be a homomorphism given by a rational map on coordinates, then we'd have to prove we have an isogeny, which is the same proof.

The shortness of Silverman's proof is deceptive when it comes to formalization in Lean. Mathlib has the Picard group of a commutative ring (`CommRing.Pic`), and its group-law proof for elliptic curves is essentially based on Silverman's $\kappa$ (points map to ideal classes of the coordinate ring) — but the proof of III.4.8 also needs the pushforward along a finite morphism and the geometric $\mathsf{Pic}^0$ machinery underlying it, none of which exists.

dannywillems (approved):
ACK - I got a bit lost in side quests as I am onboarding on CompElliptic and this repo. I don't want to block future patches on top of this one. The patch is already relatively huge.

PR #200, opened 2026-08-15
Move the generic oracle machinery to Common/Oracle

daira:
Relocates the generic oracle-computation machinery out of `Zcash/Snark/` into a new `Zcash/Common/Oracle/`, so the layer that the Fiat–Shamir soundness reductions and the group-hash indifferentiability argument both build on lives in `Common/` rather than under one of its consumers. This is preparation for a `MultiOracleComp` generalization — the multi-oracle substrate the ledger Spend-authority / Spendability games and the indifferentiability two-oracle collapse both need — kept as a self-contained move so that the generalization lands against a clean base.

The move itself is mechanical; the cleanups it surfaced are separate labelled commits:

- **Delete the dead random-oracle reprogramming helpers** — `reprogram` / `reprogram_self` / `reprogram_ne` have no call sites (deleted once in zcash/ironwood#24, later crept back into the tree).
- **Move the generic oracle machinery from Snark to Common/Oracle** — `OracleComp`, the bias-interface model, the adaptive hybrid, the labelled-oracle trees, and the read-log wrapper move to `Zcash/Common/Oracle/` under the `Zcash.Common` namespace. The `Fp`-specific uniform-challenge idealization stays on the Snark side (split into `Oracle/ChallengeUniform.lean`, leaving `Oracle/Challenge255.lean` to price the deployed conversion), so nothing `Fp`- or transcript-specific moves and `Common` gains no dependency on `Snark`.
- **Refile misfiled census pins under their module headers** — pre-existing trust-census housekeeping in the AGM / adaptive-statement tail, unrelated to the move; the set of pinned endpoints is unchanged.
- **Pin the CostedOracle present-API surface with the census primitives** — replaces bare `#check` (which prints an info line per pin) with `assert_computable` for the executable cost programs and `assert_axioms` for the theorems and coverage predicates.

There are no reverse dependencies from `Common` to `Snark`. Full build green, including the endpoint / csimp / coverage census scripts and the proof-journey validator.

🤖 Claude Opus 4.8 (1M context)


daira (commented):
Self-ACK

TalDerei (approved):

PR #206, opened 2026-08-22
Retitle the README and document the build and CI checks

daira:
README and build-documentation changes:

- Retitle to "Formalizing Ironwood". The previous title "Ironwood ← Setup(1<sup>λ</sup>)" borrowed the security-parameter argument from asymptotic security definitions, where it makes the setup polynomial-time in λ. This development is a concrete-security analysis of fixed curves and hashes, so the device does not apply. It is also misleading: it suggests a fresh setup, while Ironwood reuses a URS fixed when Orchard was designed — the point from which attacks against it could begin, as the Security Models page describes. The new title deliberately reads as an ongoing process.
- Set "The Ironwood Book" link in Raleway where styles apply. The book includes the README as its landing page, so the link picks up the book's Raleway face at the intended weight via a stylesheet class. GitHub strips the attribute and renders a plain link (confirmed on the branch).
- Correct the verification command. `lake build` was described as "exactly what CI runs", but CI treats warnings as errors and runs source-level consistency checks besides. The README now recommends `lake build --wfail` — which builds every Lean file CI builds — and links a new "Build and CI Checks" reference page in the book documenting the build targets, the check scripts, fixture regeneration and byte-pinning, the book workflow's proof-journey validation, and what is not covered of the CompElliptic dependency's own checks.

🤖 Filed with Claude Code (Fable 5)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Ironwood $\small\gets \text{Setup}(1^\lambda)$
+# Formalizing Ironwood
 
 $\textsf{Ironwood}$ is a project to deploy a new shielded pool, built to restore
 confidence for all Zcash'rs in the supply integrity of Zcash.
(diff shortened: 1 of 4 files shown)


daira (commented):
Self-ACK

TalDerei (approved):
ACK

PR #161, opened 2026-07-31
Eliminate derived keygen coherence certificates

mitschabaude:
Derives Action circuit correctness from compositional Clean/Halo2 lawfulness instead of full-circuit native checks.

- Integrates keygen, query, permutation, copy, selector, planner, domain, constraint-degree, and fixed-column laws across the circuit hierarchy.
- Deletes the corresponding Action-specific sidecars and 27 unique native certificate dependencies: 19 circuit/layout checks and eight fixed-window/scalar checks.
- Retains native checks for pinned fixture data, rather than using them to establish circuit-correctness claims.

PR #211, opened 2026-08-24
fv: the circuit-to-ledger Action witness bridge

daira:
The knowledge-soundness bridge: turn the Action circuit's extracted witnesses into the ledger's annotated actions, closing the gap that #171's `_idealizedks` capstones name as a gap.

## The Action-to-ledger refinement as data

`actionSpecToLedgerData` replaces the bridge's existential endpoint: classify the combined circuit witness, and either return the refined ledger action as data (`ActionLedgerSuccess` — the instance, the witness, and the proofs that they satisfy the games-facing statement) or the computed discrete-log relation of the first Sinsemilla escape (`ActionDLBreak`). The structure is indexed by the full circuit witness, so no component of the extracted data is projected away. The success construction computes through the ledger's own primitives: the key-binding witness's hash point is the defined value of `Pool.commitIvkHash`, its incoming viewing key is the extraction the opening relation demands (so that equation holds by construction), and the new-note commitment is the defined value of `Pool.noteCommit`. The existential-returning forms (`actionSpec_to_ledger`, `commitIvkSuccess_to_ledger`, `noteCommitNewSuccess_to_ledger`) are deleted rather than kept as corollaries: returning the data already shows the existential form is inhabited. Census and prose pointers move to the data forms.

## The named vocabulary and the constructive success layer

This is a refactoring cleanup — not strictly necessary for the circuit-to-ledger bridge, but this was a good time to do it.

The per-site hash abbreviations (`ivkHash`, `noteOldHash`, `noteNewHash`, `merkleHash`) pin the deployed generator table and each site's domain point once; `extractP` names `Extract_P` (§5.4.9.7), with the representation caveat that makes `.x` a correct implementation; and `sinsemillaCommitBlind`/`sinsemillaCommitBlindShort` name the commitment steps (§5.4.8.4), so statements do not depend on how `SinsemillaCommit` spells them. The `∃`-shaped definedness facts and success predicates become their constructive normal forms: `isSome` fields with the values recovered by `Option.get` and computed accessors (`commitIvkOf`, `merkleRootOf`). `successes_or_break`/`successes_of_noBreak` are deleted —the classifier match was already the mechanism— and the `_get_spec` dependent pairs split into `_get_valid`/`_get_eq`, derived through the Pool lemmas, which also removes the consumers' `Exists.choose` uses. The reduction's site and the extraction-failure arm's datum are likewise computed (`classifierSiteOf`, `extractFailureOf`).

## The Satisfying structure and the bundle lift

`Zcash.Common.Satisfying R x` is a Type-valued structure bundling a witness with its satisfaction proof. It is modelled on zcash-lean's `Satisfying`, with the divergences noted in its module doc. `TopLevelSemanticWitness` becomes an abbreviation of it. Each accepted bundle member's extracted witness is then either a satisfying witness of the circuit-facing `ActionSpec` (`memberSatisfying`) refined to the member's ledger data, or the computed escape relation (`memberLedgerData`), and is traversed across the bundle via `bundleLedgerData`. The census pins all of these as computed data.

## The composed ledger extractor

`actionLedgerOutcome` composes the bundle bridge with the adaptive-statement knowledge outcome: one run yields every member's ledger data, a ledger Sinsemilla escape, or the circuit-side algebraic relation over the run's own basis, and `none` is exactly the runs on which the knowledge outcome is undefined (`actionLedgerOutcome_isSome_iff`). The routing is deliberate: circuit-side relations stay in their own arm —on accepting runs, that arm and `none` together are the knowledge-failure event that the knowledge-error endpoint bounds— and only the bridge's Sinsemilla escapes surface as ledger breaks. `actionLedgerExtractor` is the composed knowledge-soundness map.

## The escape arm and the failure decomposition

`actionLedgerEscapeFinder` returns the computed discrete-log relation data of the extracted bundle's first Sinsemilla escape, and `actionLedgerExtractor_eq_none_iff` decomposes failure of the composed extractor exactly: either the witness projection of the shared knowledge outcome returned nothing (the knowledge-failure event that the knowledge-error endpoint bounds), or a member escaped the bridge. The supporting characterization lives with the knowledge layer, universally quantified so that it applies at any elaboration of the shared outcome.

## Census housekeeping

The endpoint census script now accepts pin entries whose name wraps onto a continuation line, and it fails loudly on an entry that it cannot parse a valid name from — including a herestring fix for a guard that `pipefail` would otherwise skip silently through SIGPIPE when the pin list exceeds the pipe buffer.

A separate commit updates the Sinsemilla relation section's prose in `Zcash/TrustBoundary.lean`. That prose still attributed the relation entries' `+native` budget to the deployed bases' on-curve certificates. After #161 kernel-checked the fixed-base certificates, the only native owner that those entries reach is the Pallas point-count witness.

## Remaining after this PR

Review (thanks @TalDerei) found two soundness holes in the first composed experiment: it required validity at the programmed value bases while the extracted witnesses open the value commitments at the deployed bases, making its violation arm essentially empty, and its per-slot-and-size oracle tables did not model one random oracle. That experiment has been removed from this branch (preserved on the `fv/action-witness-bridge-pre-basis-rework` ref) and will be rebuilt in a follow-up PR: ledger validity and the violation events move to the deployed primitives; the conservation arm becomes a computed nontrivial discrete-log relation among the deployed value bases — symmetric with the Sinsemilla side — with the programmed-basis step (Jaeger–Tessaro) as an isolated terminal lemma on the standalone relation game; and the oracle tables are shared per bundle size. The sequential generalization of the proof-emitting adversary model is tracked with #214 and #155.

Progress on #147: the bridge machinery lands here, and #147 stays open for the games-side composition.

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Circuits/Integration/TopLevelCircuit.lean
+++ b/Zcash/Circuits/Integration/TopLevelCircuit.lean
@@ -1,6 +1,7 @@
 import Clean.Halo2.TopLevel
 import Zcash.Circuits.Integration.CircuitIntegration
 import Zcash.Common.RelationWitness
+import Zcash.Common.Satisfying
 
 /-!
 # Generic SNARK-to-top-level-circuit endpoint
@@ -36,15 +37,16 @@ theorem topLevelSoundness
 
 end FullCircuitSatisfaction
 
-/-- Type-valued semantic evidence for a top-level circuit.  Unlike
-`TopLevelCircuit.Statement`, this retains the private witness as executable data. -/
-structure TopLevelSemanticWitness
+/-- Type-valued semantic evidence for a top-level circuit: the private witness as
+executable data, together with its specification proof — `Zcash.Common.Satisfying`
+at the circuit's own specification.  Unlike `TopLevelCircuit.Statement`, nothing is
+truncated. -/
+abbrev TopLevelSemanticWitness
     {Config : Type} {PublicInput : TypeMap}
     [ProvableType PublicInput]
     (top : TopLevelCircuit Fp Config PublicInput)
-    (publicInput : PublicInput Fp) where
-  privateWitness : top.PrivateWitness
-  valid : top.Spec publicInput privateWitness
+    (publicInput : PublicInput Fp) : Type :=
+  Zcash.Common.Satisfying top.Spec publicInput
 
 namespace TopLevelSemanticWitness
 
@@ -56,7 +58,7 @@ theorem statement
     {publicInput : PublicInput Fp}
     (witness : TopLevelSemanticWitness top publicInput) :
     top.Statement publicInput :=
-  ⟨witness.privateWitness, witness.valid⟩
+  ⟨witness.w, witness.satisfied⟩
 
 end TopLevelSemanticWitness
 
@@ -130,9 +132,8 @@ def semanticWitness_or_bad
     TopLevelSemanticWitness top
       (top.extractPublicInput (top.environment assignment)) ⊕' Bad :=
   bindOrRelationWitness witness.bridge.satisfaction_or_bad fun hsatisfied =>
-    { privateWitness :=
-        top.extractPrivateWitness (top.placedEnvironment assignment)
-      valid := top.soundness assignment
+    { w := top.extractPrivateWitness (top.placedEnvironment assignment)
+      satisfied := top.soundness assignment
         (by
           rw [witness.operations_eq]
           exact FullCircuitSatisfaction.constraints
--- a/Zcash/Circuits/Integration/TopLevelCircuit.lean
+++ b/Zcash/Circuits/Integration/TopLevelCircuit.lean
@@ -39,7 +39,7 @@ end FullCircuitSatisfaction
 
 /-- Type-valued semantic evidence for a top-level circuit: the private witness as
 executable data, together with its specification proof — `Zcash.Common.Satisfying`
-at the circuit's own specification.  Unlike `TopLevelCircuit.Statement`, nothing is
+at the circuit's own specification. Unlike `TopLevelCircuit.Statement`, nothing is
 truncated. -/
 abbrev TopLevelSemanticWitness
     {Config : Type} {PublicInput : TypeMap}
--- a/Zcash/Circuits/Integration/TopLevelCorrectness.lean
+++ b/Zcash/Circuits/Integration/TopLevelCorrectness.lean
@@ -109,11 +109,11 @@ def of_publicInputEncoding
       pp.numProofs proofIndex :=
     { polynomial := poly }
   refine
-    { privateWitness := (witness proofIndex).privateWitness
-      valid := ?_ }
+    { w := (witness proofIndex).w
+      satisfied := ?_ }
   rw [← assignment.extractPublicInput_eq
     (inputs proofIndex) (hencoding proofIndex)]
-  exact (witness proofIndex).valid
+  exact (witness proofIndex).satisfied
 
 /-- Forget the retained decoded witnesses and recover the ordinary bundle statement. -/
 theorem statement
--- a/Zcash/Common/Satisfying.lean
+++ b/Zcash/Common/Satisfying.lean
@@ -2,7 +2,7 @@
 # A satisfying witness, as data
 
 `Satisfying R x` bundles a witness with the proof that it satisfies the relation `R`
-at the instance `x`.  It is Type-valued: under Curry–Howard the witness is data, and
+at the instance `x`. It is Type-valued: under Curry–Howard the witness is data, and
 that is what keeps knowledge-soundness statements constructive — `Classical.choice`
 can conjure the truncated `∃ w, R x w` from mere soundness, but not this structure.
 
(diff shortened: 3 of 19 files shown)


Zcash/Security/Ledger/OrchardExtractionExperiment.lean
```diff
@@ -0,0 +1,461 @@
...
+The composed statement is for the reprogrammed basis: the extracted witnesses open the
+value commitments at the deployed bases, and the sampled bases stand for those under
+the reference-string idealization recorded at
+`IdealizedKSBalanceAdversary.violationEvent`.
+
+`hks` is stated so that `orchard_action_adaptiveStatement_knowledge_error_bound`
+applies to it verbatim, one application per slot-size pair; `hescape` names each
+pair's Sinsemilla-escape bound. The remaining hypotheses and the second summand are
+the KS-idealized integrity endpoint's, applied at the constructed adversary. -/
+theorem orchardBalanceIntegrityExtraction_measure_le
```
TalDerei: there's a subtle problem now that we're doing this bridging / composition. the circuit side and game side reason about different basis for `cv_net` (they're properly fixed in circuit, but randomly sampled in game which was fine in isolation before this bridging existed).

Zcash/Security/Ledger/OrchardExtractionExperiment.lean
```diff
@@ -0,0 +1,461 @@
...
+    (hescape : ∀ (t : Fin k) (n : Fin maxActions),
+      (runLawAt A t n).toOuterMeasure (runEscape H_bind A t n) ≤ ε_escape)
+    (hsin : (toIdealizedKS H_bind A hqb halg).experiment.toOuterMeasure
+      ((toIdealizedKS H_bind A hqb halg).sinsemillaRelationEvent issuance maxActions k)
+      ≤ ε_sinsemilladlr)
+    (hdl : ∀ j, TextbookDLWithCoinsAdvantageLE pallasGen
+      ((toIdealizedKS H_bind A hqb halg).conservationFinder k j) ε_dl) :
+    (toIdealizedKS H_bind A hqb halg).experiment.toOuterMeasure
+        (extractionFailureEvent H_bind A ∪
+          (toIdealizedKS H_bind A hqb halg).violationEvent issuance maxActions
```
TalDerei: ^ check out this line

Zcash/Security/Ledger/OrchardExtractionExperiment.lean
```diff
@@ -0,0 +1,461 @@
...
+The bundle size is selected adaptively. The deployed verifier absorbs the verifying
+key digest, every instance commitment, and every advice commitment before squeezing
+its first challenge, so the pre-challenge transcript determines the bundle size and
+distinct sizes never share a squeeze input. Sampling an independent transcript per
+slot and size is therefore a sound finite presentation of one random oracle, and the
+ledger machine may select each position's size after seeing the runs. The cost is
+that the knowledge-soundness terms of the composed bound sum over the slot-size
+pairs — a factor `k * maxActions` — and each per-size family carries its own query
+budget where a real adversary's attempts share one. Removing the factor is tracked
+as #214.
```
TalDerei: I think there's a spec bug here in the code? oracle tables should share tables across same size slots instead of sample per `(slot, size)`.

```ts
Sampling one oracle table per size, shared across all slots of that size, is therefore a sound finite presentation of one random oracle.

TalDerei:
### Soundness holes

**High — per-slot tables do not model one random oracle**

`Runs` samples an independent oracle table per `(slot, size)`. Same-size slots share the same transcript domain, so the same query may receive different answers—impossible with one random oracle. Share each size’s table across slots, and prove the per-size domains are disjoint or use one global table.

**High — not a general proof-emitting adversary**

`families` fixes one standalone proof producer per `(slot, size)`, while `B` only selects requests from the sampled runs. This excludes a stateful adversary whose later proofs depend on shared coins, prior proofs, or oracle history. Model one sequential proof-emitting machine or prove that every such adversary embeds into this fixed-grid model.

with the help of claude opus

Zcash/Security/Ledger/OrchardExtractionExperiment.lean
daira: That's quite significant. I will have a think how to fix it.

daira:
Thanks — all three points are right, and the `cv_net` one is the significant hole. As landed, the violation arm of the composed event requires validity at the sampled primitives while the extracted annotations satisfy `cv_net` at the deployed bases, so that arm is essentially empty and the composed endpoints bound almost nothing.

The fix we are taking, in this PR: move the basis programming out of the games entirely. Validity and the violation events move to the deployed primitives. The conservation arm then outputs a nontrivial discrete-log relation among the deployed value bases {𝒱, ℛ} — computed from the binding-signature algebra and the announced representations, exactly parallel to how the non-negativity side already works at the fixed Sinsemilla bases. The programmed-basis step (Jaeger–Tessaro, already re-proved as `relation_prob_le_of_textbookDL`) becomes an isolated terminal lemma relating that fixed-basis advantage to textbook discrete log, applied to the standalone relation game where there is no circuit to disagree with. The asymmetry between the two sides was the latent bug.

On the oracle tables: agreed, and the fix is cheap — `Coins` is determined by the proof parameters alone, so one table per size, shared across all slots of that size, typechecks directly; same-size slots then answer identical queries identically, and cross-size independence rests on the pre-challenge absorption argument in the module doc. Nothing in the marginals or the union bound used cross-slot independence.

On the fixed grid: also real. The honest fix — one sequential proof-emitting machine against one oracle, with per-accepted-proof extraction — converges with the reduction-layer surgery in #214 and the ε-slot oracle-machine layer in #155, so we will state the restriction prominently in the model doc here and do the sequential generalization as that work.

daira:
Correction to the plan above: the deployed-bases rebuild moves to a follow-up PR rather than this one. This PR now carries only the basis-independent layers — the Action-to-ledger bridge, the adaptive-family extractor composition, and the escape-arm decomposition — with the first composed experiment removed and preserved on the `fv/action-witness-bridge-pre-basis-rework` ref. #147 stays open for the games-side composition.

TalDerei (approved):
reACK

daira (commented):
Self-ACK

daira (commented):
Self-reACK

PR #212, opened 2026-08-26
ci: Always report the zizmor check

daira:
`zizmor` is a required status check, but its workflow was path-filtered to workflow and action files, so a pull request touching neither never instantiated the check and could not satisfy branch protection. This trigger gap is why recent PRs sat with the check "Expected".

The fix mirrors the Lean, book, and fixtures workflows: the workflow now triggers unconditionally, a `Detect workflow-relevant changes` job computes the path filter inside the run with the same pinned `dorny/paths-filter` action, and the audit job is gated on its output. A skipped audit satisfies the required check, and the detect job is reported on every pull request. The audit steps themselves are unchanged, including the `min-severity: informational` policy.

After this merges, `Detect workflow-relevant changes` should be added to the required status checks beside the existing `zizmor` entry, matching the other detect jobs.

🤖 Claude Fable 5

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,27 +1,47 @@
 name: Workflow audit
 
+# Always triggers, so the required `zizmor` check is always reported (a
+# workflow skipped by `on.paths` would leave a required check pending and
+# block merges).  The audit is instead gated on a path filter computed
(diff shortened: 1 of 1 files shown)


daira (commented):
Self-ACK

TalDerei (approved):
ACK thanks for fixing

PR #213, opened 2026-08-26
ci: Add the Workflow CI aggregate check

daira:
Follow-up to #212. That fix made the workflow trigger unconditionally, but the `zizmor` job itself is still conditional on the computed path filter, so a skipped audit only satisfies branch protection through GitHub's skipped-counts-as-passed semantics, and a failure of the detect job would skip the audit silently. The other workflows solve this with a single stable aggregate job (`Lean CI`, `Ironwood book CI`, `Fixtures CI`) that always runs, gates on the detect job's result, and passes exactly when the conditional job passed or was correctly skipped.

This adds the missing counterpart: a `Workflow CI` job with the same gating logic, so the workflow-audit workflow has one always-reported check suitable for branch protection. `Workflow CI` should replace `zizmor` in the required status checks.

🤖 Claude Fable 5

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,8 +1,8 @@
 name: Workflow audit
 
-# Always triggers, so the required `zizmor` check is always reported (a
-# workflow skipped by `on.paths` would leave a required check pending and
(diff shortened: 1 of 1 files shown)


daira (commented):
Self-ACK

PR #215, opened 2026-08-26
feat(fv): model Halo2 transcript bytes and proof encoding

TalDerei:
Related to https://github.com/zcash/halo2/pull/933 and https://github.com/zcash/orchard/pull/544. Closes #66. 

Visualization in this claude artifact https://claude.ai/code/artifact/aea8fef0-9637-4416-9db7-28dfafc83235.

**Before:**  trust that `blake2b_simd` is a correct BLAKE2b, and idealize that hash function as a RO.

**After:**  idealize [RFC 7693](https://datatracker.ietf.org/doc/html/rfc7693.html) (the BLAKE2b design) in Lean as the RO; check `blake2b_simd` against the design at the captures (specifically run over the transcript bytes, and check it reproduces every challenge that `blake2b_simd` produced in the captured runs).

Proving `blake2b_simd` would mean extracting the Rust (including SIMD paths) into Lean and proving it equal to the RFC formalization for all inputs which we don't do. Faults in `blake2b_simd` that affect the captured transcript executions are now in scope; this does not establish agreement on arbitrary inputs.

-------------

This connects ironwood’s verifier model to the actual bytes consumed by Halo2. So the modeled proof / challenges are derived from the same bytes Halo2 consumed, rather than being trusted as separate captured values.

  - Models Halo2’s BLAKE2b transcript and canonical proof-byte decoder/serializer.
    - `Halo2 proof bytes --> decoded points/scalars --> transcript bytes --> BLAKE2b --> FS challenges`.
  - Checks that captured bytes decode to the captured typed proof and reproduce every captured challenge.
  - Proves Fiat–Shamir query separation across different Orchard action counts.

Issue #214, opened 2026-08-26
Compose with the SNARK argument at the reduction layer to remove the k*maxActions factor


daira:
The extraction experiment (`Zcash/Security/Ledger/OrchardExtractionExperiment.lean`) consumes the Action knowledge-error endpoint (`orchard_action_adaptiveStatement_knowledge_error_bound`) as a black box. That endpoint is stated over its own sample space. The composed bound therefore transfers it by marginalization, and it covers the adversary's per-slot bundle-size choice by a union bound. As a result, the knowledge-soundness terms of `extractionFailureEvent_measure_le` carry a factor `k * maxActions`, consensus-instantiated below `2^16`. The factor applies to the whole knowledge-error formula, including its discrete-log advantage term — and there a `2^16`-scale factor costs real bits of security.

The factor is the honest count of extraction targets in the current model, but it is avoidable by composing with the SNARK argument at the reduction layer instead of treating its endpoint as a black box. Sean Bowe and Daira-Emma Hopwood independently reached the same conclusion; the black-box route cannot avoid it.

Plan sketch:

1. The SNARK argument's set-level containments (`adaptiveStatementKnowledgeFailureEvent_subset`: failure is contained in the relation event together with the statistical surfaces) are distribution-free and survive unchanged.
2. Build one combined relation finder over the composed experiment that replays every slot-size run and returns the first relation any of them yields — the `conservationRelFinder` pattern — so a single discrete-log advantage appears with no factor.
3. The statistical surfaces keep a per-slot union bound, but the factor then multiplies only `1/|F|`-scale terms, where `k * maxActions * 2^-100`-scale losses are harmless.
4. The surgery: restate the SNARK argument's probability lemmas parametrically in the sample law, or directly at the composed law.

A related looseness that the same rework could address: each per-size family carries its own query budget, where a real adversary's attempts share one budget.

🤖 Claude Fable 5

PR #219, opened 2026-08-28
fv: move the ledger value side to the deployed bases

### LINKED ISSUE #147: Compute the circuit-to-ledger Action witness bridge

daira:
The deployed-bases rework (following the review findings on #211), which closes #147: the κ layer is generalized over the presented basis so that ledger validity and the violation events can be stated at the deployed value bases rather than the sampled ones.

Review suggestion: start with `Zcash/Security/Ledger/OrchardExtractionExperiment.lean`. Pay particular attention to its `orchardBalanceIntegrityExtraction_measure_le_of_dlogProfiles` which is the new fully composed Balance integrity capstone.

## Contents

**Basis generalization (pure refactor).** The RedDSA discharge layer (`dischargeOutAt`, `effectiveRepAt`, `VerifiesAt`, `kappaEventAt`, and `badFiberAt` with its measure bound) and the ledger κ apparatus (`primitivesAtBasis`, `shapeAtBasis`, `bindingAtBasis`, `extractorAtBasis`) become `At`-forms over an arbitrary presented basis, with the sampled forms redefined as their definitional instantiations at `scalarBasis gen logs`, so consumers are unchanged. At the deployed pair, the basis-parametric value commitment is definitionally the deployed Pedersen commitment — exactly what the rework's validity events need.

**κ documentation.** The knowledge error κ is now defined where it is introduced and glossed at each point of first contact, as it was easy to meet the term before its definition.

**One relation type.** `NontrivialRelation g V` is now `AlgebraicRelationWitness` at the augmented basis of a generator family `g` and a vector `V` of distinguished points, so the two-point form is spelled `NontrivialRelation g ![U, W]`, the one-point Sinsemilla form is `NontrivialRelation g ![U]`, and the `NontrivialRelationOne` family disappears into an instantiation. The classical coefficient shapes survive as constructors, consumers read the distinguished coefficients through thin `α`/`β` accessors instead of per-shape structure fields, and the conversion that became the identity is deleted along with the consumer-less conversions and their census pins. `AugmentedIndex` and `AugmentedRelationWitness` survive as abbreviations, so the circuit-facing SNARK signatures are unchanged. Every basis family in the development has a finite index —a `Fin` family, or group-hash inputs of concretely bounded length— so total coefficient vectors suffice. The break-producing reductions read coefficients at the slots they name rather than enumerating an index type, so relation building costs nothing in efficiency; only the known-log dischargers sum over the finite index. The module split anticipates the planned layer separation: the relation vocabulary, the representation types, and the known-log dischargers live in `Common/DiscreteLogRelation`, and the programmed-basis adapter stands alone in `Common/ProgrammedBasis`; nothing stated at the deployed bases assumes it.

**Fixed-basis κ bound.** The knowledge error is bounded at a single presented basis (`kappaEventAt_measure_le`). Both arms live over the challenge-table factor, so a plain union bound composes the per-basis squeeze with a named hypothesis on the relation arm — the probability that the replayed finder returns a relation at that basis. No basis is sampled, so the bound holds at any fixed basis at `(qH+1)/#F` above the named advantage.

**Deployed conservation and cap experiments.** The conservation composition runs at a fixed presented basis (`balanceConservationBefore_measure_le_experimentAt` and the cap sibling), and the Orchard endpoints instantiate it at the deployed value bases (`orchardValueBases`, built from the named `valueCommitV` and `valueCommitR`). Validity is then at the deployed value commitment —definitionally the deployed primitives on the value side— and only the binding challenge hash remains idealized as the table, so no reference-string heuristic is involved. The named advantage `ε_valuedlr` bounds the exhibited deployed finder, which returns its relations over the combined deployed basis's value slots; the bound is `ε_valuedlr + (qH+1)/#F` — one `1/#F` below the sampled form, because the relation arm is assumed rather than discharged through the Jaeger–Tessaro step. The adversary model's algebraicity assumption is generalized to every presented basis, which the sampled endpoints consume at their sampled bases and the deployed ones at `orchardValueBases`.

**Claim precision on the #147 gap.** The violation events' critical-gap bullets are updated through the PR. Once the composed experiment lands, the events' docstrings are retold: the annotation idealization is discharged as data on the composed route, the critical-gap framing goes, and the deployed event's docstring states only its differences from the sampled one.

**Deployed integrity.** The integrity composition also runs at a fixed presented basis (`balanceIntegrityBefore_measure_le_experimentAt`, with the combined non-negativity arm event factored out as `balanceSubsetArmsUpTo`). The Orchard non-negativity transport generalizes over the presented basis —the record update never touches the fields the three arms read— so the deployed endpoint `orchardBalanceIntegrity_measure_le_idealizedks_deployed` bounds an integrity violation by ε_sinsemilladlr + (ε_valuedlr + (qH+1)/#F) at the deployed value bases.

**The composed extraction experiment, rebuilt.** `OrchardExtractionExperiment` is restored from the pre-rework ref with the redesign @TalDerei's review of #211 called for. The oracle tables are shared per bundle size — one transcript per size, read by every chain position's family at that size. That is a faithful presentation of one typed oracle exactly when the schedules of different sizes never squeeze the same transcript; the module doc states this as a modeling caveat resting on the separation across action counts, whose typed level this PR proves (`ActionCount.lean`) and whose byte level #215 proves. The composed endpoints run at the deployed value bases, so the extracted witnesses open the value commitments at exactly the bases where validity is stated and the violation arm is inhabited. The bounds are (k·maxActions)(ε_ks + ε_escape) plus the deployed endpoints' terms, with the k·maxActions factor tracked for removal as #214. The isolated Jaeger–Tessaro terminal step (`valueRelationWithCoins_prob_le_of_textbookDL`) states once, on the standalone programmed relation game, what ε_valuedlr costs against textbook discrete log, but the deployed experiments do not take that step.

**The knowledge hypotheses discharged.** The composed endpoints' per-arm hypotheses are gone: the `_of_dlogProfiles` endpoints apply the adaptive-statement knowledge capstone once per slot-size pair, taking a DLOG profile per pair, and instantiate every other arm's ε at the constructed adversary's own advantage, so no per-arm hypothesis remains. With this, the Balance integrity node on the one-picture diagram joins the machine-checked class, and its remaining quantitative wart is the k·maxActions factor (#214). The census budgets for the qualified endpoints disclose the capstone's native-checked verifying-key certificates alongside the curve certificates.

**The combined deployed basis.** Every relation arm now lands in one type: a nontrivial discrete-log relation among the Sinsemilla generator table and the eight combined deployed points (`orchardPoints`, with named `OrchardPointIndex` slots, including the nullifier base 𝒦^Orchard at `idxNullifierK`). The DLR reducers place their coefficients at their sites' slots, the escape classifier's `ActionDLBreak` loses its site-dependent vector, the value finder embeds at the value slots, and `OrchardBalanceRelation` is deleted — its three constructors carried payloads of what is now one type. The supporting operation is `AlgebraicRelationWitness.embed`, zero-extension of a relation witness along a slot injection presented with a `Function.IsPartialInv` partial inverse; it supersedes `reindex`.

**One relation term.** The composed endpoints now conclude at (k·maxActions)·`worstKnowledgeError` + (`combinedDLRAdvantage` + (qH+1)/#F). The per-pair escapes, the Sinsemilla reducer's relations, and the value-DLR finder's are one event over the composed deployed experiment, so the k·maxActions factor multiplies only the knowledge arm, and the challenge-oracle squeeze is the only counting term outside the combined advantage. The recombination happens at the set level: the conservation and cap experiments' per-coin containments are extracted as named lemmas that the experiment theorems now consume.

**Model refinements from review.**
* The query bound is a field of the proof-emitting adversary: `queryBound` says the ledger machine respects its own `qH` budget, and the assembled machine's bound is derived, so the `hqb` hypothesis disappears from every endpoint.
* `TxRequest` counts actions (`numActions`, nonzero and capped) with the selected family slot derived rather than supplied.
* `OrchardCapstone`'s module doc points readers at the composed capstones' home.

**The book retold.**
* The ledger-security-games hand-off story, the proof map's composition node, the guide, the source map, and the definitions page now present the composed extraction experiment as the formalized hand-off. The remaining openness is stated per route: quantitative on the Balance side (the per-slot-and-size knowledge-soundness terms, #214), and the annotation assumption still in force for Spendability and Spend Authority (#155).
* The one-picture diagram no longer draws the random oracle as a target: the heavy purple edges now read as reductions in the AGM+RO with a challenge oracle that the reduction may program, and the birthday count for `CollisionUpToSign` —the one purely random-oracle bound on the page— moves into the prose (extracted from the diagram improvements drafted in #217).
* A book housekeeping commit avoids literal square brackets in TeX math, which mdbook-linkcheck misreads as link syntax (upstream: Michael-F-Bryan/mdbook-linkcheck#99).
* The one-picture diagram merges the former (𝒱, ℛ) and Sinsemilla relation nodes into one combined-deployed-basis (Pallas) node. It takes some cosmetics anticipating #217: hexagons for the relation, discrete-log, and collision-resistance nodes, the darker hypothesis red for contrast with the white text, and per-curve labels on the group-hash edges.
* The knowledge-soundness node joins the machine-checked class in the one-picture diagram; its circuit-correctness conditions have been discharged in the tree. The verification chain is spelled out in [a review comment](https://github.com/zcash/ironwood/pull/219#issuecomment-5536526193). This completes the rework's planned units.

## Review response

Commits after the reviewed head address [Tal's review](https://github.com/zcash/ironwood/pull/219#issuecomment-5565217354):
* Every reduction producer introduced with the deployed bases has a direct `assert_computable` pin: the deployed finders, the break transports to the fixed primitives, the basis-parametric Balance-subset reducer, the combined-basis embedding, the basis-parametric extractor, and the Common-layer relation constructor, embedding, and U/W dischargers.
* κ is described as a bound on the extraction-failure probability, obtained as the measure of the pivot event; the relation module's doc says which reductions enumerate the index; the docstrings that called the witness annotations a gap tracked as #147 describe the composed route that discharges them.
* The extraction experiment's module doc states the per-size Fiat–Shamir presentation as a modeling caveat, naming the separation across action counts it rests on.
* The chain-collision reducer is generalized over a vector of blinding points (`relationOfChainVecPmEq`), with the one-point reducer as its instance. The recovery of a note from its Sinsemilla message words, which the note-commitment reducer performed inline, is one shared lemma, and the Poseidon nullifier scalar has a name in the Pool model.
* The combined deployed basis has a 𝒦^Orchard slot, and `relationOfNullifierCollision` computes a nullifier collision's relation at the deployed Orchard primitives, applying the generalized reducer at the randomness and nullifier bases. `NullifierCollision` carries the 64-bit value bounds like `NoteCommitBreak` (the model's `Note` holds the value as an unbounded natural number), and the games page's nullifier edge links the deployed reducer.
* The value bounds carried by the break structures and the Action statement are described as the note value's 64-bit type, which the model's `Note` holds as an unbounded natural number; `Primitives.valueBound` has a docstring.
* The ledger layer calls the K-bit groups of a Sinsemilla input words, as halo2_gadgets does and the protocol specification does as of zcash/zips#1358, rather than chunks; the identifiers it defines follow (`wordIdx`, the `_words_onCurve` lemmas), while circuit-layer identifiers keep their names pending a wider rename.
* The typed-level separation of Fiat–Shamir queries across action counts (`Soundness/FiatShamir/ActionCount.lean`, ported from #215 without its byte-level half) is in the tree and pinned, and the extraction experiment's module doc cites it as the fact its per-size tables rest on.

Outstanding: the byte-level half of the action-count separation (#215), its consumption in the experiment's sample space (#224), and widening the binding-signature AGM hypothesis to the combined basis (#188).

🤖 Written with the assistance of Claude Fable 5, Fable 5.1, and Opus 4.8.


Zcash/Snark/Soundness/AGM/Adapter.lean
```diff
@@ -1,11 +1,12 @@
...
 import Zcash.Snark.Soundness.Deployed.Binding
 
 /-!
 # The URS as an AGM basis
 
-The relation types and the relation-to-discrete-log reductions are model-free and live in
-`Zcash.Common.AlgebraicRelation`; this module is the AGM-side entry point to them. It carries the
+The relation types and the known-log relation-to-discrete-log dischargers are model-free
+and live in `Zcash.Common.DiscreteLogRelation`; this module is the AGM-side entry point to
+them. It carries the
```
daira: Short line.

daira (commented):
Self-ACK

daira:
For other reviewers: here is the verification chain behind flipping the knowledge-soundness node to the machine-checked class on the Ledger Security Games page. The node's remainder was "knowledge soundness's circuit-correctness conditions", and those are now discharged in the tree.

- The composed experiment's `hks` hypothesis is stated so that `orchard_action_adaptiveStatement_knowledge_error_bound` applies to it verbatim, one application per slot-size pair (see the docstring of `orchardBalanceIntegrityExtraction_measure_le` in `Zcash/Security/Ledger/OrchardExtractionExperiment.lean`).
- That capstone (`Zcash/Snark/Capstones/Action.lean`) is generic in `numProofs` and takes only `hB : B ≠ 0`, `hquery : Function.Injective query`, and the DLOG profile. The pair-count side condition is discharged by the proven generic theorem `adaptiveStatement_pairCount_lt` (kernel-checked, via `norm_num`), and the statistical-surface budgets by the pinned proven `_measure_le` surface lemmas, whose `+native(…)` budgets carry only the standard curve certificates.
- Neither `DeployedConstraintStaticChecks` nor `DeployedConstraintXSqueezeSchedule` appears in the adaptive-statement lane. Those interfaces sit on the straight-line-constraint consensus-maximum package, a different set of endpoints from the ones that the ledger composition consumes; where there is a capture, they are discharged by `native_decide` (`Zcash/Snark/Fixtures/MultiAction/Honest/StaticChecks.lean`).
- On the `+choice` budgets of the extractor pins: the census block in `Zcash/TrustBoundary.lean` documents the pattern — the break data is "coefficients over the public basis … terms of the transcript, not an inhabitant chosen from an existence proof", so `Classical.choice` sits in erased proof positions while `assert_computable` certifies the executable path.

The remaining amber/brown items in the page's legend are RedDSA unforgeability, and the k·maxActions factor of the composed extraction bound (#214).

book/src/formal-verification/guide.md
```diff
@@ -81,13 +81,15 @@ Zcon3 in [Understanding the Security of Zcash](https://raw.githubusercontent.com
...
-security arguments are not yet tied into the Action circuit knowledge soundness proof
-at all: the games state their circuit premiss over their own abstract types, and no
-definition is currently shared between the two developments
-([#147](https://github.com/zcash/ironwood/issues/147) and
-[#155](https://github.com/zcash/ironwood/issues/155)). Additional caveats are stated
+unbalanced consensus-valid ledger). All the important pieces have been completed,
+and the ledger security arguments are now tied into the Action circuit knowledge
+soundness proof: the composed extraction experiment annotates the ledger games' chains
+with the witnesses the circuit's extractor computes, at the deployed value bases
+([#147](https://github.com/zcash/ironwood/issues/147)). What remains on that route is
```
daira: #147 doesn't need to be mentioned because it will be closed by this PR.

book/src/formal-verification/proof-map-embed.html
```diff
@@ -409,8 +409,8 @@ <h2>Legend</h2>
...
-    anchor:'ActionBundleWitness · Security.Ledger.Bridge.actionSpec_to_ledger',
-    role:'The output-side composition gap. actionSpec_to_ledger already refines one ActionSpec witness to the games-facing ledger statement or an exhibited Sinsemilla break, and actionCircuit_spec_iff links the extractor’s circuit spec to ActionSpec. What is not yet exported is the theorem mapping every entry of an extracted ActionBundleWitness through those bridges to the bundle-level ledger relation.' },
+    anchor:'ActionBundleWitness · Security.Ledger.Bridge.actionSpecToLedgerData · OrchardExtractionExperiment',
+    role:'The output-side composition, formalized for Balance: actionSpecToLedgerData refines each extracted witness to the games-facing ledger data or a computed Sinsemilla break, bundleLedgerData traverses the extracted bundle, and the composed extraction experiment assembles the annotated chain at the deployed value bases. The floor that remains: the annotated capstones still quantify over supplied annotations, and the composed route carries per-slot-and-size knowledge-soundness hypotheses at a k·maxActions factor (#214).' },
```
daira: "the annotated capstones still quantify over supplied annotations" is not a floor; it's just how that layer is supposed to work, and is tautological. "the composed route carries per-slot-and-size knowledge-soundness hypotheses at a k·maxActions factor (#214)." also isn't a floor, it's just a suboptimal reduction.

book/src/formal-verification/source-map.md
```diff
@@ -25,13 +25,13 @@ directory level, naming the notable modules as entry points.
...
-  (Jaeger–Tessaro, [Expected-Time Cryptography: Generic Techniques and Applications to Concrete
+  `DiscreteLogRelation` carries a nontrivial `F`-linear (discrete-log) relation over an arbitrary
+  indexed basis (`AlgebraicRelationWitness`) as *computed data* — the coefficients — so the
+  reduction-style security arguments can produce a break rather than merely assert one exists (see
+  *breaks as computed data* on the [Formal Verification](../formal-verification.md) page), and
+  turns one into a discrete log against known slot logs. `ProgrammedBasis` turns a relation into a
+  discrete log against a basis programmed from the DL challenge (Jaeger–Tessaro, [Expected-Time
+  Cryptography: Generic Techniques and Applications to Concrete
   Soundness](https://eprint.iacr.org/2020/1213), Lemma 3); `RelationProbability`
   and `RelationProbabilityCoins` price that reduction's single miss hyperplane at `1/|F|`, and
```
daira:
```suggestion
  discrete log against a basis programmed from the DL challenge (Jaeger–Tessaro,
  [Expected-Time Cryptography: Generic Techniques and Applications to Concrete Soundness](https://eprint.iacr.org/2020/1213), Lemma 3);
  `RelationProbability` and `RelationProbabilityCoins` price that reduction's single miss hyperplane at `1/|F|`, and
```

book/src/formal-verification/source-map.md
```diff
@@ -526,3 +527,9 @@ computed data.
   against the binding-signature layer, `KeyBindingArm` discharges the key-binding ε in the oracle
   model, `Capstone` lifts the deterministic layer to a distribution over valid annotated ledgers,
   and `Completeness` checks the other direction — that an honest wallet's spend actually verifies.
+  The experiment lane places those capstones in the challenge-oracle model:
+  `ConservationExperiment` and `IntegrityExperiment` compose the arms in one sample space — at the
+  sampled bases and, in their `At` forms, at the deployed value bases — `OrchardIntegrityExperiment`
+  instantiates them at the deployed choices, and `OrchardExtractionExperiment` builds the annotated
+  adversary from a proof-emitting one, annotating its chain with the Action circuit's extracted
+  witnesses.
```
daira: This sentence is a bit long

Zcash/Security/Ledger/OrchardExtractionExperiment.lean
```diff
@@ -0,0 +1,634 @@
...
+  orchard_action_adaptiveStatement_knowledge_error_bound)
+open Zcash.Snark.Keygen (actionProofParamsFor)
+
+/-- The transaction-level data the ledger machine supplies for one chain position: the
+selected bundle size and the declared transaction fields. The action instances and
+witnesses are not requested — they come from the selected slot's extracted ledger
+data. -/
+structure TxRequest (MSG : Type) (maxActions : ℕ) where
+  /-- The selected bundle size: the transaction carries `size + 1` actions. -/
+  size : Fin maxActions
```
daira: Would it be more natural to have `size` be the actual number of actions with a `size_nonzero : size ≠ 0` field? It could then be renamed to `numActions`. (Nonblocking.)

Zcash/Security/Ledger/OrchardExtractionExperiment.lean
```diff
@@ -0,0 +1,634 @@
...
+  ι : Type
+  /-- The coin distribution. -/
+  coins : PMF ι
+  /-- Per slot and bundle size, the family producing that proof. -/
+  families : Fin k → (n : Fin maxActions) →
+    ComputedAdaptiveActionStatementFSFamily (actionProofParamsFor (n.1 + 1))
+  /-- The ledger machine: from the coins, the sampled runs, and the presented bases,
+  request one transaction per chain position, with its announced binding
+  representation. -/
+  B : ι → (table : ↥(Set.range query) → VestaG) →
```
daira: `B` is an uninformative name. Is it used elsewhere or just here?

Zcash/Security/Ledger/OrchardExtractionExperiment.lean
```diff
@@ -0,0 +1,634 @@
...
+(`deployedViolationEvent`), so no reference-string heuristic is involved on the value
+side.
+
+`hks` is stated so that `orchard_action_adaptiveStatement_knowledge_error_bound`
+applies to it verbatim, one application per slot-size pair; `hescape` names each
+pair's Sinsemilla-escape bound. The remaining hypotheses and the second summand are
+the deployed KS-idealized integrity endpoint's, applied at the constructed
+adversary. -/
+theorem orchardBalanceIntegrityExtraction_measure_le
+    (hqb : ∀ j b, (toLA H_bind A j b).QueryBound A.qH)
```
daira: It's not clear how to interpret this bound being here. A reader would have to chase up what it means.

Zcash/Security/Ledger/OrchardExtractionExperiment.lean
```diff
@@ -0,0 +1,634 @@
...
+    (halg : ∀ j, ∀ basis, AlgebraicAtBindingPointsAt 2 0 1 orchardQueryOf
+      (primitives spendAuthVerify (redPallasBindingVerify H_bind)) id basis
+      (toLA H_bind A j))
+    (issuance : ℕ → ℕ) (hmax : maxActions < 2 ^ 16)
+    {ε_ks ε_escape ε_sinsemilladlr ε_valuedlr : ℝ≥0∞}
+    (hks : ∀ (t : Fin k) (n : Fin maxActions),
+      (runLawAt n).toOuterMeasure (runKnowledgeFailure A t n) ≤ ε_ks)
```
daira, line 439: Isn't this supposed to be composed? We want to bound the advantage of an adversary against Balance integrity when the SNARK is Halo 2 on Vesta at the deployed bases and circuit; not the advantage for an arbitrary SNARK with a given KS error.

Zcash/Security/Ledger/OrchardExtractionExperiment.lean
```diff
@@ -0,0 +1,634 @@
...
+    (hescape : ∀ (t : Fin k) (n : Fin maxActions),
+      (runLawAt n).toOuterMeasure (runEscape H_bind A t n) ≤ ε_escape)
+    (hsin : (toIdealizedKS H_bind A hqb halg).deployedExperiment.toOuterMeasure
+      ((toIdealizedKS H_bind A hqb halg).deployedSinsemillaRelationEvent issuance
+        maxActions k)
+      ≤ ε_sinsemilladlr)
+    (hrel : ∀ j, (PMF.uniformOfFintype (OrchardQuery MSG → Fq)).toOuterMeasure
+        {table | ((toIdealizedKS H_bind A hqb halg).valueDLRFinder k j table).isSome}
+      ≤ ε_valuedlr) :
```
daira: I expected these also to be composed. The other endpoints and the `_of_dlogProfiles` variants are the same.

daira (commented):
Self-ACK

TalDerei:
Spotted some gaps and preflight skill (https://github.com/zcash/ironwood/tree/main/.claude/skills/pr-preflight) violations.

---------

At reviewed head `6324046`, I’d **request changes** for these findings:

1. **Soundness coverage gap — oracle composition; preflight §§4–5.** Independent tables can answer the same adversary query differently at different bundle sizes. Proof generation also lacks history from other sizes, so the experiment does not establish coverage of general deployed attackers. [Source](https://github.com/zcash/ironwood/blob/6324046ba5bf6970fe2cd1009d1353831b0a41c6/Zcash/Security/Ledger/OrchardExtractionExperiment.lean#L115).

2. **Soundness coverage gap — restricted representation basis; preflight §4.** Binding-signature labels require representations over only `V/R`, although attackers know other public Orchard points. Translating a representation of `noteQpt` into that basis already produces a nontrivial relation; the missing adapter cannot be assumed free. [Source](https://github.com/zcash/ironwood/blob/6324046ba5bf6970fe2cd1009d1353831b0a41c6/Zcash/Security/Ledger/OrchardExtractionExperiment.lean#L627).

3. **Preflight §3 violation — missing computability guard.** `valueDLRFinder` is currently computable but lacks its required `assert_computable` pin. Changing it to `noncomputable def` still passes the full build; adding the direct guard detects this regression. [Source](https://github.com/zcash/ironwood/blob/6324046ba5bf6970fe2cd1009d1353831b0a41c6/Zcash/Security/Ledger/OrchardIntegrityExperiment.lean#L464).

4. **Specification/diagram error — nullifier reduction target; preflight §5.** The diagram sends nullifier collisions to the combined deployed basis, but the actual reducer includes a nullifier `K` coefficient absent from that target. The diagram needs a separate target or a formalized transport. [Source](https://github.com/zcash/ironwood/blob/6324046ba5bf6970fe2cd1009d1353831b0a41c6/book/src/formal-verification/ledger-security-games.md#L66).

5. **Specification/documentation errors; preflight §§5 and 8.** κ is described as exactly extraction failure, although its event also includes successful extraction, and the claim that reductions never enumerate indices is false. Stale links and conflicting statements about closing #147 also need correction. [κ definition](https://github.com/zcash/ironwood/blob/6324046ba5bf6970fe2cd1009d1353831b0a41c6/Zcash/Security/RedDSA/KnowledgeError.lean#L157), [index enumeration](https://github.com/zcash/ironwood/blob/6324046ba5bf6970fe2cd1009d1353831b0a41c6/Zcash/Common/DiscreteLogRelation.lean#L385).

**Implementation correctness:** no additional arithmetic, relation-embedding, or probability-bound regression was found. The soundness findings concern the theorem’s coverage; no concrete inflation exploit was demonstrated.

**Posted with the help of fable.**

daira:
Status of the review findings, by number, at the pushed head:

**Addressed**

3. *Missing computability pin.* Pinned, together with the other reduction producers this PR introduced without a direct pin: `conservationFinder`, `noteCommitBreakAtBasis`, `noteCommitBreakOfKappa`, `orchardBalanceSubsetOrRelationAtBasis`, `toOrchardPoints`, `extractorAtBasis`, and in the Common layer `NontrivialRelation.ofParts`, `AlgebraicRelationWitness.embed`, and the U/W dischargers.

5. *Prose.* κ is described at every site as a bound on the extraction-failure probability, obtained as the measure of the pivot event (a verifying signature whose effective representation has a key coefficient off the ℛ slot). That event contains every extraction failure and also the samples whose off-ℛ coefficients cancel as a group element, which is why it is not the failure event itself. The relation module's doc now says that the known-log dischargers sum over the index. The three docstrings that still called the witness annotations a gap tracked as #147 describe the composed route that discharges them.

   On the stale links: every file link on the games page resolves at the reviewed head, so please point me at the ones you mean.

4. *Nullifier reduction target.* The combined deployed basis now has a 𝒦^Orchard slot (`idxNullifierK`). `relationOfNullifierCollision` computes the relation at the deployed Orchard primitives directly. The deployed derivation is the extracted coordinate of the commitment shifted by the nullifier-scalar multiple of 𝒦, so equal nullifiers give a point equation up to sign. A shifted form of the chain-collision reducer then lands it in the combined basis. `NullifierCollision` now carries the value bounds, as `NoteCommitBreak` does, since the deployed encoding recovers a note from its chunks only under them. The games page's nullifier edge links the deployed reducer.

**Outstanding, tracked**

1. *Oracle composition.* The per-size tables are functions on `BTranscript` values of a size-determined length, so queries at different bundle sizes are distinct domain elements, and the ledger machine receives every size's table. What makes that a faithful presentation of one deployed oracle is that the Blake2b squeeze inputs at different action counts never coincide. #215 proves exactly that in `Zcash/Snark/Soundness/FiatShamir/ActionCount.lean` (the pre-θ prefixes at different counts are prefix-incomparable at the typed and byte levels, and reprogramming one count's cone leaves the other count's challenges unchanged). The composed experiment does not yet consume those facts, so at this head the module doc states the per-size presentation as a modeling caveat, naming the separation across action counts as the fact it rests on. Next on this PR: port the typed-level half of that proof (it depends only on modules already here) and use it in the experiment; the byte-level half is tracked as #224, dependent on #215.

2. *Restricted representation basis.* Agreed: the binding-signature AGM hypothesis quantifies over adversaries algebraic over 𝒱 and ℛ alone, and translating a representation over another public point into that basis is itself a relation, so the class is narrower than a deployed algebraic adversary. This is the per-game enumerated-basis narrowing that #188 records, now with a concrete instance: the combined deployed basis is the natural target for the adapter (widen `AlgebraicAtBindingPointsAt` to the combined basis, with the pivot over every non-ℛ slot). [Noted on #188](https://github.com/zcash/ironwood/issues/188#issuecomment-5572296818).

Posted with the assistance of Claude Fable 5.1

TalDerei (approved):
optimistic ACK

daira (commented):
Self-reACK

PR #220, opened 2026-08-28
cleanup: spell out short argument and advantage names

daira:
Spell out the short argument names in the security layer, where the letter named a domain object: the presented basis `b` becomes `basis`, the sampled basis logs `s` become `logs`, extractors `E` become `extractor`, extraction failures `e` become `failure`, and the value-commitment and binding-signature shapes `S` and `B` become `shape` and `binding`. Break values named `b` become `brk` (`break` is a reserved keyword), including the `SpendAuthorityKBArm` field and its projection. The challenge tables `O` become `table`, the shift functions `s : Q → F` of the key-binding escape argument become `shift`, the annotated transaction-representation pairs `p` and `pr` become `tx_rep`, and the query point `q` becomes `query`. The advantage names gain the underscore the rest of the ε family already uses: `εdlr`, `εm`, `εnc`, `εkb`, `εf`, `εBias`, and `εx` become `ε_dlr`, `ε_merkle`, `ε_nc`, `ε_kb`, `ε_forge`, `ε_bias`, and `ε_bad`.

Letters naming types, events, list elements, and literature notation are unchanged: Merkle's `E`/`B` are type variables, the capstone measure lemmas' `E` is an event, `Balance`'s `find?` binders are list elements, and `σ`, `vk`, `R`, `ℓ`, and `qH` are standard notation. The witnesses named `w` also stay: the letter is unambiguous, and the bridge layer already uses `wit` for the circuit witness. The extraction module's sample paragraph now states what the logs are logs of — each presented basis element's discrete logarithm relative to the generator — and a Markov-category citation moves to the standard format.

In several references to ZIP 2005's `ε_kb`, the fact that its bound was sharpened in zcash/zips#1338 is dropped.

The κ-layer basis-generalization work (#219) stacks on this branch.

🤖 Claude Fable 5

### DIFF
--- a/Zcash/Security/KeyBinding/Basic.lean
+++ b/Zcash/Security/KeyBinding/Basic.lean
@@ -20,8 +20,8 @@ verifying Recovery-Statement witness pins the key components — `ak` up to y-si
 
 The route, each step proven here:
 
-1. `commit_scalar_pm` — two openings of one `Commitivk` value force their Pedersen scalars equal
-   or negated; `CollisionUpToSign.ofOpeningBreak` packages this as computed data.
+1. `commit_scalar_pm` — two openings of one `Commitivk` value force their Pedersen scalars
+   equal or negated; `CollisionUpToSign.ofOpeningBreak` packages this as computed data.
 2. `rivk_eq_finalOracle` — under the derivation constraints, `rivk` is the combined final
    oracle's output at the witness's decoded query.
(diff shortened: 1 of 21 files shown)


daira (commented):
Self-ACK

daira (commented):
Self-reACK

TalDerei (commented):
post-hoc ACK

PR #225, opened 2026-09-08
Remove the circuit layer's references to the phase-1 donor development

daira:
The circuit layer was copied from the phase-1 Clean-based development, "the donor". Nothing has imported that development for some time, but the layer still carried two kinds of reference to it.

* Four private lemmas in the CommitIvk and NoteCommit main bundles (`pieceChunks_donor_iff`, `zsFacts_donor_iff`, `pieceBounds_donor_iff`, `honestChunks_donor_eq`) had become statements of a proposition's equivalence with itself once the donor's definitions were replaced by the copies here. They and their eight no-op uses are deleted.
* Docstrings and comments across the layer carried provenance notes ("Phase-1 donor: …", "Donor `X`.") and described gate contracts as "the donor …". The notes are deleted; the sentences that also said something about the present code now say only that ("`Spec`, the composite's guarantee, gives …", "the gate `Spec`").
* The `toDonor` row conversions and the private `DRow`/`DSpec`/`DAssumptions` aliases (the "D" stood for the donor) are renamed `toGateRow`, `GateRow`, `GateSpec`, and `GateAssumptions`.

No statement changes: the lemma deletions remove identities, and everything else is documentation or renaming.

🤖 Claude Fable 5.1

### DIFF
--- a/Zcash/Circuits/Action/AddressIntegrity.lean
+++ b/Zcash/Circuits/Action/AddressIntegrity.lean
@@ -22,11 +22,7 @@ separate building block; its output cell feeds in here as `ivk`):
    region (`ecc/chip.rs:474-488`), two copy constraints.
 
 The block returns the witnessed `pk_d_old`. `Spec` is knowledge-sound with no
-existential: `pk_d_old = ivk • g_d_old` at the input `ivk` cell (the phase-1 donor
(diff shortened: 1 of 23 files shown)


daira (commented):
Self-ACK

mitschabaude (approved):
thanks for that cleanup!

TalDerei (approved):

